PR #2, opened 2025-05-24
feat: quick add task tool

gnapse:
## Summary

Adds a `quick-add-task` MCP tool that creates a single task from one natural-language string using Todoist's Quick Add parser — the same syntax users type into the Todoist apps. Everything goes inline in `text`: due date (`tomorrow at 5pm`, recurring `every Monday`), priority (`p1`–`p4`), labels (`@label`), project (`#Project`), section (`/Section`), assignee (`+name`/`+email`), description (after ` // `), and reminders (`!30m`, `!1h`, `!2025-12-31 at 9am`).

### Why a separate tool from `add-tasks`?

- **Inline reminders.** Quick Add parses `!30m`-style reminders server-side; with `add-tasks` the agent has to build reminder objects via a separate tool call.
- **User-style capture.** Best fit for relaying something the user typed verbatim, where coercing into structured fields would lose fidelity (e.g. natural recurring rules).
- **`add-tasks` stays the default** for bulk creation, ID-based workflows, deadlines, durations, and precise assignment validation. Tool descriptions and the MCP instructions block route the agent accordingly.

### Notes on the contract

- Single input field `text` plus an optional `autoReminder` boolean — no `note` / `reminder` side-channels, so there's one unambiguous way to express each concept.
- Project / section / assignee are resolved by **name** from the string, not by ID.
- Output uses the shared `summarizeTaskOperation` formatter so responses match `add-tasks` / `update-tasks`.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run format:check`
- [x] `npx vitest run` — 858 tests passing, including new `src/tools/__tests__/quick-add-task.test.ts`
- [ ] Smoke test via `npx tsx scripts/run-tool.ts quick-add-task '{"text":"Call mom tomorrow at 5pm #Personal p2 !30m"}'`
- [ ] Verify in MCP client that the agent prefers `quick-add-task` for NL strings with inline reminders and `add-tasks` for structured bulk input


goncalossilva (approved):
Clean & simple 👌

amix (approved):
👌

gnapse:
I’m hesitant that we need this at all. The regular current way of adding tasks works. Why would AI agents need two ways to do it? It would add to the context, and create confusion. What do you folks think about this?

scottlovegrove:
The only thing I would say, using the quick add endpoint (function in the sdk), would mean reminders like `!30mb` would then get parsed (this is a recent backend change). If we use the regular add task, then reminders have to be generated as objects.

Not saying this is an absolute reason to use it, because the agents can be told how to create reminders (maybe this is an improvement we can make to the MCP docs?)

gnapse:
My take is that there are 3 downsides to this:

1. This approach needs us to teach the agent about syntax, while the other approach, while it may require more commands, the agent can learn how to do it by just seeing that there's a tool to add reminders, with self-documented interface.
2. This introduces ambiguity. The agent sees two ways to add tasks, and has to make decisions, with both options having a different interface.
3. This adds extra context to interactions with the agent (the extra tool name, description and arg docs)

src/mcp-server.ts
```diff
@@ -205,6 +207,7 @@ function getMcpServer({
 
     // Task management tools
     registerTool({ tool: addTasks, ...toolArgs })
+    registerTool({ tool: quickAddTask, ...toolArgs })
```
doistbot, line 210: [P2] This only registers the tool with the MCP server. `quick-add-task` is still missing from `src/index.ts` and `scripts/run-tool.ts`, so `npm run tool quick-add-task ...` will currently fail, and `scripts/validate-schemas.ts` will also skip validating this schema because it imports the `tools` object from `dist/index.js`.

src/tools/quick-add-task.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ToolNames } from '../utils/tool-names.js'
+
+const ArgsSchema = {
+    text: z
+        .string()
+        .min(1)
+        .describe(
+            'The full task text using Todoist\'s natural-language syntax. Supports inline date ("tomorrow at 5pm"), priority ("p1"), labels ("@work"), project ("#Inbox"), and assignee ("+Alice").',
+        ),
+    note: z.string().optional().describe('Optional note/description to attach to the new task.'),
```
doistbot: [P2] `note` and `reminder` should use the shared `optionalString(...)` helper instead of raw `z.string().optional()`. In this repo we already normalize optional strings because models often send `""` for omitted fields; as written, an empty `note`/`reminder` is passed through to `client.quickAddTask(...)` and can trigger avoidable Todoist API validation errors.

src/tools/quick-add-task.ts
```diff
@@ -0,0 +1,57 @@
...
+            text,
+            note,
+            reminder,
+            autoReminder,
+        })
+
+        const mapped = mapTask(created)
+
+        return {
+            textContent: `Created task "${mapped.content}" (id: ${mapped.id}).`,
```
doistbot: [P3] This hand-built success message duplicates task-summary formatting that is already centralized in `src/utils/response-builders.ts`. Reuse `summarizeTaskOperation('Added', [mapped], { showDetails: true })` here so quick-add stays aligned with `add-tasks`, `update-tasks`, and `reschedule-tasks` output instead of introducing another response format to maintain.

src/tools/quick-add-task.ts
doistbot: [P2] This turns `quick-add-task` into a hybrid API: description can come from `text` (` // ...`) or `note`, and reminders can come from Quick Add syntax or the separate `reminder` field. `execute()` forwards all of them without any mutual-exclusion check, so behavior depends on backend precedence. To keep the contract predictable, either expose only the raw `text` form or reject combinations that specify the same concept both inline and structurally.

src/mcp-server.ts
```diff
@@ -72,6 +73,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 
 **Task Management:**
 - **add-tasks**: Create tasks (max 25 per call) with content, description, priority (\`p1\`, \`p2\`, \`p3\`, \`p4\` strings only; \`p1\` highest and \`p4\` lowest/default; integers are not accepted), dueString (natural language like "tomorrow", "next Friday", "2024-12-25"), deadlineDate (ISO 8601 format like "2025-12-31" for immovable constraints), duration (formats like "2h", "90m", "2h30m"), and assignments to project collaborators
+- **quick-add-task**: Create a **single** task from one natural-language string using Todoist's Quick Add syntax (the same parser the Todoist apps use). Inline syntax: due date (\`tomorrow at 5pm\`, recurring \`every Mon\`), priority (\`p1\`–\`p4\`), labels (\`@label\`), project (\`#Project\`), section (\`/Section\`), assignee (\`+name\`/\`+email\`), description (text after \` // \`). Resolves project/section/assignee by **name**, not ID. Use when relaying something the user typed in NL form. Do **not** use for: bulk creation (use add-tasks), tasks needing a deadlineDate or duration (use add-tasks), or when you already have IDs and structured fields (use add-tasks).
```
doistbot: [P2] This repeats most of `quick-add-task`’s own schema/description in the always-on server instructions. Since MCP already exposes the tool description and args separately, duplicating the full syntax/routing guidance here adds avoidable prompt-token overhead to every session for a capability that overlaps `add-tasks`. Please shorten this to a brief routing hint and keep the detailed Quick Add syntax in the tool definition.

src/tools/__tests__/tool-annotations.test.ts
```diff
@@ -24,6 +24,13 @@ const TOOL_EXPECTATIONS: ToolExpectation[] = [
         destructiveHint: false,
         idempotentHint: false,
     },
+    {
```
doistbot, line 27: [P2] This is the only test coverage added for the new tool, which means `quick-add-task` currently has no behavioral test at all. Please add `src/tools/__tests__/quick-add-task.test.ts` to exercise `execute()` end to end: parse the required `text` input, verify the optional `note`/`reminder`/`autoReminder` fields are forwarded to `client.quickAddTask`, and assert the returned task is mapped into the expected structured/text output.

doistbot (commented):
This PR updates feat: quick add task tool. The inline comments cover the issues found in the review.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-ai&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-ai%2Fpull%2F2&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae49c5540-49f2-11f1-88d2-8ececa39858b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae49c5540-49f2-11f1-88d2-8ececa39858b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove:
Have discussed with @gnapse and I made the case for this tool to exist. Basically I had an automation where I wanted to just pass through content and have it parsed into a Todoist task. My example text was our usual `buy milk tomorrow`, and that went through the normal add-tasks tool and produced a task with the same content. 

Ernesto suggested that we have an additional step in the automation that could handle parsing all those so that `add-tools` can be done, but my argument here is that users have an expectation that doing something like "buy milk tomorrow" will automatically be parsed to "buy milk" with a due date of tomorrow. They shouldn't have to faff with an additional step, it should just work. 

The concensus was that we need to make this very explicit in our tool naming and descriptions so that LLMs know exactly which tool to use, and when, which is what I've done (https://github.com/Doist/todoist-ai/pull/2/commits/bb8bd15bf59f46b8a36d8b790de0c4694b9175b0)

amix:
From a compute perspective, we’d prefer LLMs to use `ToolNames.ADD_TASKS` since it’s much more performant. `ToolNames.QUICK_ADD_TASK` pushes parsing to the backend, which makes it significantly more expensive.

I think for one-off tasks, it could work, but I would be concerned if LLMs used this as their preferred way to add tasks 🤔

scottlovegrove:
> I think for one-off tasks, it could work, but I would be concerned if LLMs used this as their preferred way to add tasks 🤔

@amix We could ensure that the wording makes it clear that `add-tasks` is the preferred way, I'm just wondering how this would apply to Automations and ensuring that what the user has passed through gets parsed correctly. We could end up with a bunch of automations all re-inventing the wheel for this, just to avoid using `quick-add`.

goncalossilva:
Where does `buy milk tomorrow` originally come from? An LLM has intelligence, so it should be able to fit prompts into `add-task` if the tool is well-described. 🤔

gnapse:
From https://github.com/Doist/todoist-cli/issues/342#issuecomment-4476539080, a good reason to rethink if we really want a tool for this:

> ### `td task add` avoids any parsing trickiness
> 
> Consider the following `quickadd` examples:
> 
>     * `td task quickadd "Call James about the quarterly report"`
> 
>     * `td task quickadd "Cancel Sunday newspaper delivery"`
> 
> 
> In the first case, the `quickadd` parser sees "quarterly", uses that as the task recurrence schedule, and you wind up with just "Call James about the report" in your task title. In the second case, "Sunday" is stripped from the task title in a similar manner and used as the due date. Likely not what you wanted for either task!

In Todoist's UI, users see the parsing in real time before submitting, and they can remove any accidental matches (i.e. they can press backspace on a word that was parsed as metadata, but it was intended for the task content).

With a tool like this, either here in the MCP server of in the CLI, it can create not only confusion, but real problems sometimes.

goncalossilva:
I think we should drop it. The LLM can play "quick add" if it wants or the user instructs it to. 😊

scottlovegrove:
It's not a hill I'm particularly bothered about dying on 😄

PR #6, opened 2025-07-31
ci: automate release process

gnapse:
This is a blatant copy of https://github.com/Doist/reactist/pull/863, hence my targeted review request for Rui.

I guess I now have to configure the secrets (`GITHUB_TOKEN` and `NPM_TOKEN`), which I'm not sure where to get from. ⚠️ **Also, please know that this repository may eventually be public, so avoid discussing anything sensitive here.**

Is this all it takes? Will this work out-of-the-box after that is configured and this gets merged? Or is there something else to be done?

~~For instance, if I merge this before #5, and then I merge that PR, will that trigger the v0.2.0 release? _(Because it's a `feat` I assume it will increase in that way, but I'm not sure about semver on v0.x which behaves differently; but you get my point.)_~~

I forgot that I merged #5 beforehand, so now I do not have a pull request with which to test this. But I still need to get the changes in #5 released. What do you think I should do? Release those changes first manually?


henningmu:
> `GITHUB_TOKEN`

It's the regular token for Action workflows and should always be available: https://docs.github.com/en/actions/tutorials/authenticate-with-github_token  👍

> `NPM_TOKEN`

needs to be added as a repository secret under settings. Can be found in Bitwarden under `NPM @doist Publish Token`

gnapse:
> would advise to run this configuration against a fork of the repository or just a test repository if you didn't already.

What do you mean exactly by that?

rmartins90:
> What do you mean exactly by that?

When I first implemented it, I created a separate repository to test different scenarios with different PRs. But it was in a tricky project, here it's pretty more straightforward.

gnapse:
> > `NPM_TOKEN`
> 
> needs to be added as a repository secret under settings. Can be found in Bitwarden under `NPM @doist Publish Token`

I see in https://github.com/Doist/todoist-ai/settings/secrets/actions that we have a `NPM_PUBLISH_TOKEN` already configured at the org level. Does that mean that maybe I do not need to configure it, and instead I should modify the var name in this PR to use that name? @henningmu

henningmu:
@gnapse try it please. I can't say for sure, but sounds sensible 👍

PR #8, opened 2025-08-12
feat: Adds completed tasks tool

scottlovegrove:
This adds a new tool that allows for fetching of completed tasks either by due date or by completion date.

### DIFF
--- a/src/tools/tasks-list-completed.ts
+++ b/src/tools/tasks-list-completed.ts
@@ -1,83 +1,69 @@
-import { z } from "zod";
-import type { TodoistTool } from "../todoist-tool";
-import { mapTask } from "./shared";
+import { z } from 'zod'
+import type { TodoistTool } from '../todoist-tool'
+import { mapTask } from './shared'
 
 const ArgsSchema = {
-	getBy: z
-		.enum(["completion", "due"])
-		.describe(
-			'The method to use to get the tasks: "completion" to get tasks by completion date (ie, when the task was actually completed), "due" to get tasks by due date (ie, when the task was due to be completed by).',
-		),
-	since: z
-		.string()
-		.date()
-		.regex(/^\d{4}-\d{2}-\d{2}$/)
-		.describe("The start date to get the tasks for. Format: YYYY-MM-DD."),
-	until: z
-		.string()
-		.date()
-		.regex(/^\d{4}-\d{2}-\d{2}$/)
-		.describe("The start date to get the tasks for. Format: YYYY-MM-DD."),
-	workspaceId: z
-		.string()
-		.optional()
-		.describe("The ID of the workspace to get the tasks for."),
-	projectId: z
-		.string()
-		.optional()
-		.describe("The ID of the project to get the tasks for."),
-	sectionId: z
-		.string()
-		.optional()
-		.describe("The ID of the section to get the tasks for."),
-	parentId: z
-		.string()
-		.optional()
-		.describe("The ID of the parent task to get the tasks for."),
-	filterQuery: z
-		.string()
-		.optional()
-		.describe(
-			"The filter query to use to filter the tasks. If set, you must provide a `filterLang` parameter.",
-		),
-	filterLang: z
-		.string()
-		.optional()
-		.describe(
-			"The filter language to use to filter the tasks. Should only be provided if the `filterQuery` parameter is set.",
-		),
-	limit: z
-		.number()
-		.int()
-		.min(1)
-		.max(200)
-		.default(50)
-		.describe(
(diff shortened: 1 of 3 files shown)


src/tools/tasks-list-completed.ts
```diff
@@ -0,0 +1,36 @@
...
+	until: z.string().date(),
+	workspaceId: z.string().optional(),
+	projectId: z.string().optional(),
+	sectionId: z.string().optional(),
+	parentId: z.string().optional(),
+	filterQuery: z.string().optional(),
+	filterLang: z.string().optional(),
+	limit: z.number().int().min(1).max(50).default(50),
+	cursor: z.string().optional(),
+};
```
gnapse: Add `.describe("…")` calls to each arg (or at least to those where it makes sense) as it helps LLMs to know when and how to use each arg.

Also, the `since` and `until` args should probably not be of type `date` but `string` instead. And then you instruct the LLM via the arg description about how to pass the dates. 

For instance, see https://github.com/Doist/todoist-ai/blob/ee88f1a85084610ccf3ca4d01c61f9e53fde5678/src/tools/tasks-list-by-date.ts#L7-L12

_(maybe in this use case we do not need to support the special value "today," but you get the gist)_

Remember that these tools are meant to be invoked by sending a JSON payload, and you cannot send `Date` objects via JSON. This is meant to be interfaced with via text.

scottlovegrove: Good point

src/tools/tasks-list-completed.ts
```diff
@@ -0,0 +1,36 @@
+import { z } from "zod";
+import type { TodoistTool } from "../todoist-tool";
+import { mapTask } from "./shared";
+
+const ArgsSchema = {
+	getBy: z.enum(["completion", "due"]),
```
gnapse: About the `describe("…")` calls, this field probably requires it. We need to briefly explain to the LLM what each value means. Also consider leaving it optional with a default value, if one of these two values makes more sense to be used in most cases.

scottlovegrove: Yep, already added it :)

src/tools/tasks-list-completed.ts
```diff
@@ -0,0 +1,69 @@
+import { z } from 'zod'
+import type { TodoistTool } from '../todoist-tool'
+import { mapTask } from './shared'
+
+const ArgsSchema = {
+    getBy: z
+        .enum(['completion', 'due'])
+        .describe(
+            'The method to use to get the tasks: "completion" to get tasks by completion date (ie, when the task was actually completed), "due" to get tasks by due date (ie, when the task was due to be completed by).',
+        ),
```
gnapse: I'd still encourage you to make this one optional and add a default. Or to even omit the argument entirely.

My thinking is this:

- remember this will be used by an LLM when instructed by a user that they want to see their completed tasks
- my intuition is that users will not be specifying _"show me my completed tasks **by the date they were due**"_
- I lean to think that the more sensible default is by completion

src/tools/tasks-list-completed.ts
```diff
@@ -0,0 +1,69 @@
...
+        .optional()
+        .describe(
+            'The filter query to use to filter the tasks. If set, you must provide a `filterLang` parameter.',
+        ),
+    filterLang: z
+        .string()
+        .optional()
+        .describe(
+            'The filter language to use to filter the tasks. Should only be provided if the `filterQuery` parameter is set.',
+        ),
```
gnapse: Your average LLM out there won't know about Todoist's filterist syntax, making this not so useful unless users would provide filterist syntax themselves, which may be unlikely.

Our own "filter assist with AI" feature requires a big prompt just to teach the LLM how the filterist syntax works. And we do not have the luxury of including this prompt here.

In the future, we could enable this via the use of MCP features other than tools, such as [sampling](https://modelcontextprotocol.io/specification/2025-06-18/client/sampling) or [prompts](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts).

I'd lean towards not including these for now.

gnapse (approved):
Approved.

Though I left a couple of comments below about a few things that I think we can drop. The overall rationale is the same: to provide what could be useful and likely to be used by users when chatting with an LLM.

src/tools/tasks-list-completed.ts
scottlovegrove: Makes sense

src/tools/tasks-list-completed.ts
scottlovegrove: Makes sense

PR #9, opened 2025-08-12
chore: Updates linting config and adds CI checks

scottlovegrove:
As the title says...


gnapse:
We did not have any tests or even jest set up here, so I added jest and a basic test, so that the new CI workflows would pass, and we're ready to add more tests in the future.

PR #16, opened 2025-08-13
chore: Update to node 22

scottlovegrove:
# Pull Request

## Short description

This updates the repo to use node 22, it was currently using no specific node version, so whatever was the default on the machine. We should be explicit.


gnapse (commented):
✅

henningmu:
Why aren't we using node24? 🤔

scottlovegrove:
So we don't run into something similar to this https://github.com/Doist/agentist/pull/72

PR #18, opened 2025-08-13
chore: configure dependabot to use conventional commits

gnapse:
## Summary
Configure dependabot to use conventional commit format for dependency updates.

This ensures that dependabot PRs follow the same commit message format as the rest of the project, which is required for release-please to properly categorize changes.

## Changes
- Added `.github/dependabot.yml` configuration
- Set commit message prefix to "chore" for dependency updates
- Configured weekly schedule for npm updates

## Test plan
- After merging, new dependabot PRs should use format like "chore: bump package-name from x.y.z to a.b.c"

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "chore"
+      prefix-development: "chore"
+      include: "scope"


henningmu:
LGTM, but why is this using dependabot and not renovate?

gnapse:
> LGTM, but why is this using dependabot and not renovate?

Just because I noticed dependabot PRs already in place and not using the commit prefix (e.g., #3 and #4). But you're right, dependabot is not even configured here. I think it's just enabled in the settings. Will add renovate config in #25, close this PR and the ones dependabot created, and will disable dependabot.

PR #28, opened 2025-08-15
feat: add support for task duration when adding or updating tasks

### LINKED ISSUE #27: Add Duration Field Support to Task Update Tool
# Add Duration Field Support to Task Update Tool

## Problem
The Todoist API supports `duration` and `duration_unit` fields for tasks (as documented in the OpenAPI spec), but the current MCP tool implementation doesn't expose these fields. This prevents users from setting actual task durations programmatically, limiting them to text-based conventions like "from 10:00 for 2h" in the task name.

## Current Behavior
- When updating a task with duration notation (e.g., "from 10:00 for 2h"), it's stored as plain text in the task content
- The actual Todoist duration feature (Pro/Business) isn't triggered
- Users cannot leverage Todoist's calendar view duration functionality via the MCP tool

## Expected Behavior
- The `tasks-update-one` tool should accept `duration` and `duration_unit` parameters
- The `tasks-add-multiple` tool should also support these fields
- Duration should be properly set in Todoist's system, not just as text

## API Support Confirmation
From `openapi-todoist.json` (lines 8346-8374):
```json
"duration": {
  "anyOf": [
    {
      "type": "integer"
    },
    {
      "type": "null"
    }
  ],
  "title": "Duration"
},
"duration_unit": {
  "anyOf": [
    {
      "type": "string"
    },
    {
      "type": "null"
    }
  ],
  "title": "Duration Unit"
}
```

## Affected Files
- `src/tools/tasks-update-one.ts`
- `src/tools/tasks-add-multiple.ts`
- `src/tools/__tests__/tasks-update-one.test.ts`
- `src/tools/__tests__/tasks-add-multiple.test.ts`

## Use Case
Users following time-blocking methodologies need to:
1. Set estimated durations when planning tasks
2. Update durations based on actual time spent
3. Generate time reports from completed tasks with duration data

## Proposed Solution
Add `duration` (integer, in minutes) and `durationUnit` (string: "minute" or "hour") as optional parameters to the task creation and update tools.

## Additional Context
- This is particularly important for Pro/Business users who have access to Todoist's duration feature
- The feature has a 24-hour maximum duration limit
- Requires tasks to have both date and time set

gnapse:
Closes #27 

## Short description

It follows a similar approach to the proposed solution in https://github.com/Doist/todoist-ai/issues/27#issuecomment-3190219489, but with an improvement on top of it that allows the LLM to pass a more LLM/human readable duration in string form (e.g. `2h30`, `1h`, `1.5h`, `45m`, etc.) and this value is parsed internally and converted to a raw minutes number to be passed to the API.

## PR Checklist

### DIFF
--- a/src/tools/tasks-add-multiple.ts
+++ b/src/tools/tasks-add-multiple.ts
@@ -1,13 +1,20 @@
-import type { Task } from '@doist/todoist-api-typescript'
+import type { AddTaskArgs, Task } from '@doist/todoist-api-typescript'
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { mapTask } from '../tool-helpers.js'
+import { DurationParseError, parseDuration } from '../utils/duration-parser.js'
 
 const TaskSchema = z.object({
     content: z.string().min(1).describe('The content of the task to create.'),
     description: z.string().optional().describe('The description of the task.'),
     priority: z.number().int().min(1).max(4).optional().describe('The priority of the task (1-4).'),
     dueString: z.string().optional().describe('The due date for the task, in natural language.'),
+    duration: z
+        .string()
+        .optional()
+        .describe(
+            'The duration of the task. Use format: "2h" (hours), "90m" (minutes), "2h30m" (combined), or "1.5h" (decimal hours). Max 24h.',
+        ),
 })
 
 const ArgsSchema = {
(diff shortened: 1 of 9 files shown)


src/tools/tasks-add-multiple.ts
```diff
@@ -1,13 +1,20 @@
...
 import { mapTask } from '../tool-helpers.js'
+import { DurationParseError, parseDuration } from '../utils/duration-parser.js'
 
 const TaskSchema = z.object({
     content: z.string().min(1).describe('The content of the task to create.'),
     description: z.string().optional().describe('The description of the task.'),
     priority: z.number().int().min(1).max(4).optional().describe('The priority of the task (1-4).'),
     dueString: z.string().optional().describe('The due date for the task, in natural language.'),
+    duration: z
+        .string()
```
scottlovegrove, line 13: 💭 I wonder if this should be a regex, would allow us to remove some of the validation code from the `parseDuration` function.

scottlovegrove (approved):
Approving. My only comment was a thought really, feel free to ignore.

src/tools/tasks-add-multiple.ts
gnapse: Makes sense, though it would not remove the validation in the function, because deep down that function is independent from this, and receives a string anyway[^1], so it has to double-check what it received and be able to parse it and fail if needed.

So, OTOH, this makes me lean toward not adding it for now. Because then we'd have two separate validation processes that could even end up diverging over time. We can always revisit this later.

[^1]: Adding a regex to the zod schema does not narrow down the type, and the `duration` attribute after zod validates the input will still be of type string.

scottlovegrove: > 1. Adding a regex to the zod schema does not narrow down the type, and the `duration` attribute after zod validates the input will still be of type string

It would still be a string, but it would mean if we did `schema.parse(input)` it would get validated through the schema's parsing.

gnapse: Yes, but the function will still receive a string, and yes, it could parse it optimistically, assuming that its input is valid because of the regex validation filter. But it will still need to parse it, and that parsing could fail (e.g. imagine we do not get the regex entirely right for some edge cases). The amount of supported syntax here is not trivial to express in a regex. Plus, validating programmatically inside the function also allows us to return errors with custom messages for what failed, which are then consumed by the LLM using the MCP server, which can help redirect it to the right syntax, if it ever gets it wrong.

PR #47, opened 2025-08-19
feat!: Remove single update task and replace with multiple

scottlovegrove:
# Pull Request

## Short description

Nearly all of our tools for updating a task (completing, organising, etc) allow for multiple tasks to be sent through, but doing a regular update does not. This PR removes the single task update tool and replaces it with a tool that allows for updating multiple tasks in one go. 

## PR Checklist


src/tools/__tests__/tasks-update-one.test.ts
gnapse, line 1: We should also rename this file.

And it should include at least one test about how it works when passing in multiple tasks.

gnapse (approved):
I agree with this change. Just noted a few things that I think are important before we merge it.

Also, this may be a reason to remove `tasks-organize-multiple`. What do you think? If you agree, we can do it in this very pull request, as it all is already a breaking change.

scottlovegrove:
> Also, this may be a reason to remove `tasks-organize-multiple`. What do you think? If you agree, we can do it in this very pull request, as it all is already a breaking change.

Makes sense, I'll look at removing that.

PR #51, opened 2025-08-19
feat: migrate from Release Please to semantic-release

### LINKED ISSUE #46: Use of non-approved third-party GitHub Actions
# Use of non-approved third-party GitHub Actions

Some GitHub Actions workflows in todoist-ai repository use third-party actions that fall outside our approved tiers:

- [ ] [amannn/action-semantic-pull-request@db6e259b93f286e3416eef27aaae88935d16cf2e](https://github.com/search?type=code&q=repo%3ADoist%2Ftodoist-ai+amannn%2Faction-semantic-pull-request%40db6e259b93f286e3416eef27aaae88935d16cf2e)

According to our [GitHub Actions handbook][handbook], we only allow:

- **Tier 1**: Actions from trusted organizations (GitHub, AWS, Google, etc.)
- **Tier 2**: Audited actions pinned to specific commit SHAs

**Required action:**

Choose one of these options:

1. **Replace** with an approved alternative or custom script
2. **Audit** the action's code and add it to [Tier 2][allowlist] (pinned to full SHA)
3. **Discuss** in [#Doist Dev](https://twist.com/a/1585/ch/1265/) if you believe the author should be added to Tier 1

See the [handbook][] for detailed guidance on each option.

(Relates to https://github.com/Doist/platform-backlog/issues/983)

[handbook]: https://handbook.doist.com/doc/github-actions-TNArIDr8Jb
[allowlist]: https://github.com/Doist/allowed-actions

gnapse:
## Summary
- Migrates from Release Please to semantic-release for automated releases
- Removes the non-approved third-party action `amannn/action-semantic-pull-request` (fixes #46)
- Implements PR title validation using official GitHub Actions and commitlint

## Changes
- ✅ Removed all Release Please configuration files and workflows
- ✅ Removed `amannn/action-semantic-pull-request` workflow
- ✅ Added semantic-release with conventionalcommits preset
- ✅ Added commitlint for PR title validation (uses only official GitHub Actions)
- ✅ Configured release workflow to trigger after CI validation passes
- ✅ Followed configuration patterns from the typist repository

## Benefits
- **Faster releases**: Automatic release on every merge to main (no PR review needed)
- **Tier 1 compliance**: Uses only official GitHub Actions for PR validation
- **Proven approach**: Follows the same pattern used in typist repository
- **Better automation**: Releases happen immediately after CI passes

## Breaking Changes
⚠️ **BREAKING CHANGE**: Release process now uses semantic-release instead of Release Please. Releases will be triggered automatically on merge to main after CI passes.

## Test Plan
- [ ] PR title validation workflow runs successfully
- [ ] CI workflow completes successfully
- [ ] After merge, semantic-release workflow triggers and creates a release
- [ ] NPM package is published correctly

## Migration Notes
After merging this PR:
1. Ensure the `NPM_TOKEN` secret is properly configured in the repository
2. Future PRs must use conventional commit format in their titles
3. Releases will happen automatically - no manual intervention needed

Resolves #46


.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: npm ci
+
+      - name: Run tests
+        run: npm test
+
+      - name: Build
+        run: npm run build
+
+      - name: Release
+        id: semantic-release
```
rfgamaral, line 47: We can remove this, it's not being used.

```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rfgamaral, line 49: In Typist we are using `secrets.GH_REPO_TOKEN`, which comes from organization secrets. I don't see that one in this repo's secrets. Not sure if I'm missing permissions, or it's just not there for some reason. But I would suggest use the same token to avoid issues.

rfgamaral: I would also suggest we add the following env vars:

```
GIT_AUTHOR_EMAIL: doistbot@users.noreply.github.com
GIT_AUTHOR_NAME: Doist Bot
GIT_COMMITTER_EMAIL: doistbot@users.noreply.github.com
GIT_COMMITTER_NAME: Doist Bot
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+      - main
+    types:
+      - completed
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+  packages: write
+  id-token: write
```
rfgamaral, line 17: Just so it's clear what these are for.

```suggestion
permissions:
    # Enable the use of OIDC for npm provenance
    id-token: write
    # Enable the use of GitHub Packages registry
    packages: write
    # Enable `semantic-release` to publish a GitHub release and post comments on issues/PRs
    contents: write
    issues: write
    pull-requests: write
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
```
rfgamaral, line 19: It's important we add this for the reasons mentioned in the comment (although we are not publishing this go GH Packages, I still think it's important to prevent a release in progress from being cancelled):

```suggestion
# The release workflow involves many crucial steps that once triggered it shouldn't be cancelled
# until it's finished, otherwise we might end up in an inconsistent state (e.g., a new release
# published to npm but not GitHub Packages). To prevent this, concurrency is disabled with
# `cancel-in-progress: false`, and new workflow runs will be queued to be started only when the
# previous one has completely finished.
concurrency:
    group: todoist-ai-package-release
    cancel-in-progress: false

jobs:
```

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Install commitlint
+        run: npm install @commitlint/cli @commitlint/config-conventional
+
+      - name: Lint PR title
+        run: echo "${{ github.event.pull_request.title }}" | npx commitlint
```
rfgamaral, line 27: This is mentioned in the PR description:

> Removes the non-approved third-party action `amannn/action-semantic-pull-request`

However, the version that Typist is using has been validated and approved, so I think we can safely use that one:

```suggestion
name: Semantic Pull Request

on:
    pull_request_target:
        types:
            - edited
            - opened
            - synchronize

env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

jobs:
    validate-title:
        name: Validate Title
        runs-on: ubuntu-latest
        timeout-minutes: 5

        steps:
            - name: Validate pull request title
              uses: amannn/action-semantic-pull-request@01d5fd8a8ebb9aafe902c40c53f0f4744f7381eb

```

This keeps the workflow consistent with the Typist one, which doesn't need to check out the repo and setup Node (slower).

.releaserc.json
```diff
@@ -0,0 +1,34 @@
...
+        ],
+        "@semantic-release/changelog",
+        "@semantic-release/npm",
+        [
+            "@semantic-release/git",
+            {
+                "assets": ["package.json", "package-lock.json", "CHANGELOG.md"],
+                "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
+            }
+        ],
```
rfgamaral, line 24: These are pretty much the defaults, so we don't need them:

```suggestion
        "@semantic-release/git",
```

.releaserc.json
```diff
@@ -0,0 +1,34 @@
...
+        "@semantic-release/github",
+        [
+            "@semantic-release/exec",
+            {
+                "verifyConditionsCmd": "echo \"package-published=false\" >> $GITHUB_OUTPUT",
+                "successCmd": "echo \"package-published=true\" >> $GITHUB_OUTPUT"
+            }
+        ]
```
rfgamaral, line 32: This is only required if we are publishing to npm and GH Packages in the same GH workflow, which we are not currently doing in this repo.

```suggestion
        "@semantic-release/github"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
```
rfgamaral, line 29: Like I mentioned below, we are using the `secrets.GH_REPO_TOKEN` for this token in Typist, and I have a feeling these tokens are the difference between a successful release and a failed one. I would look into using `GH_REPO_TOKEN` from the organization secrets in both these instances.

commitlint.config.js
rfgamaral, line 1: If we go with `amannn/action-semantic-pull-request@01d5fd8a8ebb9aafe902c40c53f0f4744f7381eb` as mentioned above, we can delete this file as it's no longer needed.

package.json
```diff
@@ -52,6 +52,13 @@
     },
     "devDependencies": {
         "@biomejs/biome": "1.9.4",
+        "@commitlint/cli": "^19.8.1",
+        "@commitlint/config-conventional": "^19.8.1",
```
rfgamaral, line 56: If we go with `amannn/action-semantic-pull-request@01d5fd8a8ebb9aafe902c40c53f0f4744f7381eb` as mentioned above, we don't need these dependencies.

We do need `conventional-changelog-conventionalcommits`, which is missing from this file but need in the `.releaserc.json` configuration.

package.json
```diff
@@ -52,6 +52,13 @@
...
+        "@semantic-release/changelog": "^6.0.3",
+        "@semantic-release/exec": "^7.1.0",
```
rfgamaral, line 58: If my suggestion to remove `@semantic-release/exec` from `.releaserc.json`, we can also remove this.

scottlovegrove:
@gnapse Please reopen if you feel otherwise, but I think we have a pretty good release workflow now and switching to semantic-release may not be needed anymore.

gnapse:
> I think we have a pretty good release workflow now and switching to semantic-release may not be needed anymore.

I agree

PR #53, opened 2025-08-20
feat!: verb-first tool names, unify find-by-container into find-tasks

gnapse:
## 🗺️ Overview

- Renamed tools to use a verb+object approach to naming
- Use the verb "find" for all listing and searching tools
- Unified a couple of task-finding tools into a single one

### Renamed Tools (Object-Verb → Verb-First)

  | Original Name           | New Name             | Description                             |
  |-------------------------|----------------------|-----------------------------------------|
  | `tasks-add-multiple`      | `add-tasks`            | Add multiple tasks to projects/sections |
  | `tasks-complete-multiple` | `complete-tasks`       | Mark multiple tasks as completed        |
  | `tasks-update-multiple`   | `update-tasks`         | Update multiple task properties         |
  | `tasks-list-by-date`      | `find-tasks-by-date`   | Find tasks by specific date ranges      |
  | `tasks-list-completed`    | `find-completed-tasks` | Find completed tasks with filtering     |
  | `projects-list`           | `find-projects`        | Search and list projects                |
  | `projects-manage`         | `manage-projects`      | Create and update projects              |
  | `sections-search`         | `find-sections`        | Find sections within projects           |
  | `sections-manage`         | `manage-sections`      | Create and update sections              |
  | `overview`                | `get-overview`         | Get account or project overview         |
  | `delete-one`              | `delete-object`        | Delete projects, sections, or tasks     |

### Consolidated Tools

  | Original Tools                          | Consolidated Into | Description |
  |-----------------------------------------|-------------------|----------------------------------------------------------------------------------------------------------------|
  | `tasks-search` + `tasks-list-for-container` | `find-tasks`        | Merged functionality: Text search combined with container-based filtering (by project, section, or parent task) |

## ⚠️ Breaking Changes

- API Changes: All tool names changed from object-verb to verb-first convention
- Tool Consolidation: Two separate tools merged into one unified interface
- Function Names: All exported functions updated to match new naming pattern
- Import/Export: All module imports and exports reorganized consistently

## ✨ Organization Improvements

- Consistent Grouping: Tools organized by category (Task → Project → Section → General)
- Centralized Names: All tool names managed through ToolNames constant
- Test Consistency: All tests use tool name constants instead of hardcoded strings
- File Structure: All source files, test files, and snapshots renamed consistently

Total: 11 tools renamed + 2 tools consolidated into 1 = 12 tools with consistent verb-first naming

## 🧪 Testing

- Follow the README to run the MCP server locally and connect it to Claude Desktop (or find and follow instructions online on how to use other MCP client app)
- Chat with the language model and ask it about your todoist task. Inspect the tool calling that it performs and see that it all works.


src/tools/find-tasks.ts
```diff
@@ -0,0 +1,232 @@
...
+
+        return getToolOutput({
+            textContent,
+            structuredContent: {
+                tasks: result.tasks,
+                nextCursor: result.nextCursor,
+                totalCount: result.tasks.length,
+                hasMore: Boolean(result.nextCursor),
+                appliedFilters: args,
+                searchQuery: searchText ? `search: ${searchText}` : undefined,
```
scottlovegrove: Curious, how would `searchText` be falsy here? Or is this just to satisfy the Typescript gods?

scottlovegrove (approved):

src/tools/find-tasks.ts
gnapse: It could be seen like that, but technically that's not needed because typescript does not complain even if the type appears to be `string | undefined`, because interpolating an `undefined` value into a string is not a type error.

However, your comment and my looking into it made me realize that we do not need this line at all. The `searchText` is already conveyed in the `appliedFilters` part of the structured content. So I removed it.

docs/tool-design.md
```diff
@@ -0,0 +1,68 @@
+# Tool Design Guidelines
+
+## Philosophy: Intelligent Consolidation
+
+We use **12 workflow-focused tools** instead of 30+ API-endpoint tools.
+
+**Traditional** (API-centric): `add-project`, `update-project`, `delete-project`, `get-project`...  
+**Our approach** (workflow-centric): `manage-projects` (handles create/update), `delete-object` (universal deletion)
```
goncalossilva, line 8: I am unsure about the real-world benefits of this design.

One major reason that split additions and modifications being split is considered good API design still applies if the workflow is LLM- or human-led. It makes the intent more clear and the behavior more explicit, which are essentially always properties of good design.

If we were saving tons of tools, I could be considered a small sacrifice for optimization, but we’re saving 3 tools total by my count (for tasks it's already split), so there is no meaningful difference.

In the end, this saves one tool by leaning into implicit behavior based on the presence of a parameter (and [related complexity](https://github.com/Doist/todoist-ai/pull/53/files#diff-b6c0415e65a3945421f96e20f735c46715de8a3b9ae7119849bbdb9d5f8d25f7R21-R26)), which doesn't seem like a net benefit to me.

docs/tool-design.md
```diff
@@ -0,0 +1,68 @@
...
+    structuredContent: { data, metadata }
+})
+```
+
+## Key Design Decisions
+
+| Choice | Rationale |
+|--------|-----------|
+| `manage-projects` (create/update) vs separate tools | Users don't distinguish "add" vs "update" in natural language |
+| `delete-object` (universal) vs type-specific tools | Deletion logic similar across types, reduces LLM cognitive load |
```
goncalossilva, line 40: I like this! I wonder if this practice could be expanded to updates via a universal `update-object` tool (so only additions would be per-object following a simple `id` /  `property` combo), or if it's overdoing it.

goncalossilva (commented):
Changes look great overall. 👌

What I'm most unsure about is the design around `manage-X`, as I think there are more downsides than upsides in this design, not to mention LLM performance (which we don't know; we'd have to test, and that's not the point of my feedback 😄). I expanded in a line comment.

One other surprise was the renaming of the files. Yesterday, when we discussed this, you made a point about related files being close to each other (`tasks-*.ts`) which made sense to me. That said, I don't have a preference here, so albeit unexpected, this is absolutely fine.

docs/tool-design.md
gnapse: That's a fair concern. I'm totally open to revisit this. Will keep it in mind for a future instance where we already have some other breaking change lined up, to batch them.

docs/tool-design.md
gnapse: Indeed, I do stand behind this one. Otherwise, we'd have one deletion tool per object type.

> I wonder if this practice could be expanded to updates via a universal update-object tool 

I'm not so sure. The args schema of such a tool would widely differ depending on the type of object being updated, and I'm hesitant to have a complex conditional union-type thing in the args. I think that would elicit more confusion from the language model than the unification of `update` and `add` under a single `manage` tool.

goncalossilva: You're right. It would get confusing fast. Explicit > implicit.

docs/tool-design.md
goncalossilva: Awesome, I'm glad it resonated. Let's perhaps batch it with adding `search` and `fetch` for OpenAI?

gnapse: > adding `search` and `fetch` for OpenAI?

That's not a breaking change. But yeah, we can do that anyway. If no other breaking change comes up in the horizon of the next week or two, we'll do it anyway.

A bit more context on why perhaps we should not do it (I'm convinced already of doing it, so take this as a devil's advocate kind of argument):

- Even if it's a couple of tools in the total count being saved, we still have lots of tools to add in the future: support for
    - managing comments (which I have in the works because it's a top priority)
    - managing labels (could benefit from the `manage-*` naming approach)
    - managing workspaces and folders (these could benefit from the `manage-*` naming approach)
    - read activity log and productivity info
    - _(and I'm probably missing a thing or two here, but maybe not)_

That already can add around 12 more tools. And that can vary depending on which of these could benefit from the `manage-*` approach, which could make that 12 count even larger. Putting us already around the 30 tools mark.

I guess the key question here is: 👉 **Is it a problem in practice for the LLM using this tools to confuse them?** (I guess that also depends on the quality of the model being used).

goncalossilva: > I'm convinced already of doing it, so take this as a devil's advocate kind of argument

😈 😄

In terms of tools saved by collapsing add and update, I count 6: workspaces, folders, projects, tasks, comments, and labels. The rest don't have add or edit, so there is no difference. Also note that for object types where we support bulk adding, there is no saving (e.g., for tasks, we can have `add_tasks` and `update_task` instead of `add_tasks` and `manage_task`).

Between 30 and 36, I think it's less about the raw number of tools, and more about the clarity in their organization and descriptions!

Do note that bulk adding could make sense for other kinds of objects like projects, sections, comments or labels, so the practical savings may be in the 1-2 range. 😅

> 👉 Is it a problem in practice for the LLM using this tools to confuse them? (I guess that also depends on the quality of the model being used).

I don't think we can know this without testing it, but I think it's reasonable to assume that something text-based that's easier for humans to understand is _likely_ easier for LLMs as well, considering they are primarily trained on human material. Plus, humans will also occasionally read these. 😄 

So ultimately, I think there is very little real upside to this optimization and some potential downside.

gnapse: Good. Let's do it then. Another argument in favor is that I hope or even sure that these tool limitations will ease up over time.

Episode 12 Nbc GIF by Law & Order

PR #58, opened 2025-08-22
feat: add comment tools

### LINKED ISSUE #54: Add support for task and project comments
## Summary
Currently, our MCP server lacks support for managing comments on tasks and projects. This is a significant gap compared to the full Todoist API capabilities.

## Missing Functionality
- Adding comments to tasks and projects
- Retrieving comments (single, all, filtered by task/project)
- Updating existing comments
- Deleting comments

## Proposed Implementation
Following our intelligent consolidation approach, we could implement this with two tools:

1. **`manage-comments`** - Handle all comment operations (add/update/delete) with operation type parameter
2. **`find-comments`** - Search and retrieve comments with filters for task/project/date ranges

This maintains our philosophy of consolidating related functionality rather than creating separate tools for each API endpoint.

## Notes
The specific tool design may evolve during implementation to best fit our workflow-centric approach and user experience goals.

gnapse:
Closes #54 

## Short description

- New tools: `find-comments`, `add-comments`, `update-comments`
- Updated tool `delete-object` to support deleting comments

## PR Checklist


src/tools/update-comments.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ToolNames } from '../utils/tool-names.js'
+
+const { FIND_COMMENTS, DELETE_OBJECT } = ToolNames
+
+const ArgsSchema = {
+    id: z.string().min(1).describe('The ID of the comment to update.'),
+    content: z.string().min(1).describe('The new content for the comment.'),
+}
+
+const updateComments = {
```
goncalossilva, line 19: Should this be named `updateComment` instead (sans `s`), considering it operates on a single comment?

src/tools/add-comments.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (args.taskId && args.projectId) {
+            throw new Error('Cannot provide both taskId and projectId. Choose one.')
+        }
+
+        const commentArgs: AddCommentArgs = {
+            content: args.content,
+            ...(args.taskId ? { taskId: args.taskId } : { projectId: args.projectId }),
+        } as AddCommentArgs
+
+        const comment = await client.addComment(commentArgs)
```
goncalossilva: Should we support adding multiple comments, similar to `add-tasks.ts`?

src/tools/add-comments.ts
```diff
@@ -0,0 +1,81 @@
...
+    targetType,
+    targetId,
+}: {
+    comment: Comment
+    targetType: 'task' | 'project'
+    targetId: string
+}): string {
+    const hasAttachment = comment.fileAttachment !== null
+    const attachmentInfo = hasAttachment
+        ? ` • Attachment: ${comment.fileAttachment?.fileName || 'file'}`
```
goncalossilva: I noticed in other tools that we tend not to use symbols like • to include metadata, but English (e.g., " with attachment (...)"). Should we do that here, too?

goncalossilva (approved):
LGTM 🚀 

Left a few comments, mostly nits.

src/tools/update-comments.ts
gnapse: Yes, you're right. But what I'll do is change it to allow it to operate on multiple comments, which was the intention, and I did not notice or did not tell Claude well enough 😅

goncalossilva: Perfect, I agree that's ideal. 👌

src/tools/add-comments.ts
gnapse: Yes! it was the intention, but I did not notice. Maybe I assumed that the name was a good enough indication for claude, and then I missed this detail during my code review (I swear I review claude's code before submitting it, but I'm not doing a good job at it 😅)

src/tools/add-comments.ts
gnapse: Will look into that.

src/tools/add-comments.ts
goncalossilva: I know you do! Otherwise the tests wouldn't be as reasonable as they are. 🤣 

It's a small detail, I'm glad we caught it. 💪

PR #59, opened 2025-08-22
feat!: restructure management tools into specialized add/update tools

gnapse:
## Summary

Split management tools into specialized operations following verb-first naming:

- `manage-projects` → `add-projects` + `update-projects`
- `manage-sections` → `add-sections` + `update-sections`
- Enhanced project tools with `isFavorite` and `viewStyle` support
- Updated the `tool-design.md` documentation

Increased tool count to 14 (will be 17 overall when we also introduce comment tools from #58).

## Key Features

- **Batch Operations**: All tools support multiple entities per call
- **Parallel Execution**: Concurrent API calls for better performance
- **Enhanced Projects**: Full support for `isFavorite` and `viewStyle` attributes

## 🧪 Test Plan

Manually tested connecting to Claude Desktop:

- [x] All existing functionality preserved (add and update projects)
- [x] Add and update in bulk works with a single tool call, as expected
- [x] New project attributes work correctly

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,7 @@ This project is in its early stages. Expect more and/or better tools soon.
 
 Nevertheless, our goal is to provide a small set of tools that enable complete workflows, rather than just atomic actions, striking a balance between flexibility and efficiency for LLMs.
 
-### Design Philosophy
-
-Our tools follow an **intelligent consolidation** approach that reduces cognitive load on LLMs by providing fewer, more powerful tools that handle complete user workflows rather than requiring multiple API calls for common tasks.
-
-Key principles include workflow-centric design, smart parameter handling, batch operations, and context-aware responses with suggested next steps.
-
-For detailed design guidelines, examples, and development patterns, see [docs/tool-design.md](docs/tool-design.md).
+For our design philosophy, guidelines, and development patterns, see [docs/tool-design.md](docs/tool-design.md).
 
 ### Available Tools
 
--- a/docs/tool-design.md
+++ b/docs/tool-design.md
@@ -1,68 +1,69 @@
 # Tool Design Guidelines
 
-## Philosophy: Intelligent Consolidation
+## Philosophy: Specialized Workflow Tools
 
-We use **12 workflow-focused tools** instead of 30+ API-endpoint tools.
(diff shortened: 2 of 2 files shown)


src/tools/update-projects.ts
```diff
@@ -0,0 +1,115 @@
...
+    viewStyle: z.enum(['list', 'board', 'calendar']).optional().describe('The project view style.'),
+})
+
+type ProjectUpdate = z.infer<typeof ProjectUpdateSchema>
+
+const ArgsSchema = {
+    projects: z.array(ProjectUpdateSchema).min(1).describe('The projects to update.'),
+}
+
+const updateProjects = {
```
goncalossilva, line 23: Consider `updateProject` (singular) since this operates on a single project.

src/tools/update-sections.ts
```diff
@@ -0,0 +1,95 @@
...
+const SectionUpdateSchema = z.object({
+    id: z.string().min(1).describe('The ID of the section to update.'),
+    name: z.string().min(1).describe('The new name of the section.'),
+})
+
+const ArgsSchema = {
+    sections: z.array(SectionUpdateSchema).min(1).describe('The sections to update.'),
+}
+
+const updateSections = {
```
goncalossilva, line 19: Consider `updateSection` (singular) since this operates on a single section.

goncalossilva (approved):
🚀 🚀

src/tools/update-projects.ts
gnapse: No, it operates on multiple projects. Doesn't it?

src/tools/update-sections.ts
gnapse: Same. Am I missing something?

goncalossilva: You're not; I misread both of these, sorry!

gnapse: I now realize that the comments PR did have the opposite issue, and that could help in misleading you. Anyway, the intention is for most tools to handle operations in bulk in a single call.

I'm hesitant to do it for `delete-object` for now, given how sensitive it is to delete something, so I'd rather have the user keep the ability to review all delete operations individually if they so want to.

PR #62, opened 2025-08-22
feat: add todoist base url to dev setup, plus doc improvements

gnapse:
## Short description

- new (optional) environment variable `TODOIST_BASE_URL`
- minor documentation improvements

## PR Checklist

### DIFF
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -12,7 +12,10 @@ Update the `.env` file with your Todoist token:
 
 ```env
 TODOIST_API_KEY=your-key-goes-here
+TODOIST_BASE_URL=https://local.todoist.com/rest/v2
 ```
+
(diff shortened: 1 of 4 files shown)


docs/dev-setup.md
```diff
@@ -12,8 +12,11 @@ Update the `.env` file with your Todoist token:
 
 ```env
 TODOIST_API_KEY=your-key-goes-here
+TODOIST_BASE_URL=https://local.todoist.com/rest/v2
```
scottlovegrove, line 15: This is the wrong endpoint isn't it? It should still be pointing to `/api/v1`

gnapse: Oh, but it works with that one, I think. Anyway, yes. We should update it.

PR #69, opened 2025-08-25
feat: add label filtering to find-tasks tool

gnapse:
## Summary
- Add `label` parameter to find-tasks tool for filtering tasks by Todoist labels
- Support label-only filtering using getTasks API  
- Support label + text search using text search API with `& @label` syntax
- Support label + container filters (project/section/parent) combinations
- Add comprehensive test coverage with 26 test cases and snapshots

## Test plan
- [x] All existing tests pass
- [x] 6 new test cases covering label filtering scenarios
- [x] Manual testing verified with real API calls
- [x] Type checking passes
- [x] Code formatting and linting applied

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #71, opened 2025-08-27
feat: Add support for filtering by label

scottlovegrove:
# Pull Request

## Short description

A suggestion from @amix was that we should be able to filter task acquisition by labels. This adds support to the three main task tools such that labels can be provided, and it can be set whether a task should have any of the labels or all of them. 

## PR Checklist

-   [ ] Added tests for bugs / new features


gnapse:
We both ended up working on this in parallel 🤣 #69 

Will review yours tomorrow to see which one we keep and what improvements we can bring in from the other, if it makes sense.

src/tools/__tests__/find-tasks-by-date.test.ts
```diff
@@ -110,7 +110,7 @@ describe(`${FIND_TASKS_BY_DATE} tool`, () => {
             mockGetTasksByFilter.mockResolvedValue(mockResponse)
 
             const result = await findTasksByDate.execute(
-                { startDate: 'today', limit: 50, daysCount: 7 },
+                { startDate: 'today', limit: 50, daysCount: 7, labels: [], labelsOperator: 'or' },
```
amix: I wonder if we could set labels to `[]` by default and `labelsOperator` to `or`. This way, we wouldn't need to explicitly set them everywhere 🤔 A similar approach could work for `limit`.

src/tools/__tests__/find-tasks.test.ts
```diff
@@ -409,21 +453,239 @@ describe(`${FIND_TASKS} tool`, () => {
...
+            },
+        ])(
+            'should filter tasks by labels in text search: $name',
+            async ({ params, expectedQuery }) => {
+                const mockTasks = [
+                    createMappedTask({
+                        id: TEST_IDS.TASK_1,
+                        content: 'Task with work label',
+                        labels: ['work'],
+                    }),
```
amix, line 466: I wonder if we should ommit `@` for the params (`labels: ['@personal', '@shopping']`). 

Including @ doesn't seem consistent with how we specify them for `createMappedTask`, where they're marked without @ (`labels: ['work']`).

src/tools/find-completed-tasks.ts
```diff
@@ -42,18 +44,26 @@ const ArgsSchema = {
...
-                ? await client.getCompletedTasksByCompletionDate(rest)
-                : await client.getCompletedTasksByDueDate(rest)
+                ? await client.getCompletedTasksByCompletionDate({
+                      ...rest,
+                      ...(labelsFilter ? { filterQuery: labelsFilter, filterLang: 'en' } : {}),
+                  })
+                : await client.getCompletedTasksByDueDate({
+                      ...rest,
+                      ...(labelsFilter ? { filterQuery: labelsFilter, filterLang: 'en' } : {}),
+                  })
```
amix, line 66: Nit: `filterLang` could be set in `parameters` (with a default to `en`) instead of hardcoding it to `en` here.

scottlovegrove: I'm not going to make this change, because having `filterLang` as a parameter when we don't have a `filter` parameter is a bit weird, even if it's just for setting a default. Omitting the `filter` parameter was a suggestion by @gnapse a couple of weeks ago, due to the complexities of our filter language ([ref](https://github.com/Doist/todoist-ai/pull/8#discussion_r2271055961)).

src/tools/__tests__/find-tasks.test.ts
scottlovegrove: That's fair enough, I've updated it to handle both scenarios where labels have been provided with `@` and where they haven't been.

src/tools/__tests__/find-tasks-by-date.test.ts
scottlovegrove: Ignore previous comment (if you saw it), just thinking about it a little more...

scottlovegrove:
I've made the changes @amix suggested (most of them), but I won't merge this until @gnapse has taken a look as he has a similar PR.

src/tools/__tests__/find-tasks-by-date.test.ts
```diff
@@ -158,7 +158,7 @@ describe(`${FIND_TASKS_BY_DATE} tool`, () => {
             const mockResponse = { tasks, nextCursor: cursor }
             mockGetTasksByFilter.mockResolvedValue(mockResponse)
 
-            const result = await findTasksByDate.execute(params, mockTodoistApi)
+            const result = await findTasksByDate.execute({ ...params }, mockTodoistApi)
```
gnapse: Why is it needed to create a new object to pass as argument? Same concern for the other instance below.

src/tools/find-tasks.ts
gnapse, line 1: I can't comment in the relevant lines for this concern, so here it goes.

You need to update this condition in this file:

Image

It now needs to also see if labels was passed. Or else, if the tool is called with only labels, it will fall into this `if` condition and result in an error. Haven't tried it, but it looks like it, right?

The new condition should be something like this:

```js
        // Validate at least one filter is provided
        if (!searchText && !projectId && !sectionId && !parentId && (!labels || labels.length === 0)) {
            throw new Error(
                'At least one filter must be provided: labels, searchText, projectId, sectionId, or parentId',
            )
        }
```

If this is correct, then it also means that this usage with only labels is not part of the tests.

gnapse, line 1: Another instance on which I cannot comment in the relevant lines:

The `generateTextContent` function in this file is not changed. Which means it doesn't show label filters in the text output. I think this applies in general to all the tools changed in this PR. The text content output of the tools normally gives a summary of what was the listing filtered by.

Tip: I handled this in my PR, so a way to do this here would be to tell Claude Code to look into my PR only for that bit (how `generateTextContent` does it), and tell it to apply the same pattern to this PR in all the affected `generateTextContent` functions.

src/utils/labels.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const LABELS_OPERATORS = ['and', 'or'] as const
+type LabelsOperator = (typeof LABELS_OPERATORS)[number]
+
+export const LabelsSchema = {
+    labels: z
+        .string()
+        .array()
+        .optional()
+        .describe('The labels to filter the tasks by. Do not include the "@" symbol prefix.'),
```
gnapse: If you nicely support below to adapt if it is included, I think we can omit this extra instruction, right?


```suggestion
        .describe('The labels to filter the tasks by'),
```

src/utils/labels.ts
```diff
@@ -0,0 +1,27 @@
...
+    labelsOperator: z
```
gnapse, line 8: The `generateLabelsFilter` function parameter below is `labelOperator` but everywhere else it's `labelsOperator` (plural). I would lean to call the argument `labelOperator` everywhere (singular). But it can also be the change to plural, which is easier because it's updating the function below.

src/utils/labels.ts
```diff
@@ -0,0 +1,27 @@
...
+        .enum(LABELS_OPERATORS)
+        .optional()
+        .describe(
+            'The operator to use when filtering by labels. This will dictate whether a task has all labels, or some of them. Default is "or".',
+        ),
```
gnapse, line 13: You can also express the default via zod:


```suggestion
        .optional()
        .default('or')
        .describe(
            'The operator to use when filtering by labels. This will dictate whether a task has all labels, or some of them. Default is "or".',
        ),
```

src/tools/__tests__/find-tasks.test.ts
gnapse, line 1: A couple of edge test cases to be handled here:

- it's not testing when calling it with only labels as the filtering criteria
- it might be good to test that it handles labels with or without the `@` symbol (which I think it's a great feature BTW)

src/utils/labels.ts
scottlovegrove: You can, but this then ends up changing the type that's inferred in Typescript, it goes from `labelsOperator?: 'and' | 'or'` to `labelsOperator: 'and' | 'or'`, which means it's no longer optional when calling `execute()` and must be included. The reason for omitting the default here was to reduce how much redundant code in the tests was written.

PR #81, opened 2025-09-09
feat: add task assignment management

diskwarrior:
# Pull Request

Closes #...

## Short description

Add task assignment management capabilities to the Todoist MCP server, enabling users to assign tasks to project collaborators and manage assignments.

**New Assignment Tools**: Added `find-project-collaborators` and `manage-assignments` tools for discovering project members and performing bulk assignment operations.

**Enhanced Task Management**: Extended `add-tasks` and `update-tasks` tools to support assignment during creation and updates via `responsibleUid` parameter.

**Assignment Validation System**: Implemented validation for task assignments including project collaboration checks and user resolution.

**User Resolution**: Added user resolver utility to handle assignment by user ID, name, or email address.

**Task Data Enrichment**: Extended task mapping to include `responsibleUid` and `assignedByUid` fields for assignment visibility.

Key features include:
- Assign tasks during creation or update to project collaborators
- Discover project collaborators with search filtering
- Perform bulk assignment/unassignment operations with dry-run support
- Validate assignments against project permissions and collaboration status
- Handle user resolution via multiple identifiers (ID, name, email)
- Comprehensive error handling with helpful suggestions

This feature takes the MCP server from individual task management to collaborative project workflows, supporting team-based Todoist usage patterns.

## PR Checklist

### DIFF
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -74,7 +74,7 @@ async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi)
     } = task
 
     let taskArgs: AddTaskArgs = { ...otherTaskArgs, projectId, sectionId, parentId }
-        
+
     // Prevent assignment to tasks without sufficient project context
     if (!projectId && !sectionId && !parentId) {
         throw new Error(


src/tools/find-completed-tasks.ts
```diff
@@ -144,7 +144,7 @@ function generateTextContent({
         limit: args.limit,
         nextCursor: nextCursor ?? undefined,
         filterHints,
-        previewLines: previewTasks(tasks),
+        previewLines: previewTasks(tasks, Math.min(tasks.length, args.limit)),
```
scottlovegrove, line 147: What was the need for this change?

src/tools/find-project-collaborators.ts
```diff
@@ -0,0 +1,188 @@
...
+import { getToolOutput } from '../mcp-helpers.js'
+import type { TodoistTool } from '../todoist-tool.js'
+import { summarizeList } from '../utils/response-builders.js'
+import { ToolNames } from '../utils/tool-names.js'
+import { type ProjectCollaborator, userResolver } from '../utils/user-resolver.js'
+
+const { FIND_PROJECTS, ADD_TASKS, UPDATE_TASKS } = ToolNames
+
+const ArgsSchema = {
+    projectId: z.string().min(1).describe('The ID of the project to search sections in.'),
```
scottlovegrove: Not sure the description is correct here, as you're not looking for sections, but collaborators.

src/tools/find-project-collaborators.ts
```diff
@@ -0,0 +1,188 @@
...
+    description: 'Search for collaborators by name or other criteria in a project.',
+    parameters: ArgsSchema,
+    async execute(args, client) {
+        const { projectId, searchTerm } = args
+
+        // First, validate that the project exists and get basic info
+        let projectName = projectId
+        try {
+            const { results: projects } = await client.getProjects({})
+            const project = projects.find((p) => p.id === projectId)
```
scottlovegrove: We would be more effective to use `client.getProject(projectId)` here.

src/tools/find-tasks.ts
```diff
@@ -21,6 +22,10 @@ const ArgsSchema = {
     projectId: z.string().optional().describe('Find tasks in this project.'),
     sectionId: z.string().optional().describe('Find tasks in this section.'),
     parentId: z.string().optional().describe('Find subtasks of this parent task.'),
+    responsibleUid: z
```
scottlovegrove: Given this could be id, name, or email address, I would suggest renaming this from `responsibleUid` to just `responsibleUser` or `assignee` (and I'm torn on which to go for, because `responsibleUser` is more in-keeping with our API naming, but `assignee` makes more readable sense).

src/tools/find-tasks.ts
```diff
@@ -76,47 +97,120 @@ const findTasks = {
...
+                    break
+                }
+
+                // Add additional exit condition if no matches found in batch and results are small
+                if (batchFilteredTasks.length === 0 && results.length < batchSize) {
+                    // No more potential matches, exit early
+                    break
+                }
+
+                currentCursor = nextCursor
```
scottlovegrove: A lot of this could be refactored out by using `client.getTasksByFilter()`

```ts
const tasks = await client.getTasksByFilter({
  query: `assigned to: ${assigneeDisplayName}`,
  lang: 'en',
  ...cursorAndLimitDetails
})
```

src/tools/find-tasks.ts
```diff
@@ -142,19 +236,26 @@ const findTasks = {
             limit: args.limit,
         })
 
+        // Always filter by assignee client-side (API doesn't reliably support assignee filtering)
```
scottlovegrove: Does this mean you've already tried using `getTasksByFilter`?

src/utils/assignment-validator.ts
```diff
@@ -0,0 +1,360 @@
...
+    PROJECT_NOT_SHARED: 'PROJECT_NOT_SHARED',
+    TASK_NOT_ACCESSIBLE: 'TASK_NOT_ACCESSIBLE',
+    PERMISSION_DENIED: 'PERMISSION_DENIED',
+    PROJECT_NOT_FOUND: 'PROJECT_NOT_FOUND',
+    TASK_NOT_FOUND: 'TASK_NOT_FOUND',
+} as const
+
+export type AssignmentErrorType = (typeof AssignmentErrorType)[keyof typeof AssignmentErrorType]
+
+export interface ValidationError {
```
scottlovegrove: As a code style, we do prefer types to interfaces (although I appreciate we probably don't have this stated anywhere).

src/utils/assignment-validator.ts
```diff
@@ -0,0 +1,360 @@
...
+                        "Try using the user's full email address",
+                    ],
+                },
+            }
+        }
+
+        // Validate the project exists and is accessible
+        try {
+            const { results: projects } = await client.getProjects({})
+            const targetProject = projects.find((p) => p.id === projectId)
```
scottlovegrove: Again, we can use `client.getProject` here

src/utils/assignment-validator.ts
```diff
@@ -0,0 +1,360 @@
...
+            inaccessibleTasks: number
+        }
+        recommendations: string[]
+    }> {
+        const recommendations: string[] = []
+        let canAssign = false
+
+        // Get project information
+        const { results: projects } = await client.getProjects({})
+        const project = projects.find((p) => p.id === projectId)
```
scottlovegrove: `client.getProject`

src/utils/user-resolver.ts
```diff
@@ -0,0 +1,242 @@
+import type { TodoistApi } from '@doist/todoist-api-typescript'
+
+export interface ResolvedUser {
```
scottlovegrove: Again, types preferred over interfaces (sorry).

src/utils/user-resolver.ts
```diff
@@ -0,0 +1,242 @@
...
+                    for (const collaborator of collaborators) {
+                        if (collaborator && !seenIds.has(collaborator.id)) {
+                            allCollaborators.push(collaborator)
+                            seenIds.add(collaborator.id)
+                        }
+                    }
+                } catch (_error) {
+                    // Skip failed project, continue with others
+                }
+            }
```
scottlovegrove: I wonder if this could be parellelised?

scottlovegrove (changes requested):

scottlovegrove:
Great first stab at this @diskwarrior, thank you for the PR.

src/tools/find-completed-tasks.ts
diskwarrior: I was seeing odd behavior when asking for things like list ALL tasks from userB. I added the second parameter to ensure the preview doesn't show more/less tasks than requested. I was defaulted to 5 task preview when requesting 10, for example. This tries to make the preview respect the limit preference. Does that make sense?

src/tools/find-project-collaborators.ts
diskwarrior: Fixed. Good catch!

src/tools/find-project-collaborators.ts
diskwarrior: Fixed.

src/tools/find-tasks.ts
diskwarrior: Yeah, I like responsibleUser better too. close call though.

src/tools/find-tasks.ts
diskwarrior: Fixed.

src/tools/find-tasks.ts
diskwarrior: Yes, I tested getTasksByFilter with combined assignee filters and it didn't return reliable results, so I only used it for assignee-only queries and filter client-side for combos.

src/utils/assignment-validator.ts
diskwarrior: yes, understood. fixed!

src/utils/assignment-validator.ts
diskwarrior: fixed

src/utils/assignment-validator.ts
diskwarrior: 👍

src/utils/user-resolver.ts
diskwarrior: nope, no apology required. fixed.

src/utils/user-resolver.ts
diskwarrior: fix attempted. have a look. thx!

diskwarrior:
Thanks for the quick and thorough review! I had a go at your suggestions and left a couple others open w/ comments. Also, have a look over the tests (not exciting) but I think I captured the spirit!

src/tools/find-completed-tasks.ts
scottlovegrove: I think so, yes. Was this something you could reproduce consistently? I'm wondering if this is a bug somewhere that we should look into.

src/tools/add-tasks.ts
```diff
@@ -25,6 +26,12 @@ const TaskSchema = z.object({
     projectId: z.string().optional().describe('The project ID to add this task to.'),
     sectionId: z.string().optional().describe('The section ID to add this task to.'),
     parentId: z.string().optional().describe('The parent task ID (for subtasks).'),
+    responsibleUid: z
```
scottlovegrove: Looks like I missed this one. If the intention is for this to be id, name or email, we should probably rename this one to `responsibleUser` too.

src/tools/add-tasks.ts
```diff
@@ -77,6 +92,58 @@ async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi)
         }
     }
 
+    // Handle assignment if provided
+    if (responsibleUid) {
+        // Prevent assignment to inbox tasks
+        if (!projectId && !sectionId && !parentId) {
+            throw new Error(
+                `Task "${task.content}": Cannot assign tasks in inbox. Please specify a project.`,
```
scottlovegrove: This error message isn't correct. In theory, you could try assigning to the inbox by providing the inbox project id. What I think the error message should be here is that not enough information has been provided in order to assign a task.

src/tools/find-tasks.ts
```diff
@@ -40,14 +45,15 @@ const ArgsSchema = {
 const findTasks = {
     name: ToolNames.FIND_TASKS,
     description:
-        'Find tasks by text search, or by project/section/parent container. At least one filter must be provided.',
+        'Find tasks by text search, or by project/section/parent container/assignee. At least one filter must be provided.',
```
scottlovegrove: Let's stay consistent and call it `responseible user`, it will match with the parameters then.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+const ArgsSchema = {
+    operation: z
+        .enum(['assign', 'unassign', 'reassign'])
+        .describe('The assignment operation to perform.'),
+    taskIds: z
+        .array(z.string())
+        .min(1)
+        .max(MAX_TASKS_PER_OPERATION)
+        .describe('The IDs of the tasks to operate on (max 50).'),
+    responsibleUid: z
```
scottlovegrove: I missed this file on my first pass, sorry.

Should this be responsibleUid, or responsibleUser? Are we allowing just ids here, or names/emails, too? We should probably include all for consistency.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+        .string()
+        .optional()
+        .describe('The user to assign tasks to. Required for assign and reassign operations.'),
+    fromAssigneeId: z
```
scottlovegrove: We should probably do the same here, id/name/email.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+            'For reassign operations: the current assignee to reassign from. Optional - if not provided, reassigns from any current assignee.',
+        ),
+    dryRun: z
+        .boolean()
+        .optional()
+        .default(false)
+        .describe('If true, validates operations without executing them.'),
+}
+
+export interface OperationResult {
```
scottlovegrove: types over interfaces.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+        // Fetch all tasks first to validate they exist and get project information
+        const tasks = await Promise.allSettled(
+            taskIds.map(async (taskId) => {
+                try {
+                    return await client.getTask(taskId)
+                } catch (_error) {
+                    throw new Error(`Task ${taskId} not found or not accessible`)
+                }
+            }),
+        )
```
scottlovegrove, line 76: Ugh, this is messy, but that's on us, not you. I might have a look at creating a `client.getTasks()` function in the future. Nothing to change here, thinking aloud.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+                    error: 'Task not accessible',
+                })
+            }
+        }
+
+        if (validTasks.length === 0) {
+            const textContent = generateTextContent({
+                operation,
+                results: taskErrors,
+                dryRun: true, // Show as dry run since nothing was executed
```
scottlovegrove: I don't think we should be setting `dryRun` to true here. We could end up in this code because things went wrong, which isn't really a dry run, but a very unhappy path.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+
+            return getToolOutput({
+                textContent,
+                structuredContent: {
+                    operation,
+                    results: taskErrors,
+                    totalRequested: taskIds.length,
+                    successful: 0,
+                    failed: taskErrors.length,
+                    dryRun: true,
```
scottlovegrove: Same applies to this.

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+            } else {
+                validationErrors.push({
+                    taskId: assignment?.taskId || 'unknown',
+                    success: false,
+                    error: validation?.error?.message || 'Validation failed',
+                })
+            }
+        }
+
+        if (dryRun) {
```
scottlovegrove: Could be worth just adding a comment to say what sort of assignment is left here, only because we're a long way down the function., and I know it's the only option left, but in reviewing this, I had to scroll all the way back up to confirm what was left 😄

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+                )
+                .map(({ assignment, validation }) => {
+                    const task = validTasks.find((t: Task) => t.id === assignment.taskId)
+                    return {
+                        taskId: assignment.taskId || 'unknown',
+                        success: true,
+                        originalAssigneeId: task?.responsibleUid || null,
+                        newAssigneeId: validation.resolvedUser?.userId || 'unknown',
+                    }
+                })
```
scottlovegrove: We effectively do both of these regardless of whether it's a dryrun or not. I wonder if there's a way we could optimise the code so we don't have the same code in two places. Makes it easier if something needs to change later on. I will accept it might only be the filtering, but it's a start :)

src/tools/manage-assignments.ts
```diff
@@ -0,0 +1,411 @@
...
+        const executePromises = validAssignments
+            .filter(
+                (item): item is { assignment: Assignment; validation: ValidationResult } =>
+                    item.assignment != null && item.validation != null,
+            )
+            .map(async ({ assignment, validation }): Promise<OperationResult> => {
+                const task = validTasks.find((t: Task) => t.id === assignment.taskId)
+
+                try {
+                    await client.updateTask(assignment.taskId || 'unknown', {
```
scottlovegrove: In what situation would `assignment.taskId` be empty/undefined/null that would cause `unknown` to be used? Should we just be filtering out those assignments?

src/tools/update-tasks.ts
```diff
@@ -34,6 +35,13 @@ const TasksUpdateSchema = z.object({
         .describe(
             'The duration of the task. Use format: "2h" (hours), "90m" (minutes), "2h30m" (combined), or "1.5h" (decimal hours). Max 24h.',
         ),
+    responsibleUid: z
```
scottlovegrove: Same comment here about `responsibleUser` change, for consistency. Don't forget to update the description.

src/tools/find-completed-tasks.ts
diskwarrior: As far as I can tell, this is an API behavior inconsistency with filter queries. Different API endpoints behave differently:      
                                                                                                              
Searching "all my tasks" uses `client.getTasks()` or general `client.getTasksByFilter()` with default API limit from `ApiLimits.TASKS_DEFAULT: 10`. You'd get 10+ tasks returned.     
                                                                                                              
Searching "all userB's tasks" uses `client.getTasksByFilter({ query: "assigned to: userB" })`. The Todoist API's filter query seems to have its own internal limits and was only returning 5 tasks for assignee queries, regardless of specified limit.                                                                                                                
                                                                                                                                                               
The change to `Math.min(tasks.length, args.limit)` should fix the issue. The fix ensures the filter shows what was actually returned by the API, not the hardcoded default. The default 5 is still useful for other contexts where you want a reasonable preview size regardless of user input (like operation summaries).

Thoughts?

diskwarrior:
SOrry, I dove straight at your comments and didn't really look around. I _think_ I was a bit more thorough here. Also found another optimization and moved user resolution from inside the loop to outside, eliminating redundant API calls when processing multiple tasks with the same `fromAssigneeUser`.  Glob commit...

src/tools/find-tasks.ts
diskwarrior: (think this is resolved now)

diskwarrior:
other interfaces outside the scope of this feature as preferred:

```
diff --git a/src/tools/get-overview.ts b/src/tools/get-overview.ts
index e9d4328..a7c93f1 100644
--- a/src/tools/get-overview.ts
+++ b/src/tools/get-overview.ts
@@ -151,7 +151,7 @@ type AccountOverviewStructured = Record<string, unknown> & {
     hasNestedProjects: boolean
 }

-interface ProjectOverviewStructured extends Record<string, unknown> {
+type ProjectOverviewStructured = Record<string, unknown> & {
     type: 'project_overview'
     project: {
         id: string
diff --git a/src/utils/duration-parser.ts b/src/utils/duration-parser.ts
index f56e185..fa6d7cb 100644
--- a/src/utils/duration-parser.ts
+++ b/src/utils/duration-parser.ts
@@ -10,7 +10,7 @@
  * - Supports optional spaces: "2h 30m"
  */

-interface ParsedDuration {
+type ParsedDuration = {
     minutes: number
 }

diff --git a/src/utils/response-builders.ts b/src/utils/response-builders.ts
index f9a152b..b728e64 100644
--- a/src/utils/response-builders.ts
+++ b/src/utils/response-builders.ts
@@ -162,7 +162,7 @@ export function previewTasks(tasks: TaskLike[], limit = 5): string {
     return lines
 }

-interface SummarizeListParams {
+type SummarizeListParams = {
     subject: string
     count: number
     limit?: number
```

src/tools/find-completed-tasks.ts
scottlovegrove: I'm happy to keep the workaround fix in place for now, but I'll make a note for myself to check this behaviour out as it doesn't feel right to me.

scottlovegrove:
> other interfaces outside the scope of this feature as preferred:

I've taken care of those in another PR 👍🏻

scottlovegrove:
Thanks for doing this PR @diskwarrior and thanks for sticking through the review.

PR #82, opened 2025-09-09
feat: Add user-info tool and upgrade Todoist API

amix:
## Short description

- Adds a new `user-info` tool to retrieve the session user's information, such as name, email, timezone, time, etc. 
- Upgrades @doist/todoist-api-typescript to latest version
- Includes test coverage for the new tool

## PR Checklist

### DIFF
--- a/src/tools/user-info.ts
+++ b/src/tools/user-info.ts
@@ -1,0 +1,168 @@
+import type { TodoistApi } from '@doist/todoist-api-typescript'
+import { getToolOutput } from '../mcp-helpers.js'
+import type { TodoistTool } from '../todoist-tool.js'
+import { ToolNames } from '../utils/tool-names.js'
+
+const ArgsSchema = {}
+
+type UserInfoStructured = Record<string, unknown> & {
+    type: 'user_info'
+    fullName: string
+    timezone: string
+    currentLocalTime: string
+    startDay: number
+    startDayName: string
+    weekStartDate: string
+    weekEndDate: string
+    currentWeekNumber: number
+    completedToday: number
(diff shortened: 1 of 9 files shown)


src/tools/user-info.ts
```diff
@@ -0,0 +1,168 @@
...
+
+function getUserPlan(user: { isPremium: boolean; businessAccountId?: string | null }): string {
+    if (user.businessAccountId) {
+        return 'Todoist Business'
+    }
+    if (user.isPremium) {
+        return 'Todoist Pro'
+    }
+    return 'Todoist Free'
+}
```
scottlovegrove: Rather than return a string, maybe we should have this as a string union

```ts
type UserPlan = 'Todoist Business' | 'Todoist Pro' | 'Todoist Free'
```

src/tools/user-info.ts
```diff
@@ -0,0 +1,168 @@
...
+    return days[index] ?? 'Unknown'
+}
+
+function formatDate(date: Date): string {
+    return date.toISOString().split('T')[0] ?? ''
+}
+
+function formatLocalTime(date: Date, timezone: string): string {
+    return date.toLocaleString('en-US', {
+        timeZone: timezone,
```
scottlovegrove: The one thing I'd say to be careful with is Todoist can return timezones that are not in the `Europe/London` format, and quite often those other timezones can be problematic when passing to formatting functions. Just something to double check doesn't break this code.

amix: Good point! I'll try this out, and also do a small test to fall-back to UTC if we can't resolve the timezone

PR #84, opened 2025-09-10
feat: Add user's id to user_info tool

amix:
## Context

A note to add the user's ID to the user_info tool. It's quite critical for any kind of filtering 😊


scottlovegrove:
👍🏻

PR #109, opened 2025-09-25
feat: Document the MCP setup better, including the remote endpoint

amix:
## Short description
Just a small PR to document our MCP setup better, including how to use it with popular code editors and via our remote endpoint.


scottlovegrove:
👍🏻

PR #121, opened 2025-09-30
fix: Enhance task filtering by responsible user and update related tools

scottlovegrove:
# Pull Request

## Short description

It has been noticed that when using one of the find task tools (`find-tasks`, `find-tasks-by-date`) that it's returning all tasks, regardless of who is assigned. This should behave like our clients in that if no assignedUser is provided to the tool, then a task should only be included in the result if it is unassigned, or assigned to the authenticated user. 

~In order to do this, I've updated the tool `execute` function to now include the authenticated user, this is acquired via the Todoist SDK's `getUser()` function and is run on each usage of a tool.~

## PR Checklist

-   [ ] Added tests for bugs / new features

### DIFF
--- a/src/tools/find-tasks.ts
+++ b/src/tools/find-tasks.ts
@@ -35,7 +35,7 @@ const ArgsSchema = {
         .enum(RESPONSIBLE_USER_FILTERING)
         .optional()
         .describe(
-            'How to filter by responsible user when responsibleUser is not provided. "assigned" = only tasks assigned to others; "unassignedOrMe" = only unassigned tasks or tasks assigned to me; "all" = all tasks regardless of assignment.',
+            'How to filter by responsible user when responsibleUser is not provided. "assigned" = only tasks assigned to others; "unassignedOrMe" = only unassigned tasks or tasks assigned to me; "all" = all tasks regardless of assignment. Default value will be `unassignedOrMe`.',
         ),
     limit: z
         .number()


amix:
For this (and `filterTasksByResponsibleUser`), could it make sense to have three options:

- Only return tasks that are either unassigned or assigned to the user (default)
- Only return tasks assigned to a user
- No filter applied

If we don't have this option, the MCP will not be able to fetch all tasks in a filter.

src/mcp-helpers.ts
```diff
@@ -74,7 +74,8 @@ function registerTool<Params extends z.ZodRawShape>(
         _context,
     ) => {
         try {
-            const result = await tool.execute(args as z.infer<z.ZodObject<Params>>, client)
+            const user = await client.getUser()
+            const result = await tool.execute(args as z.infer<z.ZodObject<Params>>, client, user)
```
gnapse: I wonder if we should let each tool call `.getUser()` if they need it, instead of doing it for all tools. Maybe it's not a big deal, but I wonder if making this request all the time, when most tools do not need it, is justified.

src/tool-helpers.ts
```diff
@@ -19,6 +19,29 @@ export function isWorkspaceProject(project: Project): project is WorkspaceProjec
...
+ * @param tasks - Array of tasks to filter (must have responsibleUid property)
+ * @param resolvedAssigneeId - The resolved assignee ID to filter by (optional)
+ * @param currentUserId - The current authenticated user's ID
+ * @returns Filtered array of tasks
+ */
+export function filterTasksByResponsibleUser<T extends { responsibleUid: string | null }>(
+    tasks: T[],
+    resolvedAssigneeId: string | undefined,
+    currentUserId: string,
+): T[] {
```
gnapse: nit: prefer named args instead of positional ones.

src/mcp-helpers.ts
scottlovegrove: Ok, I don't feel strongly about this, so will revert this change and let tools fetch it themselves.

scottlovegrove:
> For this (and `filterTasksByResponsibleUser`), could it make sense to have three options:
> 
> * Only return tasks that are either unassigned or assigned to the user (default)
> * Only return tasks assigned to a user
> * No filter applied
> 
> If we don't have this option, the MCP will not be able to fetch all tasks in a filter.

I'm a bit unsure how this would work. Could you maybe elaborate on it a little? Are you saying when there's a filter specified, we shouldn't perform the responsible user checks?

amix:
@scottlovegrove I am not 100% sure if it makes sense.

For the tool argument name, what about one of these:

* `assignee_filter` with values like `"unassigned_or_user"` (default), `"assigned_to_user"`, `"all"`

or

* `scope` with values like `"mine"` (default), `"assigned"`, `"all"`

If we don't have this, we don't have a way to not filter these results t🤔

scottlovegrove:
> For the tool argument name, what about one of these:
> 
> * `assignee_filter` with values like `"unassigned_or_user"` (default), `"assigned_to_user"`, `"all"`

We already have a `responsibleUser` argument on the tools, maybe we should making use of that instead? 

> If we don't have this, we don't have a way to not filter these results t🤔

Maybe it would help here if I had some examples of what you're wanting to filter out?

gnapse:
@scottlovegrove I think that what Amir points out is that we have these two options:

- filter for tasks assigned to a particular user
- filter tasks unassigned or assigned to me (default)

and we lack a third option:

- do not filter the results, return everything

scottlovegrove:
> and we lack a third option:
> 
> * do not filter the results, return everything

I see. I think I might address this outside of this PR as I can see a way of doing a bit of a refactor around this.

scottlovegrove:
> I see. I think I might address this outside of this PR as I can see a way of doing a bit of a refactor around this.

Ok, I thought this was all used in a number of places (hence the refactor suggestion), but it's not, so I'll do it in this PR 👍🏻

scottlovegrove:
@amix Does this satisfy your concern? [`3e73498` (#121)](https://github.com/Doist/todoist-ai/pull/121/commits/3e7349878b781241cba7bf492bd52890ff604db8)

src/tools/find-tasks.ts
```diff
@@ -26,6 +31,12 @@ const ArgsSchema = {
         .string()
         .optional()
         .describe('Find tasks assigned to this user. Can be a user ID, name, or email address.'),
+    responsibleUserFiltering: z
+        .enum(RESPONSIBLE_USER_FILTERING)
+        .optional()
+        .describe(
+            'How to filter by responsible user when responsibleUser is not provided. "assigned" = only tasks assigned to others; "unassignedOrMe" = only unassigned tasks or tasks assigned to me; "all" = all tasks regardless of assignment.',
+        ),
```
gnapse, line 39: Add the `.default()` which I think helps document to the LLM what the default option is.

scottlovegrove: I did originally have that, but when you have the `default()` option set, it changes the typescript type and makes it no longer optional, which meant having to put it in on all the tests, so I decided not to. I can do if you'd prefer though.

gnapse: Ah ok

scottlovegrove: But maybe I can add it to the description 👍🏻

PR #134, opened 2025-10-07
feat: add OpenAI-compliant `search` and `fetch` tools

gnapse:
## 🗺️ Overview

This PR implements OpenAI-compliant `search` and `fetch` tools for the Todoist MCP server, enabling integration with OpenAI's MCP protocol.

#### New Tools

- **`search`**: Searches across both tasks and projects, returning results with composite IDs (`task:{id}` or `project:{id}`), titles, and URLs
- **`fetch`**: Retrieves full content of a task or project by composite ID, including all metadata

#### Implementation Details

- Both tools follow the [OpenAI MCP specification](https://platform.openai.com/docs/mcp) exactly:
  - Return JSON-encoded strings in a single text content item
  - `search` returns array with `id`, `title`, `url` per result
  - `fetch` returns object with `id`, `title`, `text`, `url`, `metadata`
- Search executes parallel API calls to search tasks (via filter) and projects (client-side filtering)
- Composite IDs prevent collisions between task and project IDs
- Added `buildTodoistUrl()` helper function for consistent URL generation
- Comprehensive test coverage (30 new tests)

## PR Checklist

- [x] New tools added to `getMcpServer` AND exported in `src/index.ts`.


src/tools/search.ts
```diff
@@ -0,0 +1,75 @@
...
+        const { query } = args
+
+        // Search both tasks and projects in parallel
+        const [tasksResult, projectsResponse] = await Promise.all([
+            getTasksByFilter({
+                client,
+                query: `search: ${query}`,
+                limit: ApiLimits.TASKS_DEFAULT,
+                cursor: undefined,
+            }),
```
scottlovegrove: Nice!

src/utils/constants.ts
```diff
@@ -10,7 +10,7 @@ export const ApiLimits = {
     /** Default limit for task listings */
     TASKS_DEFAULT: 10,
     /** Maximum limit for task search and list operations */
-    TASKS_MAX: 50,
+    TASKS_MAX: 100,
```
goncalossilva, line 13: Why was the limit increased? Seems unrelated to this PR, and 100 is a lot. Is it a requirement from OpenAI?

goncalossilva (commented):
Changes LGTM though it doesn't feel right to expose them everywhere. Seeing as the MCP protocol allows for dynamic tools depending on the context, I think we should consider leveraging that and expose these tools to ChatGPT only, keeping our tight, well-designed set everywhere else.

src/utils/constants.ts
gnapse: This is tasks max, the maximum that can be requested per call in the find-tasks-* tools. I increased it because I saw that the projects max was already 100. But maybe we can lower both to 50 instead. It is not related to openai tools.

gnapse:
@goncalossilva about this:

> it doesn't feel right to expose them everywhere

The changes here is not yet what exposes the tools. It is in the private repo where we implement the web-service that exposes these tools via HTTP. That's where we need to selectively not include them if we choose not to. We'll discuss there once I open the PR to bring these tools to that web service.

goncalossilva:
@gnapse you're right, let's wait. I wanted to discuss this more thoroughly before we put effort into it. 🙏

PR #139, opened 2025-10-10
feat: Clarify content vs description usage and Markdown support

amix:
## Short description
Updated the `add-tasks` and `update-tasks` tools schema descriptions to provide clearer guidance on when to use `content` vs `description` fields.

## Changes

- **`content` field**: Now explicitly states it should be concise and actionable (the task name/title), with examples
- **`description` field**: Now clearly indicates it's for longer details and additional context
- Both fields now document Markdown support

### Why

The previous descriptions were minimal and didn't communicate:
- When to use `content` vs `description` 
- That task content should be concise, not verbose
- Both fields support Markdown formatting

This enhances the LLM's understanding of proper field usage, leading to more structured tasks (concise titles with detailed descriptions when necessary).

### Testing

- All existing tests pass (16/16)
- Build completes successfully
- No breaking changes to the API

### DIFF
--- a/src/tools/update-tasks.ts
+++ b/src/tools/update-tasks.ts
@@ -13,8 +13,18 @@ const { FIND_TASKS_BY_DATE, GET_OVERVIEW } = ToolNames
 
 const TasksUpdateSchema = z.object({
     id: z.string().min(1).describe('The ID of the task to update.'),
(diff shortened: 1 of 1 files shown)


scottlovegrove:
👍🏻

PR #140, opened 2025-10-10
feat: Add basic OTEL instrumentation to MCP server

goncalossilva:
This is a rough draft, not yet tested. I'm mostly looking for feedback on the high-level direction.

Is there a better way to go about this?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,7 @@
 TODOIST_API_KEY=your-key-goes-here
+
+# Optional: enable OpenTelemetry telemetry export (disabled when unset)
+# OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318/v1/traces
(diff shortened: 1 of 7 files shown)


gnapse:
@goncalossilva I feel that we should not do this here, in the open source tools implementation. Can we deal with this in the web server that exposes these tools via http MCP instead?

scottlovegrove:
> I feel that we should not do this here, in the open source tools implementation. Can we deal with this in the web server that exposes these tools via http MCP instead

Agreed, or we could end up logging users who have set it up locally 😄

goncalossilva:
Makes sense! But where exactly is that @gnapse? The MCP server is set up here.

> Agreed, or we could end up logging users who have set it up locally 😄

Not without the secrets being set up. 😅

gnapse:
@goncalossilva I created an internal issue and mentioned you with the answer. Let's close this for now.

PR #143, opened 2025-10-13
feat: Improve findCompletedTasks to handle user timezones

amix:
## Context
The find-completed-tasks tool doesn't work properly because it ignores time zones. This PR fixes this by:

- Adding support for converting user local dates to UTC in the findCompletedTasks tool
- Updating the execute method to accept 'since' and 'until' parameters, adjusting them based on the user's timezone
- I've also added a test to verify timezone handling, including a case for users in the Europe/Madrid timezone


scottlovegrove:
Are these changes not required for the task fetching tools?

amix:
@scottlovegrove I only had issues with the completed endpoint. For task fetching, we seem to be using Filterist, which uses Dateist to compute the dates (/date ranges).

PR #149, opened 2025-10-20
feat: extend responsibleUser filtering to more tools

amix:
## Context
Continuation of https://github.com/Doist/todoist-ai/pull/145. We're adding responsible user filtering to more core tools. This PR:

- Adds the `responsibleUser` parameter to `find-tasks` and `find-completed-tasks` (it was already in `find-tasks-by-date`)
- Switches user filtering to use email addresses instead of display names in Todoist filter queries (more robust as emails are unique)
- Updates `ResolvedUser` type to include the `email` field

### DIFF
--- a/src/tools/find-tasks-by-date.ts
+++ b/src/tools/find-tasks-by-date.ts
@@ -1,8 +1,14 @@
 import { addDays, formatISO } from 'date-fns'
 import { z } from 'zod'
+import {
+    appendToQuery,
+    buildResponsibleUserQueryFilter,
+    RESPONSIBLE_USER_FILTERING,
+    resolveResponsibleUser,
+} from '../filter-helpers.js'
 import { getToolOutput } from '../mcp-helpers.js'
 import type { TodoistTool } from '../todoist-tool.js'
-import { getTasksByFilter, RESPONSIBLE_USER_FILTERING } from '../tool-helpers.js'
+import { getTasksByFilter } from '../tool-helpers.js'
 import { ApiLimits } from '../utils/constants.js'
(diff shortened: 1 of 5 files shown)


goncalossilva:
> Switches user filtering to use email addresses instead of display names in Todoist filter queries (more robust as emails are unique)

@amix Why not both? This would match the UX of Todoist.

src/tools/find-tasks-by-date.ts
```diff
@@ -93,17 +113,24 @@ const findTasksByDate = {
...
+            const filteringMode = args.responsibleUserFiltering || 'unassignedOrMe'
+            if (filteringMode === 'unassignedOrMe') {
+                // Exclude tasks assigned to others (keeps unassigned + assigned to me)
+                if (query.length > 0) query += ' & '
+                query += '!assigned to: others'
+            } else if (filteringMode === 'assigned') {
+                // Only tasks assigned to others
+                if (query.length > 0) query += ' & '
+                query += 'assigned to: others'
+            }
```
scottlovegrove: This is identical to the code in `find-tasks-by-date`, we should extract this to a shared function.

amix:
> @amix Why not both? This would match the UX of Todoist.

The tool actually supports both at the tool level - we just use email when converting it to the API call. Here's how we describe the parameter:

> "Find tasks assigned to this user. Can be a user ID, name, or email address."

src/tools/find-tasks-by-date.ts
amix: Good input, @scottlovegrove. I've now refactored this and removed the duplicated code. I also moved some of the other duplicates 👍

PR #153, opened 2025-10-21
feat: Expose completion and modification metadata in task mapping

### LINKED ISSUE #152: Expose completion state in single-task fetch API
Hi Todoist team,

When I sync my weekly notes with Todoist through the API the agent rely on `todoist.fetch` to pull a single task by ID. After completing a task in the app, the fetch response still looks like this:

{
  "id": "task:xxxxxxxxxxxxxxxxxxxxxx",
  "title": "Example task title",
  "text": "Example task title\n\nDescription: Some extra context\nDue: 2025-10-20\nLabels: Example",
  "url": "https://app.todoist.com/app/task/xxxxxxxxxxxxxxxxxxxxxx",
  "metadata": {
    "priority": 4,
    "projectId": "projectId123",
    "sectionId": "sectionId123",
    "parentId": null,
    "recurring": false,
    "duration": null,
    "responsibleUid": null,
    "assignedByUid": null
  }
}

There is no `completed`, `is_done`, `completed_at`, or any other flag that tells the agent the task has been archived. To confirm completion the agent has to make a second call to `completed.get_all` (or the Sync API with the `completed_tasks` resource) and check whether the same ID appears there.

### Expected behaviour
It would be helpful if the single-task fetch response exposed completion state—for example via a boolean `completed` field and/or a timestamp `completed_at`. That would let the agent determine status without issuing a second API call.

### Why this matters
My automation updates documentation based on Todoist. A single fetch that returns both the task data and its completion state would simplify the integration and reduce API usage. At the moment the agent has to “double check” every task by hitting two different endpoints.

Thanks for considering this request!

scottlovegrove:
Add checked, completedAt, and updatedAt fields to mapTask output. These fields were available in the Todoist SDK but not exposed through our mapping layer.

This allows agents to:
- Determine task completion status without additional API calls
- Track when tasks were completed for productivity analysis
- Monitor task modification history

Fixes #152

## PR Checklist

### DIFF
--- a/src/tool-helpers.ts
+++ b/src/tool-helpers.ts
@@ -86,6 +86,9 @@ function mapTask(task: Task) {
         duration: task.duration ? formatDuration(task.duration.amount) : null,
         responsibleUid: task.responsibleUid,
         assignedByUid: task.assignedByUid,
+        checked: task.checked,
+        completedAt: task.completedAt,
+        updatedAt: task.updatedAt,
     }
 }
 
--- a/src/utils/test-helpers.ts
+++ b/src/utils/test-helpers.ts
@@ -19,6 +19,9 @@ export type MappedTask = {
     duration: string | null
     responsibleUid: string | null
     assignedByUid: string | null
+    checked: boolean
+    completedAt: string | null
+    updatedAt: string | null
 }
 
 /**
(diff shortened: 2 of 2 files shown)


scottlovegrove:
> In this case, the only one I'd question is `updatedAt`. Do we really need to expose `updatedAt` for all tasks in a task listing?

I guess maybe not. I only added it as an afterthought anyway, so happy for it to be removed.

Issue #45, opened 2025-08-18
feat: Enable activities-list tool once SDK supports activities endpoint


kyzzen:
## Enhancement description

Complete the implementation of the `activities-list` tool in the Todoist MCP server once the TypeScript SDK adds support for the `/api/v1/activities` endpoint. The tool structure and types are already in place as a placeholder - it just needs the execute function to be activated when the SDK dependency is updated.

The current placeholder implementation in `src/tools/activities-list.ts` is ready to be activated with minimal changes:
- Remove the error throw statement
- Uncomment the actual implementation code
- Update the tool description to remove the limitation notice

## The problem it solves

The activities-list tool will unlock powerful automation and analysis capabilities for Todoist users through AI assistants:

1. **Automated Daily Standups**
   - "What did I complete yesterday?" - Generate instant standup reports
   - "Show my team's progress this week" - Track team productivity
   - "What's been happening in the Marketing project?" - Project-specific updates

2. **Smart Task Recovery**
   - "What was the original due date for this task?" - Check historical changes
   - "Who moved this task to a different project?" - Audit trail for accountability
   - "Show me what changed before the issue occurred" - Debug workflow problems

3. **Productivity Insights**
   - "When am I most productive?" - Analyze completion patterns by hour/day
   - "How long do tasks typically stay in my inbox?" - Identify procrastination patterns
   - "Which projects are getting the most attention?" - Resource allocation analysis

4. **Context-Aware Assistance**
   - AI can understand task history to provide better suggestions
   - Detect patterns in how users work with certain types of tasks
   - Learn from past behavior to improve recommendations

5. **Team Collaboration Features**
   - "Who's been working on what?" - Team activity summaries
   - "Show changes made by John this week" - Individual contribution tracking
   - "What tasks were reassigned today?" - Workload monitoring

### Current Limitations

Without the activities-list tool, users must:

1. **Manually track changes** - No automated way to see what happened
2. **Use only completed tasks list** - Missing updates, additions, deletions, and other changes
3. **Lose historical context** - Cannot see how tasks evolved over time
4. **Miss collaboration insights** - No visibility into team member activities

### Temporary Workarounds

Currently, users can only:
- Use `tasks-list-completed` for partial activity data (completions only)
- Manually check individual tasks for current state (not historical)
- Export data periodically and compare manually

### Example 1: Daily Standup Generation
```typescript
// User prompt: "What did I accomplish yesterday?"
const result = await mcp.callTool('activities-list', {
[...]

PR #169, opened 2025-10-23
feat: Add find-activity tool

amix:
## Short description
Adds `find-activity` tool to retrieve activity logs from Todoist, enabling users to track who did what and when.

## Features
- Filter by object type (task/project/comment)
- Filter by event type (added/updated/deleted/completed/etc.)
- Filter by specific objects (objectId, projectId, taskId)
- Filter by initiator (user who performed the action)
- Cursor-based pagination for large result sets
- Extracts and displays actual content from events
- Shows all users' activity by default

## Limitations
- Date-based filtering not supported (API limitation)
- Shows recent events only (API doesn't specify exact time window)

## Testing
```bash
npm test -- src/tools/__tests__/find-activity.test.ts
```
All tests passing ✅

### DIFF
--- a/src/tools/find-activity.ts
+++ b/src/tools/find-activity.ts
@@ -1,0 +1,261 @@
+import { z } from 'zod'
+import { getToolOutput } from '../mcp-helpers.js'
+import type { TodoistTool } from '../todoist-tool.js'
+import { mapActivityEvent } from '../tool-helpers.js'
+import { ApiLimits } from '../utils/constants.js'
+import { summarizeList } from '../utils/response-builders.js'
+import { ToolNames } from '../utils/tool-names.js'
+
+const { FIND_TASKS, USER_INFO } = ToolNames
+
+const ArgsSchema = {
+    objectType: z
+        .enum(['task', 'project', 'comment'])
+        .optional()
+        .describe('Type of object to filter by.'),
+
+    objectId: z
+        .string()
+        .optional()
+        .describe('Filter by specific object ID (task, project, or comment).'),
+
+    eventType: z
+        .enum([
+            'added',
+            'updated',
+            'deleted',
+            'completed',
+            'uncompleted',
+            'archived',
+            'unarchived',
+            'shared',
+            'left',
+        ])
+        .optional()
+        .describe('Type of event to filter by.'),
+
+    projectId: z.string().optional().describe('Filter events by parent project ID.'),
+
+    taskId: z.string().optional().describe('Filter events by parent task ID (for subtask events).'),
+
+    initiatorId: z.string().optional().describe('Filter by the user ID who initiated the event.'),
+
+    limit: z
+        .number()
+        .int()
+        .min(1)
+        .max(ApiLimits.ACTIVITY_MAX)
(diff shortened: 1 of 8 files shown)


gnapse:
@amix can we say this closes #45?

amix:
@gnapse I think so. At least, it's a major jab at making it possible for the LLM to query Todoist's activity log 👌

src/utils/tool-names.ts
```diff
@@ -33,6 +33,9 @@ export const ToolNames = {
     FIND_PROJECT_COLLABORATORS: 'find-project-collaborators',
     MANAGE_ASSIGNMENTS: 'manage-assignments',
 
+    // Activity and audit tools
```
goncalossilva, line 36: Audit? :D

src/tools/find-activity.ts
```diff
@@ -0,0 +1,261 @@
...
+        .int()
+        .min(1)
+        .max(ApiLimits.ACTIVITY_MAX)
+        .default(ApiLimits.ACTIVITY_DEFAULT)
+        .describe('Maximum number of activity events to return.'),
+
+    cursor: z
+        .string()
+        .optional()
+        .describe('Pagination cursor for retrieving the next page of results.'),
```
goncalossilva, line 54: We briefly chatted about this on Twist, but I think we need to experiment with a different way of paginating that is more human-like (after all, LLMs are closer to human behavior than APIs).

I'm not exactly sure what we could play with, but the concept of a cursor may be too opaque, especially if the information that “there is more data” is a [metadata JSON entry like `"hasMore": true`](https://github.com/Doist/todoist-ai/blob/c3186f9ede9fb54834f53f300716308c2fa6e08f/src/mcp-helpers.ts#L51). I doubt it's immediately obvious to the LLM that there is, in fact, more and that it should use the cursor it received back to fetch more. That could explain why it doesn't perform as well as we'd like it to.

goncalossilva: /cc @gnapse and @scottlovegrove who usually have excellent ideas on how to design and optimize for LLMs.

src/utils/tool-names.ts
amix: 🚔

src/tools/find-activity.ts
amix: 💯. At least, I don't think Flash 2.5 knows how our pagination works.

gnapse: My earlier attempt at helping LLMs deal with this involved adding cues in the text response of tools about the fact that there's more info to fetch. See:

https://github.com/Doist/todoist-ai/blob/cf52c0237e37e644ab4c7a2ff1a57d0002d89e8b/src/utils/response-builders.ts#L192-L196

and also

https://github.com/Doist/todoist-ai/blob/cf52c0237e37e644ab4c7a2ff1a57d0002d89e8b/src/tools/__tests__/__snapshots__/find-tasks.test.ts.snap#L113

We should isolate if the issue happens only in our internal use of MCP servers inside Todoist, or if it's happening also when we use the MCP server conversationally in Claude Desktop or other MCP host apps.

PR #172, opened 2025-10-24
fix: Update todoist-typescript-sdk to 5.7.1 (follow-up)

amix:
## Short description

A small follow-up to force release-please to push a new version.

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,9 +45,9 @@
         "prepare": "husky"
     },
     "dependencies": {
-        "@doist/todoist-api-typescript": "5.7.1",
         "@modelcontextprotocol/sdk": "^1.11.1",
         "date-fns": "^4.1.0",
+        "@doist/todoist-api-typescript": "5.7.1",
         "dotenv": "^17.0.0",
         "zod": "^3.25.7"
     },


henningmu:
@gnapse can we have a better way than reordering dependencies to manually trigger a release, please? 🙌 😁

amix:
@henningmu we do have it, but it didn't trigger because I used `chore:` 😅

scottlovegrove:
> @gnapse can we have a better way than reordering dependencies to manually trigger a release, please? 🙌 😁

I was going to look at this this morning, cos it's getting tiresome anyway.

gnapse:
I'm not even sure what does it mean that we needed to reorder dependencies to trigger a release 🤔

Isn't it the problem that merging the release PRs were not properly publishing to npm? Which I think is what Scott is going to look into, or already did. Thanks BTW!

gnapse:
Ah, never mind, I now understand after having looked at the diff here and the fact that Amir used `chore` before.

Anyway, yeah, the problem is not what type of change we make, but that the publish to npm part was not being triggered properly.

scottlovegrove:
> but that the publish to npm part was not being triggered properly.

Which it is now 🚀

PR #181, opened 2025-10-25
fix: Remove nullable type from update_tasks to fix Gemini API compatibility

### LINKED ISSUE #179: Bug: update_tasks tool schema appears incompatible with Gemini API
When integrating @doist/todoist-ai tools with an agent using the Google Gemini API, adding the update_tasks tool to the list of available tools causes the agent to fail to respond.

If update_tasks is omitted from the toolset, other tools (like addTasks, findTasksByDate, etc.) work correctly, and the agent responds as expected. When update_tasks is included, any request (even one that doesn't intend to use the tool) results in no response from the agent. I was able to reproduce the bug across multiple MCP clients when using Gemini. I didn't encounter any issues when using OpenAI.

scottlovegrove:
## Summary

Fixes #179 - The `update_tasks` tool was incompatible with Google Gemini's function calling API due to the use of Zod's `.nullable()` modifier on the `responsibleUser` field.

### Changes Made

- **Removed `.nullable()` from schema**: Changed `responsibleUser` from `z.string().nullable().optional()` to `z.string().optional()`
- **Updated description**: Now instructs AI models to use `"unassign"` string value to remove task assignments
- **Maintained backward compatibility**: Runtime logic handles both `null` (existing code) and `"unassign"` (new recommended way)
- **Added test coverage**: New test for `"unassign"` value, kept backward compatibility test for `null`

### Root Cause

Zod's `.nullable()` generates JSON schemas with `["string", "null"]` format (OpenAPI 3.1 style), which causes Gemini's function calling API to fail with HTTP 400 errors or complete failure. This is a documented issue with Gemini's schema parser.

### Why This Fix Works

- Removes the problematic nullable type from the schema sent to Gemini
- Uses explicit string value `"unassign"` which is more self-documenting and aligns with the `manage-assignments` tool pattern
- Maintains full backward compatibility for existing integrations using `null`

## Test plan

- [x] All 329 existing tests pass
- [x] TypeScript type checking passes
- [x] New test verifies `"unassign"` string works correctly
- [x] Backward compatibility test verifies `null` still works
- [x] Pre-commit hooks (biome check, type-check) pass
- [ ] Manual verification with Gemini API (requires Gemini API access)

### DIFF
--- a/src/tools/__tests__/assignment-integration.test.ts
+++ b/src/tools/__tests__/assignment-integration.test.ts
@@ -230,13 +230,34 @@ describe('Assignment Integration Tests', () => {
             expect(extractTextContent(result)).toContain('Updated 1 task')
         })
 
-        it('should unassign task when responsibleUser is null', async () => {
+        it('should unassign task when responsibleUser is "unassign"', async () => {
             await updateTasks.execute(
                 {
                     tasks: [
                         {
                             id: 'task-123',
-                            responsibleUser: null,
+                            responsibleUser: 'unassign',
+                        },
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            expect(mockTodoistApi.updateTask).toHaveBeenCalledWith(
+                'task-123',
+                expect.objectContaining({
+                    assigneeId: null,
+                }),
+            )
+        })
+
+        it('should unassign task when responsibleUser is null (backward compatibility)', async () => {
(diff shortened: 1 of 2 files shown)


goncalossilva:
🚀

amix:
Thanks for the fix, @scottlovegrove 🎩-tip

PR #200, opened 2025-11-04
chore: migrate to vite and vitest

gnapse:
Migrates build and tests to vite. The motivation is to make it a bit easier to support React and JSX in a future where we need something for MCP-UI or ChatGPT apps. We may also make use of [vite's importing assets as strings](https://vite.dev/guide/assets.html#importing-asset-as-string) for HTML or CSS templates.

Most of the changes are in the tests, that change a bit from jest to vitest.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,6 +1,6 @@
 import { resolve } from 'node:path'
-import { defineConfig } from 'vite'
 import dts from 'vite-plugin-dts'
+import { defineConfig } from 'vitest/config'
 
 export default defineConfig({
     plugins: [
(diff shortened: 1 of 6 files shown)


vite.config.ts
```diff
@@ -0,0 +1,93 @@
...
+                // Generate declarations separately
+                preserveModules: false,
+            },
+        },
+        target: 'node18', // Target Node.js 18+
+        outDir: 'dist',
+        emptyOutDir: true,
+        sourcemap: false,
+        ssr: true, // Server-side rendering mode for Node.js
+        minify: false, // Keep readable for debugging
```
henningmu: Do we really want this? 🤔

gnapse: Nah, this config was generated by Claude Code. Will change it and look into the CI issue.

scottlovegrove:
No objections from me.

rfgamaral:
✅

PR #201, opened 2025-11-05
fix: Show next steps only for paginated tools and clarify as suggestions

amix:
## Content

- Removed next steps from non-paginated tools (add/update/delete/complete)
- Kept next steps for paginated find-* tools only when the cursor is set
- Changed "Next:" to "Possible suggested next step(s):" to clarify they're suggestions, not instructions

This prevents LLMs from treating suggestions as **mandatory actions** (which looks pretty problematic in an agentic loop).


gnapse:
Me too, but I'm ok with being pragmatic and aggressive in reducing the output. Let's see if this causes some backlash or not.

I may look into reducing the json output of listing tools too. I suspect it may be too much.

PR #216, opened 2025-11-13
fix: correct priority inversion for task outputs

### LINKED ISSUE #215: Claude Code : Adding Task "P1" vs reading tasks are inverted
Adding a task as a "P1" creates an urgent task, however when this is read by the MCP, it outputs the priorities in the opposite order (e.g. P1 = 4, P2 = 3, etc).  This is confusing Claude Code. 

Example Output 

> lets try adding a task.  Can you add a "test task" at the highest priority 

⏺ todoist - add-tasks (MCP)(tasks: [{"content":"test task","priority":"p1"}])
  ⎿  {                             
       "tasks": [
         {
     … +12 lines (ctrl+o to expand)

⏺ Aha! This is the smoking gun!

  I just created a task with "priority": "p1" (which should be highest priority), but the API returned it with "priority": 4.

  This means the numbering is INVERTED:
  - p1 (highest priority) = 4 in the data
  - p2 (high priority) = 3 in the data
  - p3 (medium priority) = 2 in the data
  - p4 (lowest priority) = 1 in the data

  So I've been reading all your priorities completely backwards!

scottlovegrove:
## Summary
Fixes a priority inversion issue where task outputs returned confusing priority values.

## Problem
The Todoist API uses inverted priority values (4=highest, 1=lowest), but the MCP server was returning raw API values without proper inversion, causing confusion:

- **Input**: User specifies `"priority": "p1"` for highest priority → API correctly receives `4` ✅
- **Output**: API returns `4` → User receives `4` (should be `1` for consistency) ❌
- **Documentation**: Schema incorrectly claimed "1 is highest priority" when returning `4` values

This created an asymmetric interface where input and output used different numbering systems.

## Solution
- Added `invertPriorityForOutput()` helper function to convert API values to user-friendly format
- Updated `mapTask()` to apply priority inversion for consistent output  
- Fixed schema documentation to correctly state "1=highest, 4=lowest"
- Updated `formatPriorityForDisplay()` to work with inverted values
- Maintained the numeric API contract (no breaking changes to return types)

## Result
Now provides **symmetric priority handling**:
- **Input**: `"priority": "p1"` → **Output**: `"priority": 1` (both represent highest priority)
- **Input**: `"priority": "p4"` → **Output**: `"priority": 4` (both represent lowest priority)
- **Display**: Shows as "P1", "P2", etc. in task summaries

## Test Plan
- [x] All existing tests pass with updated expectations
- [x] Priority conversion works correctly in both directions
- [x] Display formatting shows proper priority labels (P1, P2, P3, P4)
- [x] Schema documentation matches actual behavior

Closes #215

### DIFF
--- a/src/tool-helpers.test.ts
+++ b/src/tool-helpers.test.ts
@@ -34,7 +34,7 @@ describe('shared utilities', () => {
                 description: 'Test description',
                 dueDate: '2024-01-15',
                 recurring: false,
-                priority: 1,
+                priority: 4,
                 projectId: 'proj-1',
                 sectionId: undefined,
                 parentId: undefined,
--- a/src/tool-helpers.ts
+++ b/src/tool-helpers.ts
@@ -9,6 +9,7 @@ import type {
 } from '@doist/todoist-api-typescript'
 import z from 'zod'
 import { formatDuration } from './utils/duration-parser.js'
+import { invertPriorityForOutput } from './utils/priorities.js'
 
 // Re-export filter helpers for backward compatibility
 export {
@@ -81,7 +82,7 @@ function mapTask(task: Task) {
         dueDate: task.due?.date,
         recurring: task.due?.isRecurring && task.due.string ? task.due.string : false,
         deadlineDate: task.deadline?.date,
-        priority: task.priority,
+        priority: invertPriorityForOutput(task.priority),
         projectId: task.projectId,
         sectionId: task.sectionId ?? undefined,
         parentId: task.parentId ?? undefined,
--- a/src/tools/__tests__/__snapshots__/find-tasks-by-date.test.ts.snap
+++ b/src/tools/__tests__/__snapshots__/find-tasks-by-date.test.ts.snap
@@ -10,29 +10,29 @@ exports[`find-tasks-by-date tool > label filtering > should combine date filters
 "Tasks for 2025-08-15: 1 (limit 25).
 Filter: 2025-08-15; labels: @important.
 Preview:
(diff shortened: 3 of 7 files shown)


scottlovegrove:
> What are your thoughts on never using numbers when expressing the user-facing value of priorities, but to use the `p1`, `p2` notation instead? That way there's no ambiguity: if we ever see "p1" we know is the highest priority, if we see 1 we know is the lowest priority.

Interestingly, the first iteration of this fix was doing just that, converting numbers to their `p1`/`p2`/etc form. I opted to keep it as a number purely because this was a small fix and I didn't want to update the API contract with the MCP server. 

I have no problem with us doing that in a future update, I just didn't want to do it for this fix.

gnapse:
Ok, at least that validates that the idea makes sense.

Also, 

> I didn't want to update the API contract with the MCP server.

~~Unlike with regular APIs that are integrated into external systems in hardcoded ways, this one is about output that will be consumed by a flexible language model that'll be able to pick up the new way in which we convey data. Right? Or am I missing something?~~

~~If that's right, then this would not be a breaking change, technically speaking. Language models receiving priority data from our tools would still understand it. Presumably even better with the new notation.~~

---

Update:

As I wrote the text above, I realized that this would be a breaking change for systems that use our tools directly, such as our own internal use of AI agents generating code dynamically to use the tools directly, and relying on our tools' output schema. Any generated code using the old output schema would break.

So yeah, I was missing something. This would be a breaking change. But I think we should make it anyway.

gnapse:
I gave it a try in https://github.com/Doist/todoist-ai/pull/217

PR #217, opened 2025-11-14
feat!: semantic priority values (P1, P2, P3, P4) instead of numeric

gnapse:
> [!WARNING]
> This is a breaking change.

## 🗺️ Overview

Two things:

- What the title says. Now our tools all return tasks data structures where the priority is always expressed as p1, p2, p3 or p4, and never as a number.
- In a 2nd commit there's an attempt to promote not using numeric literals for priority in the code, even in place where the number value is expected. So instead of saying `priority: 1` we'll now say `priority: convertPriorityToNumber('p4')` instead.

See https://github.com/Doist/todoist-ai/pull/216#pullrequestreview-3460766807 and the subsequent few comments below it.

## ✅ PR Checklist

## 🎥 Demo

https://github.com/user-attachments/assets/69401b79-cf08-427f-b752-d6a8ade96995

### DIFF
--- a/src/tool-helpers.test.ts
+++ b/src/tool-helpers.test.ts
@@ -1,4 +1,4 @@
-import type { PersonalProject, Task, WorkspaceProject } from '@doist/todoist-api-typescript'
+import type { PersonalProject, WorkspaceProject } from '@doist/todoist-api-typescript'
 import {
     createMoveTaskArgs,
     isPersonalProject,
@@ -6,20 +6,17 @@ import {
     mapProject,
     mapTask,
 } from './tool-helpers.js'
-import { convertPriorityToNumber } from './utils/priorities.js'
+import { createMockTask } from './utils/test-helpers.js'
 
 describe('shared utilities', () => {
     describe('mapTask', () => {
         it('should map a basic task correctly', () => {
-            const mockTask = {
+            const mockTask = createMockTask({
                 id: '123',
                 content: 'Test task',
                 description: 'Test description',
                 projectId: 'proj-1',
-                sectionId: null,
-                parentId: null,
                 labels: ['work'],
-                priority: convertPriorityToNumber('p4'),
                 due: {
                     date: '2024-01-15',
                     isRecurring: false,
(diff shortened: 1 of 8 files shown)


src/tool-helpers.test.ts
```diff
@@ -18,7 +19,7 @@ describe('shared utilities', () => {
                 sectionId: null,
                 parentId: null,
                 labels: ['work'],
-                priority: 1,
+                priority: convertPriorityToNumber('p4'),
```
scottlovegrove: Why do we need to do these function calls for test objects? This applies to all our tests in this PR.

Side note: The fact we do this so often in this file suggests we could probably create a reusable task for testing.

scottlovegrove (approved):

scottlovegrove:
> * In a 2nd commit there's an attempt to promote not using numeric literals for priority in the code, even in place where the number value is expected. So instead of saying `priority: 1` we'll now say `priority: convertPriorityToNumber('p4')` instead.

Is this to try and ensure we don't break it again somehow?

gnapse:
> > * In a 2nd commit there's an attempt to promote not using numeric literals for priority in the code, even in place where the number value is expected. So instead of saying `priority: 1` we'll now say `priority: convertPriorityToNumber('p4')` instead.
> 
> Is this to try and ensure we don't break it again somehow?

It's for us to be able to always think in terms of the user-facing priority value, even when writing code that requires the under-the-hood numeric values.

src/tool-helpers.test.ts
gnapse: Good point. I'll look into it.

scottlovegrove:
> It's for us to be able to always think in terms of the user-facing priority value, even when writing code that requires the under-the-hood numeric values.

Makes sense

src/tool-helpers.test.ts
gnapse: Done ✅

amix:
👌 makes sense. This also aligns with the idea that MCP should mirror our UI, not our API.

PR #222, opened 2025-11-18
feat: add read-only and destructive hints to all tools

Schnouki:
# Pull Request

This is based on a feature request that a user send to the CX team.

## Short description

This adds a mutability field to the tool metadata, which is then translated into `readOnlyHint` and `destructiveHint` in the MCP server. These annotations allow clients to offer a better UX.

Spec: https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations

## PR Checklist

### DIFF
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -1,7 +1,7 @@
 import type { TodoistApi } from '@doist/todoist-api-typescript'
 import type { McpServer, ToolCallback } from '@modelcontextprotocol/sdk/server/mcp.js'
 import type { ZodTypeAny, z } from 'zod'
-import type { TodoistTool } from './todoist-tool.js'
+import type { TodoistTool, ToolMutability } from './todoist-tool.js'
 import { removeNullFields } from './utils/sanitize-data.js'
 
 /**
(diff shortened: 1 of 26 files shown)


scottlovegrove:
YES! Awesome @Schnouki, this has been on my list of things to sort out for a while and not been able to get going on it.

Schnouki:
Sure, this is something we can consider.

I think every read-write tool would have `destructiveHint` except `add-*` (since they can always overwrite changes made by another client), and no command would be idempotent (since state can always be altered between two calls by another client). So, we could actually have a simpler type:

```ts
type ToolMutability = 'readonly' | 'additive' | 'mutating'
```
WDYT?

gnapse:
> Sure, this is something we can consider.
> 
> I think every read-write tool would have `destructiveHint` except `add-*` (since they can always overwrite changes made by another client), and no command would be idempotent (since state can always be altered between two calls by another client). So, we could actually have a simpler type:
> 
> ```ts
> type ToolMutability = 'readonly' | 'additive' | 'mutating'
> ```
> 
> WDYT?

Yes, I agree. I wonder if "complete tasks" is idempotent. Or maybe it fails with an error when called on completed tasks? But those are implementation details. Do what makes sense. Maybe we do not have idempotent tools anyway.

Schnouki:
> I wonder if "complete tasks" is idempotent.

It's not: 
- a task can be uncompleted in another client, then completed again, causing side effects that make this not idempotent (for example: webhooks, activity log, notifications to the assignee, etc.)
- if the task is recurring, completing it will [schedule it to its next occurence](https://developer.todoist.com/api/v1/#tag/Tasks/operation/close_task_api_v1_tasks__task_id__close_post), and possibly uncomplete sub-tasks: definitely not idempotent.

It should be a no-op on already completed tasks, though.

Will implement this and ask for another round of review :)

scottlovegrove:
This has been published and deployed to production.

scottlovegrove:
image

Schnouki:
In Claude Web: 
image

PR #253, opened 2025-12-19
feat: Add fetch-object tool for direct object retrieval by ID

scottlovegrove:
## Summary

Add a new `fetch-object` tool that allows fetching individual Todoist objects (tasks, projects, comments, sections) directly by their ID. This complements existing `find-*` tools by providing a simpler interface when the object ID is already known.

## Why a new tool instead of extending `fetch`?

The existing `fetch` tool is designed for OpenAI MCP spec compliance:
- Uses composite ID format (`"task:123"`, `"project:456"`)
- Returns MCP document format (id, title, text, url, metadata)
- Only supports tasks and projects (MCP spec limitation)
- Optimized for OpenAI MCP consumers

The new `fetch-object` tool provides a cleaner, more flexible API:
- Separate `type` and `id` parameters (more intuitive)
- Returns full typed objects with all fields
- Supports all object types including comments and sections
- Better for general programmatic use by AI agents

Keeping them separate maintains MCP spec compliance for `fetch` while providing a more powerful general-purpose tool for direct object retrieval.

## Changes

- **New tool**: 
  - Supports fetching tasks, projects, comments, and sections by ID
  - Returns both human-readable text content and structured typed data
  - Follows existing tool patterns and conventions

## Use Case

This tool is particularly useful for AI agents that:
- Have an object ID from a previous operation
- Want to retrieve full details without complex search operations
- Need direct ID-based retrieval rather than filtered searches

## Test Results

✅ All 363 tests pass  
✅ Type-check: No errors  
✅ Linting: All checks pass  
✅ Schema validation: Gemini API compatible  

### DIFF
--- a/src/tools/__tests__/fetch-object.test.ts
+++ b/src/tools/__tests__/fetch-object.test.ts
@@ -9,7 +9,7 @@ const mockTodoistApi = {
     getTask: vi.fn(),
     getProject: vi.fn(),
     getComment: vi.fn(),
-    getSections: vi.fn(),
+    getSection: vi.fn(),
 } as unknown as Mocked<TodoistApi>
 
 const { FETCH_OBJECT } = ToolNames
@@ -28,14 +28,6 @@ const MOCK_SECTION: Section = {
     isDeleted: false,
     isCollapsed: false,
     url: 'https://todoist.com/sections/section123',
-}
-
-const MOCK_SECTION_2: Section = {
-    ...MOCK_SECTION,
-    id: 'section456',
-    name: 'Other Section',
-    sectionOrder: 2,
-    url: 'https://todoist.com/sections/section456',
 }
 
 function createMockComment(overrides: Partial<Comment> = {}): Comment {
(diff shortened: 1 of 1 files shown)


henningmu:
@scottlovegrove @gnapse what are your thoughts on making this plural `fetch-objects` and accepting an array of composite IDs?

One use case is when using `find-tasks-by-date` we have a list of tasks with just project IDs. Now we need to call `fetch-object` for each project ID separately. Would be much simpler to just dump all IDs into a single tool call 🤔

scottlovegrove:
The problem is the Todoist api not being built to do that. We'd have to either revert to using the sync api for this tool, or we'd potentially have to put in delays to avoid being rate limited.

henningmu:
I thought it might be beneficial to hide the complexity of multiple API calls in the tool is better than pushing it on the consumer of the tools. But you're right it makes a lot of things more complex, e.g. partial failure scenarios :/

PR #275, opened 2026-01-20
chore: add run-tool script for direct tool execution

gnapse:
## Summary
- Adds `scripts/run-tool.ts` for executing any MCP tool directly without the server
- Moves `CLAUDE.md` to repo root (was incorrectly at `.claude/claude.md`)
- Documents the script in `CLAUDE.md`

## Usage
```bash
npx tsx scripts/run-tool.ts <tool-name> '<json-args>'
npx tsx scripts/run-tool.ts --list  # list all tools with descriptions
```

Examples:
```bash
npx tsx scripts/run-tool.ts add-tasks '{"tasks":[{"content":"Test task","order":1}]}'
npx tsx scripts/run-tool.ts find-tasks '{"query":"meeting"}'
npx tsx scripts/run-tool.ts get-overview '{}'
```

Requires `TODOIST_API_KEY` in `.env` (and optionally `TODOIST_BASE_URL`).

## Test plan
- [x] Script lists all tools with `--list`
- [x] Script executes tools against staging API
- [x] All 362 tests pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -49,3 +49,21 @@ When adding new tool features:
 2. Update `src/mcp-server.ts` tool usage guidelines
 3. Add tests demonstrating the feature
 4. Include examples in descriptions where helpful
+
+## Running Tools Directly
+
+Use `scripts/run-tool.ts` to execute any tool without the MCP server:
+
+```bash
+npx tsx scripts/run-tool.ts <tool-name> '<json-args>'
+npx tsx scripts/run-tool.ts --list  # list all tools
+```
+
+Examples:
+```bash
+npx tsx scripts/run-tool.ts add-tasks '{"tasks":[{"content":"Test task"}]}'
+npx tsx scripts/run-tool.ts find-tasks '{"query":"meeting"}'
+npx tsx scripts/run-tool.ts get-overview '{}'
+```
+
+Requires `TODOIST_API_KEY` in `.env` (and optionally `TODOIST_BASE_URL`).
--- a/scripts/run-tool.ts
+++ b/scripts/run-tool.ts
@@ -1,0 +1,168 @@
+#!/usr/bin/env npx tsx
+/**
+ * Run any Todoist tool directly without going through MCP.
+ *
+ * Usage:
+ *   npx tsx scripts/run-tool.ts <tool-name> '<json-args>'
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -49,3 +49,21 @@ When adding new tool features:
 2. Update `src/mcp-server.ts` tool usage guidelines
 3. Add tests demonstrating the feature
 4. Include examples in descriptions where helpful
+
+## Running Tools Directly
+
+Use `scripts/run-tool.ts` to execute any tool without the MCP server:
+
+```bash
+npx tsx scripts/run-tool.ts <tool-name> '<json-args>'
```
scottlovegrove, line 58: I wonder if we could make the DX a bit better so it could just be something like `npm run run-tool <tool-name> <json-args>`

Would mean it's available for any terminals that offer autocomplete on jpm scripts.

gnapse: I thought about it, but: (a) this is mostly for AI agents to do it, and (B) the `npm` command would need a `--` to pass args:

```
npm run run-tool -- <tool-name> <json-args>
```

Not horrible. Maybe better. But I'm always forgetting this for the few npm commands that have that issue.

scottlovegrove: > the `npm` command would need a `--` to pass args

You can include it in the script itself 🤷🏻‍♂️ 

Anyway, not a blocker, was just a thought.

gnapse: Not sure what you mean with "you can include it in the script itself". Any script you declare in package json that requires arguments provided by you dynamically, you need to do this notation. For instance:

```json
{
  "scripts": {
    "run-tool": "npx tsx scripts/run-tool.ts"
  }
}
```

then to call it:

```shell
# 👎 won't work
npm run run-tool tool-name 'args'

# 👍 works
npm run run-tool -- tool-name 'args'
```

scottlovegrove: I meant just do something like:

```json
{
  "scripts": {
    "run-tool": "npx tsx scripts/run-tool.ts -- "
  }
}
```

gnapse: I do not think that's how it works. It's not the `run-tool.ts` script the one who needs the `--`. It's the `npm` call.

If you do that, then `run-tool.ts` will receive the value "--" as the name of the tool to call.

scottlovegrove: Let's leave it then.

PR #278, opened 2026-01-20
feat: Add support for 'clientType', do not expose emails to chatgpt

henningmu:
## Short description

As discussed internally ([ref](https://twist.com/a/1585/ch/820139/t/7542998/)), I decided to go ahead and add conditionals for ChatGPT after all. I'll follow up with a PR in todoist-ai-integrations setting this `clientType`. For now this is an optional parameter that does not affect any other usage.

Should any consumer decide to set the `clientType` to `chatgpt`, they'll not receive collaborator emails in their tool responses. They can still use emails to assign tasks.

## PR Checklist

### DIFF
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -7,16 +7,21 @@ import { removeNullFields } from './utils/sanitize-data.js'
 import { ToolNames } from './utils/tool-names.js'
 
 /**
- * Configuration options for the MCP server that affect tool behavior.
+ * A feature that modifies tool behavior.
  */
-type ServerConfig = {
+type Feature = {
     /**
-     * The type of client connecting to the server.
-     * - 'chatgpt': ChatGPT clients - emails are stripped from collaborator tool outputs for privacy.
-     * - 'unknown': Default behavior, no special handling.
+     * The feature name. Supported features:
+     * - 'strip_emails': Strips email addresses from collaborator tool outputs
+     *   (affects: find-project-collaborators, find-completed-tasks)
      */
-    clientType?: 'chatgpt' | 'unknown'
+    name: string
 }
+
+/**
+ * Array of features to enable.
+ */
+type Features = Feature[]
 
 type McpTextResource = {
     name: string
@@ -113,6 +118,8 @@ function addMetaToTool<Params extends z.ZodRawShape, Output extends z.ZodRawShap
 
 /**
  * Tools that expose user emails in their outputs.
+ * When adding new tools that return user emails, update this list
+ * and the FeatureFlags.stripEmails JSDoc.
  */
 const TOOLS_WITH_USER_EMAILS = [
     ToolNames.FIND_PROJECT_COLLABORATORS,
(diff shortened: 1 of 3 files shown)


src/mcp-helpers.ts
```diff
@@ -98,24 +111,91 @@ function addMetaToTool<Params extends z.ZodRawShape, Output extends z.ZodRawShap
     }
 }
 
+/**
+ * Tools that expose user emails in their outputs.
+ */
+const TOOLS_WITH_USER_EMAILS = [
+    ToolNames.FIND_PROJECT_COLLABORATORS,
+    ToolNames.FIND_COMPLETED_TASKS,
+] as const
```
gnapse, line 127: This implies that we now have the potential to expand email presence in some tools without realizing that we need to update this list. Not sure how to deal with that.

scottlovegrove (approved):
I wonder if we should add something to the description that says "if you are definitely ChatGPT, please say so" 😄

I see we are doing something server side, which is fine if using the remote MCP, but won't be useful if it's a locally running MCP server.

src/mcp-server.ts
```diff
@@ -107,9 +107,18 @@ Always provide clear, actionable task titles and descriptions. Use the overview
...
  */
-function getMcpServer({ todoistApiKey, baseUrl }: { todoistApiKey: string; baseUrl?: string }) {
+function getMcpServer({
+    todoistApiKey,
+    baseUrl,
+    clientType,
+}: {
+    todoistApiKey: string
+    baseUrl?: string
+    clientType?: 'chatgpt' | 'unknown'
```
gnapse: I wonder if instead of making the config be about what client is connected, it should be about what features are enabled. Something like:

```typescript
type FeatureFlags = {
  stripEmails?: boolean
}

type McpServerParams = {
  todoistApiKey: string
  baseUrl?: string
  featureFlags?: FeatureFlags
}

function getMcpServer({ todoistApiKey, baseUrl, featureFlags = {} }) {
  // …
}
```

src/mcp-helpers.ts
```diff
@@ -98,24 +111,91 @@ function addMetaToTool<Params extends z.ZodRawShape, Output extends z.ZodRawShap
...
  * @param tool - The tool to register.
  * @param server - The server to register the tool on.
  * @param client - The Todoist API client to use to execute the tool.
+ * @param config - Server configuration options.
  */
 function registerTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape = z.ZodRawShape>(
     tool: TodoistTool<Params, Output>,
     server: McpServer,
     client: TodoistApi,
+    config: ServerConfig = {},
```
gnapse: nit: let's consider making `registerTool` receive named args instead of positional ones.

src/mcp-helpers.ts
henningmu: Yes, added a comment for now 👍

src/mcp-server.ts
henningmu: Good idea. Went with `features` instead of `featureFlags` in the end 👍

src/mcp-helpers.ts
henningmu: Agreed. done 👍

PR #289, opened 2026-01-26
feat(mcp): add ToolAnnotations hints

goncalossilva:
Adds MCP tool annotation hints to all registered tools.

- `title`: prefix with "Todoist:" to disambiguate generic names (e.g. search/fetch) in multi-server clients.
- `openWorldHint`: `false` for all tools since they only operate within the user's Todoist account.
- `idempotentHint`:
  - `true` for read-only tools and `delete-object` (repeat calls should have no additional effect)
  - `false` otherwise (additive tools duplicate on retry; mutating tools can have additional effects, e.g., recurring task completions)

Equivalent to https://github.com/Doist/twist-ai/pull/80

### DIFF
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -1,8 +1,8 @@
 import type { TodoistApi } from '@doist/todoist-api-typescript'
 import type { McpServer, ToolCallback } from '@modelcontextprotocol/sdk/server/mcp.js'
-import type { TextResourceContents } from '@modelcontextprotocol/sdk/types.js'
+import type { TextResourceContents, ToolAnnotations } from '@modelcontextprotocol/sdk/types.js'
 import type { z } from 'zod'
-import type { TodoistTool, ToolMutability } from './todoist-tool.js'
+import type { TodoistTool } from './todoist-tool.js'
 import { removeNullFields } from './utils/sanitize-data.js'
 import { ToolNames } from './utils/tool-names.js'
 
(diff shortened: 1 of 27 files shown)


src/todoist-tool.ts
```diff
@@ -59,6 +60,14 @@ type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape> = {
      */
     _meta?: Record<string, unknown>
 
+    /**
+     * Optional MCP ToolAnnotations overrides for this tool.
+     *
+     * Use this when a tool's behavior can't be fully captured by `mutability`,
+     * e.g. marking a mutating tool as idempotent.
+     */
+    mcpAnnotations?: Partial<ToolAnnotations>
```
goncalossilva: I wonder if it wouldn't be simpler to just drop `mutability` and have per-tool MCP annotations.

Explicit > implicit. :)

gnapse:
> `idempotentHint`:
>     * `true` for delete-object (repeat calls should have no additional effect)

Yes, it is idempotent, but subsequent calls won't return success. Just saying. Technically it is idempotent in the sense that it won't modify the data twice. But a weird kind of idempotent.

src/tools/delete-object.ts
```diff
@@ -27,6 +27,7 @@ const deleteObject = {
     parameters: ArgsSchema,
     outputSchema: OutputSchema,
     mutability: 'mutating' as const,
+    mcpAnnotations: { idempotentHint: true },
```
gnapse: We already had these annotations in `mutability`. Maybe we should drop `mutability` in favor of the new `mcpAnnotations` that you're adding here. Not sure what's the point on keeping them both.

goncalossilva:
> Yes, it is idempotent, but subsequent calls won't return success.

Are you sure, @gnapse? I traced our API code for 2 object types and in both cases, they do return success. That was the deciding factor for labeling it idempotent!

src/tools/delete-object.ts
goncalossilva: `mutability` does not capture all possible variations, unfortunately. Delete is the only tool that mutates _and_ is idempotent. We don't have a definition for that.

I do like the idea of dropping it, though. This `mcpAnnotations` dictionary follows `ToolAnnotations` from the official MCP SDK, [which I do prefer—it's more explicit](https://github.com/Doist/todoist-ai/pull/289#discussion_r2728963779), and it drops custom types.

src/mcp-helpers.ts
```diff
@@ -112,18 +112,46 @@ function getErrorOutput(error: string) {
...
+    return tool.mcpAnnotations ? { ...annotations, ...tool.mcpAnnotations } : annotations
+}
+
+function formatToolTitle(toolName: string): string {
+    return toolName
+        .split('-')
+        .filter(Boolean)
+        .map((segment) => `${segment.charAt(0).toUpperCase()}${segment.slice(1)}`)
+        .join(' ')
 }
```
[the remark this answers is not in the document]
goncalossilva: I'm unsure, this is fairly straightforward.

src/todoist-tool.ts
goncalossilva: Done in the latest `main`, after @gnapse suggested the same exact thing. We reuse `ToolAnnotations` from the MCP SDK almost directly, just making all entries required (so it's explicit for all tools).

src/tools/delete-object.ts
goncalossilva: Done in the latest `main`. We reuse `ToolAnnotations` from the MCP SDK almost directly, just making all entries required (so it's explicit for all tools).

gnapse:
> > Yes, it is idempotent, but subsequent calls won't return success.
> 
> Are you sure, @gnapse? I traced our API code for 2 object types and in both cases, they do return success. That was the deciding factor for labeling it idempotent!

Hmmm, is it the API or is it the MCP server? I do think we should return an error if we receive the command to delete an object and the object is not found. 🤔

src/tools/delete-object.ts
gnapse: Yes, agree. That was my point: that what you're introducing here is more complete and more explicit.

goncalossilva:
I only checked the API, and only 2 of these endpoints. But since the MCP server uses the SDK and the SDK uses the API, most likely the API behavior cascades up.

As for wether the API should error out or not, you're probably right, though I haven't looked up what's expected of DELETE in general.

goncalossilva:
It seems [this is more controversial than I assumed](https://stackoverflow.com/questions/4088350/is-rest-delete-really-idempotent), even if the server were to start raising 404 on subsequent deletes.

PR #297, opened 2026-02-03
fix: include current user in user resolution

scottlovegrove:
## Summary
- Fix user resolution to check the current authenticated user before searching collaborators
- This allows `responsibleUser` to be resolved by name/email even when the user has no shared projects

## Test plan
- [x] Added tests for resolving current user by exact full name
- [x] Added tests for resolving current user by email
- [x] Added tests for resolving current user by partial name
- [x] Added tests for resolving current user by user ID
- [x] Added test for error when user doesn't match and no collaborators exist
- [x] All 409 tests pass
- [x] Build passes


scottlovegrove:
> Am I right in understanding that we have to do this because `getAllCollaborators` does not include the current user?

That's my belief, yes. 

> If so, would it be better to prepend the current user to `allCollaborators` and reuse the rest of the logic, instead of repeating it?

That would make a lot of sense, yes. I'll update to this approach as that's a much more sensible idea. 

> When we're done resharding, we should immediately convert this to `getWorkspaceCollaborators` (or similar) and stop iterating projects/workspaces.

To be honest, it would be nice if there were actually more publicly documented workspace endpoints. Did you know, we don't even have one to list the workspaces to which a user belongs? It can only be fetched via Sync.

PR #315, opened 2026-02-14
feat: add project-move tool for moving projects between personal and workspace

scottlovegrove:
## Summary

- Adds a new `project-move` tool that uses the SDK's `moveProjectToWorkspace` and `moveProjectToPersonal` methods
- Bumps `@doist/todoist-api-typescript` from 6.4.0 to 6.5.0 to access the new move APIs
- Supports optional `folderId` and `visibility` parameters when moving to a workspace
- Registers the tool in the MCP server and adds it to the run-tool CLI script

Inspired by [todoist-cli PR #72](https://github.com/Doist/todoist-cli/pull/72).

## Test plan

- [x] 8 new tests covering all move scenarios (workspace with/without folder/visibility, personal, missing workspaceId error, error propagation)
- [x] All 432 tests pass
- [x] TypeScript type-check passes
- [x] Biome lint/format passes
- [x] Schema validation passes (24 tools, 70 parameters, Gemini compatible)
- [x] `project-move` appears in `run-tool.ts --list`

### DIFF
--- a/src/tools/project-move.ts
+++ b/src/tools/project-move.ts
@@ -1,0 +1,87 @@
+import type {
+    MoveProjectToWorkspaceArgs,
+    PersonalProject,
+    WorkspaceProject,
+} from '@doist/todoist-api-typescript'
+import { z } from 'zod'
+import type { TodoistTool } from '../todoist-tool.js'
+import { mapProject } from '../tool-helpers.js'
+import { ProjectSchema } from '../utils/output-schemas.js'
+import { ToolNames } from '../utils/tool-names.js'
+
+const ArgsSchema = {
+    action: z
+        .enum(['move-to-workspace', 'move-to-personal'])
+        .describe('The action to perform on the project.'),
+    projectId: z.string().min(1).describe('The ID of the project to move.'),
+    workspaceId: z
+        .string()
+        .min(1)
+        .optional()
+        .describe('The target workspace ID. Required when action is move-to-workspace.'),
+    folderId: z
+        .string()
+        .min(1)
+        .optional()
+        .describe('Optional target folder ID within the workspace.'),
+    visibility: z
+        .enum(['restricted', 'team', 'public'])
+        .optional()
+        .describe(
+            'Optional access visibility for the project in the workspace (restricted, team, or public).',
+        ),
+}
+
+const OutputSchema = {
+    project: ProjectSchema.describe('The moved project.'),
+    success: z.boolean().describe('Whether the move was successful.'),
+}
+
+const projectMove = {
+    name: ToolNames.PROJECT_MOVE,
+    description: 'Move a project between personal and workspace contexts.',
+    parameters: ArgsSchema,
+    outputSchema: OutputSchema,
+    annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true },
+    async execute(args, client) {
+        let project: PersonalProject | WorkspaceProject
+
+        if (args.action === 'move-to-workspace') {
+            if (!args.workspaceId) {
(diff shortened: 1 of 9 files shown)


frankieyan:
Hey @scottlovegrove, I ran out of time to look at this one + https://github.com/Doist/todoist-cli/pull/72. I'll pick them up tomorrow AM.

src/tools/project-move.ts
```diff
@@ -0,0 +1,87 @@
...
+    PersonalProject,
+    WorkspaceProject,
+} from '@doist/todoist-api-typescript'
+import { z } from 'zod'
+import type { TodoistTool } from '../todoist-tool.js'
+import { mapProject } from '../tool-helpers.js'
+import { ProjectSchema } from '../utils/output-schemas.js'
+import { ToolNames } from '../utils/tool-names.js'
+
+const ArgsSchema = {
```
frankieyan, line 12: Would we be able to use zod's [discriminated unions](https://zod.dev/api?id=discriminated-unions) to create a stronger validator and type here? I.e. `workspaceId` should be required when `move-to-workspace` is used, while `folderId` and `visiblity` should not be a part of args when moving to personal.

src/tools/project-move.ts
```diff
@@ -0,0 +1,87 @@
...
+const projectMove = {
+    name: ToolNames.PROJECT_MOVE,
+    description: 'Move a project between personal and workspace contexts.',
+    parameters: ArgsSchema,
+    outputSchema: OutputSchema,
+    annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true },
+    async execute(args, client) {
+        let project: PersonalProject | WorkspaceProject
+
+        if (args.action === 'move-to-workspace') {
```
frankieyan, line 49: If we are able to make the schema stricter, could we use zod's `parse` or `safeParse` here?

src/tools/project-move.ts
scottlovegrove: Good question! I looked into this — discriminated unions are technically possible at the Zod level, but there are a couple of reasons this wouldn't work well here:

1. **LLM compatibility**: The tool schema gets converted to JSON Schema for LLM consumption. A discriminated union produces a `oneOf` schema, which many LLM providers (Gemini in particular) handle poorly compared to a flat object with optional fields and good descriptions.

2. **Framework constraints**: Our `TodoistTool` type constrains `Params extends z.ZodRawShape` (a flat record of Zod fields). A `z.discriminatedUnion()` is a `ZodDiscriminatedUnion` — a complete schema, not a raw shape. Adopting it would require changes to `TodoistTool`, `registerTool` in `mcp-helpers.ts`, and the `execute` type inference — all shared infrastructure for a single tool.

3. **Established pattern**: Every other tool with conditionally-required params uses this same flat pattern (e.g., `manage-assignments.ts` has `operation: z.enum([...])` with `responsibleUser` as optional + a runtime check). It's a deliberate design choice for LLM compatibility, not an oversight.

The current approach (optional fields with descriptive `.describe()` + runtime validation) is the most reliable pattern for MCP tools consumed by diverse LLM providers.

src/tools/project-move.ts
scottlovegrove: The MCP SDK framework already validates inputs against the tool's schema before the `execute` callback is invoked, so adding explicit `z.parse()`/`z.safeParse()` would be redundant. No tools in the codebase use explicit Zod parsing for this reason — the runtime check here (`if (!args.workspaceId)`) is just business logic validation for the conditional requirement, which is the same pattern used across all tools (e.g., `manage-assignments.ts` line 93).

PR #328, opened 2026-02-27
feat: typed color support for add-projects and update-projects

scottlovegrove:
## Summary

- **New `ColorSchema` utility** (`src/utils/colors.ts`) — normalises color keys (`berry_red`) or display names (`Berry Red`) to the canonical key; unrecognised values silently resolve to `undefined` so the Todoist default (charcoal) is used
- **`ColorKeySchema`** — plain enum of the 20 valid SDK color keys, used in the output `ProjectSchema` in place of `z.string()`
- **`add-projects` and `update-projects`** — both now accept an optional `color` field in their input schemas
- **Improved skip messaging in `update-projects`** — distinguishes `"no changes"` (no fields provided at all) from `"no valid field values"` (fields provided but all normalised away, e.g. an unrecognised color), so the LLM gets actionable feedback
- **Test fixes** — corrected `'grey'` → `'gray'` in `get-overview` and `find-projects` tests (not a valid SDK color key)

## Test plan

- [x] All 447 tests pass
- [x] Build and type-check pass
- [x] `ColorSchema.parse('Berry Red')` → `'berry_red'`
- [x] `ColorSchema.parse('hotpink')` → `undefined`
- [x] Update with only an invalid color reports `"skipped - no valid field values"`
- [x] Update with invalid color + valid field still applies the valid field

### DIFF
--- a/src/tool-helpers.ts
+++ b/src/tool-helpers.ts
@@ -1,5 +1,6 @@
 import type {
     ActivityEvent,
+    ColorKey,
     Comment,
     CurrentUser,
     MoveTaskArgs,
@@ -236,7 +237,7 @@ function mapProject(project: Project) {
     return {
         id: project.id,
         name: project.name,
-        color: project.color,
+        color: project.color as ColorKey,
         isFavorite: project.isFavorite,
         isShared: project.isShared,
         parentId: isPersonalProject(project) ? (project.parentId ?? undefined) : undefined,
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "7.10.1",
             "license": "MIT",
             "dependencies": {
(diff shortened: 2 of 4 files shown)


src/tool-helpers.ts
```diff
@@ -236,7 +237,7 @@ function mapProject(project: Project) {
     return {
         id: project.id,
         name: project.name,
-        color: project.color,
+        color: project.color as ColorKey,
```
frankieyan, line 240: Would it be reasonable to narrow this to `ColorKey` in the TS SDK? While we're using it for request params, they aren't added to the various entity types. We might be technically lying since bad data can exist (as seen in #344), but we're also lying here when we're using it for fetched projects. 

Alternatively, we could do something like `ColorKey | string` in the SDK to at least hint that those colours are are a part of the dataset, despite having no guarantees.

PR #341, opened 2026-02-28
chore: Stops renovate updating to newer major node versions

scottlovegrove:
As per the description. basically, if you look at #288, it's trying to update to v24 of `@types/node`, even though we are at 22 in this project. 

CleanShot 2026-02-28 at 22 17 56@2x


frankieyan:
✅

PR #350, opened 2026-03-03
fix: default find-completed-tasks date window

### LINKED ISSUE #335: `find-completed-tasks` has undocumented required parameters
From Anthropic:

> The since and until parameters are listed as optional in the schema, but calling the tool without them returns a validation error ("Invalid input: expected string, received undefined"). These should either be marked as required in the schema, or the tool should provide sensible defaults (e.g., last 7 days).

### LINKED ISSUE #340: `find-completed-tasks` could benefit from sensible defaults
From Anthropic: 

> Rather than erroring when since/until are omitted, consider defaulting to the last 7 days. This would make the tool more user-friendly for common queries like "what did I complete recently?"

gnapse:
# Pull Request

Closes #335
Closes #340

## Short description

Made `since`/`until` optional for `find-completed-tasks` and apply a deterministic 7-day default window based on the user GMT offset when omitted.
The tool now reports the effective date range in both text output and `structuredContent.appliedFilters`, while preserving existing explicit-date behavior and local-date-to-UTC boundaries.
Added tests for optional schema parsing, default date-window application, and timezone-safe default behavior.
Extracted reusable date primitives into `src/utils/date.ts` and updated MCP tool guidance text to reflect the optional date behavior.

## PR Checklist

### 1) Missing `since`/`until` no longer fails

**Before**
- Calling `find-completed-tasks` without `since`/`until` could fail validation.

**After**
```json
// input
{ "getBy": "completion", "limit": 50 }

// effective filters in response
{
  "appliedFilters": {
    "getBy": "completion",
    "since": "2026-02-25",
    "until": "2026-03-03",
    "limit": 50
  }
}
```

### 2) Timezone-safe default window

For a user with `gmtString: "+02:00"` and local “today” = 2025-10-12:

- effective local range is 2025-10-06 → 2025-10-12
- API UTC boundaries sent are:
    - since: `2025-10-05T22:00:00.000Z`
    - until: `2025-10-12T21:59:59.000Z`

### 3) Explicit dates unchanged

If caller provides:

```json
{ "since": "2025-08-01", "until": "2025-08-31" }
```

…the tool keeps those dates (same behavior as before), only converting local-day boundaries to UTC for the API request.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -111,7 +111,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 - **Team Assignment**: find-project-collaborators → add-tasks with responsibleUser → manage-assignments for bulk changes
 - **Task Search**: find-tasks with multiple filters → update-tasks or complete-tasks based on results
 - **Project Organization**: add-projects → add-sections → add-tasks with projectId and sectionId
-- **Progress Reviews**: find-completed-tasks (optionally with explicit date ranges) → get-overview for project summaries
+- **Progress Reviews**: find-completed-tasks (defaults to last 7 days; optionally use explicit date ranges) → get-overview for project summaries
 - **Activity Auditing**: find-activity with event/object filters to track changes, monitor team activity, or investigate specific actions
 
 Always provide clear, actionable task titles and descriptions. Use the overview tools to give users context about their workload and project status.


src/utils/date.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Shift a YYYY-MM-DD date string by the specified number of days in UTC.
+ */
+export function shiftDateStringByDays(dateString: string, days: number): string {
+    const [yearPart, monthPart, dayPart] = dateString.split('-')
+    const year = Number(yearPart)
+    const month = Number(monthPart)
+    const day = Number(dayPart)
+
+    if (!Number.isFinite(year) || !Number.isFinite(month) || !Number.isFinite(day)) {
```
frankieyan, line 43: No problem keeping this check in place, but just curious, all input should be validated by zod at this point, correct?

src/mcp-server.ts
```diff
@@ -111,7 +111,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 - **Team Assignment**: find-project-collaborators → add-tasks with responsibleUser → manage-assignments for bulk changes
 - **Task Search**: find-tasks with multiple filters → update-tasks or complete-tasks based on results
 - **Project Organization**: add-projects → add-sections → add-tasks with projectId and sectionId
-- **Progress Reviews**: find-completed-tasks with date ranges → get-overview for project summaries
+- **Progress Reviews**: find-completed-tasks (optionally with explicit date ranges) → get-overview for project summaries
```
frankieyan: Would it be worth mentioning the 7-day default here again, or should agents already know this from above/be able to grep this?

src/utils/date.ts
gnapse: Maybe, but from the function's POV I still think it is valuable that it guards its own input, just in case it ends up used by something not properly vetted.

src/mcp-server.ts
gnapse: Will improve it before merging.

PR #351, opened 2026-03-03
fix: make due/deadline clearing explicit in update-tasks

### LINKED ISSUE #336: No documented way to remove a due date from a task
From Anthropic: 

> We attempted "no date" as a dueDate value in update-tasks — it was accepted without error but the original due date remained. There's no clear guidance on how to clear a due date (null, empty string, specific keyword, etc.). This should be documented or a clear mechanism should be provided.

gnapse:
# Pull Request

Closes #336

## Short description

Make due-date removal deterministic in `update-tasks` by treating `dueString: "remove"` as clear while keeping compatibility aliases (`"no date"` and `null`) and mirroring alias-based clearing for `deadlineDate`.
Update MCP guidance for `update-tasks` to document the canonical removal syntax (`remove`) and add tests covering due/deadline clearing behavior, including output state after due-date removal.

## PR Checklist


src/tools/update-tasks.ts
```diff
@@ -129,15 +143,28 @@ const updateTasks = {
...
-                    // Set new deadline
-                    updateArgs = { ...updateArgs, deadlineDate }
+            // Handle due date changes if provided
+            const dueStringUpdate = normalizeNullEquivalentValue(
+                dueString,
+                DUE_DATE_REMOVAL_ALIASES,
+            )
+            if (dueStringUpdate === null) {
+                // SDK typings may lag API support for null dueString. Cast keeps runtime null payload.
+                updateArgs = {
```
pedroalves0: I would’ve thought the SDK does input validation and that this wouldn’t pass, but it doesn’t seem to be the case ([ref](https://github.com/Doist/todoist-api-typescript/blob/d3ac04735043976d446f13374c7b1d192db960d1/src/todoist-api.ts#L595-L614)). The API docs also aren’t clear about this: there’s no indication that a `null` dueString is supported ([ref](https://developer.todoist.com/api/v1/#tag/Tasks/operation/update_task_api_v1_tasks__task_id__post)), although it definitely should be.

Have you confirmed this works reliably? If so, I think the best path is to first update the API docs, and then update the SDK to properly support `null` as a dueString value.

pedroalves0 (approved):

src/tools/update-tasks.ts
gnapse: Good catch. I double-checked today and it does not clear the date. I guess my confusion came from the fact that `null` on the deadline field does clear the date.

So here's what I did: I kept the outer interface of this tool accepting `null`, and internally I'm passing the string `"no date"` which is the value that seems to clear the due date in practice.

But I'll soon check in the typescript API library and the API itself how does this all work, and why the inconsistency between deadline and due date when passing `null`.

PR #357, opened 2026-03-05
feat: add filter management tools (find, add, update, delete)

RonenMars:
## Summary

Adds three new MCP tools for management of personal Todoist filters, and extends `delete-object` to support filter deletion. Filters are saved custom task views using Todoist's query syntax (e.g. `today & p1`, `#Work & overdue`).

- **`find-filters`** — list all personal filters or search by name (read-only, idempotent)
- **`add-filters`** — create one or more filters with name, query, color, and isFavorite flag
- **`update-filters`** — modify existing filters; entries with no changed fields are skipped gracefully
- **`delete-object`** — extended with `'filter'` type to delete filters by ID, consistent with how other entity types are deleted

All tools follow the established patterns: per-tool MCP `ToolAnnotations`, flat Zod schemas for LLM compatibility, named-args registration, and Sync API usage (filters are not available via the REST API).

### Sync API usage
Filters are managed entirely through `client.sync()` + `createCommand()`:
- `filter_add`, `filter_update`, `filter_delete` commands
- `resourceTypes: ['filters']` for reads

### tempId / tempIdMapping quirk
The Todoist Sync API returns `tempIdMapping` with **camelCase keys** (e.g. `tempFilterAdd0`), so tempIds must be generated in camelCase to match. Combining commands + `resourceTypes` + `syncToken:'*'` in one call suppresses `tempIdMapping` — write and read calls are therefore kept separate.

## Example Usage

```bash
# List all filters
npx tsx scripts/run-tool.ts find-filters '{}'

# Search filters by name (partial, case-insensitive)
npx tsx scripts/run-tool.ts find-filters '{"search":"work"}'

# Create filters (supports batch)
npx tsx scripts/run-tool.ts add-filters '{
  "filters": [
    { "name": "Today P1", "query": "today & p1", "color": "red", "isFavorite": true },
    { "name": "Work overdue", "query": "#Work & overdue", "color": "orange" }
  ]
}'

# Update a filter — entries with no changed fields are skipped gracefully
npx tsx scripts/run-tool.ts update-filters '{
  "filters": [
    { "id": "2345678", "query": "today & (p1 | p2)", "isFavorite": true }
  ]
}'

# Delete a filter by ID via delete-object
npx tsx scripts/run-tool.ts delete-object '{"type": "filter", "id": "2345678"}'
```

**Typical LLM workflow:**
1. `find-filters` → get filter IDs and current queries
2. `update-filters` → modify specific filters by ID
3. `delete-object` with `type: "filter"` → remove a filter no longer needed

## Test Plan

- [x] New tests across filter test files (624 total, all passing)
- [x] `tool-annotations.test.ts` updated to cover all new filter tools
- [x] `delete-object.test.ts` updated with filter deletion tests
- [x] TypeScript type-check passes (`npx tsc --noEmit`)
- [x] Biome lint + format passes (`biome check`)
- [x] Schema validation passes (Gemini-compatible flat schemas, no `oneOf`)
- [x] All tools appear in `npm run tool -- --list`

## PR Checklist

- [x] Added tests for new features
- [x] Updated docs (`src/mcp-server.ts` tool usage guidelines)
- [x] New tools added to `getMcpServer` AND exported in `src/index.ts`
- [x] Tools added to `tool-annotations.test.ts`
- [x] All 624 tests pass
- [x] Build and type-check pass (`npx tsc --noEmit`)
- [x] Biome lint + format pass (`biome check`)
- [x] Schema validation: Gemini-compatible flat Zod schemas (no `oneOf`, no `nullable`)

## Design Rationale

Follows the repository's workflow-centric philosophy:

- **User intent**: Mirrors how users think about filters — "show me my Work overdue tasks" maps directly to a named filter with `#Work & overdue`
- **Sync API as the right primitive**: The REST API has no filter endpoints, so `client.sync()` with `filter_add/update/delete` commands is the only correct path — matching how the Todoist app itself manages filters
- **Filter deletion via `delete-object`**: Rather than a standalone `delete-filter` tool, filter deletion is handled by the existing `delete-object` tool with a new `'filter'` type, keeping the tool surface consistent with how projects, sections, tasks, comments, and labels are deleted
- **Separate write + read calls**: Combining `commands` with `resourceTypes` + `syncToken:'*'` in one sync call suppresses `tempIdMapping`, making it impossible to correlate created IDs back to inputs. Keeping writes and reads as separate calls is the only reliable pattern for batch creation
- **Graceful no-op skipping in `update-filters`**: Passing an update entry with no changed fields is silently skipped (not an error), matching how the existing `update-tasks` tool behaves and reducing noise in multi-item batch updates
- **Batch-first design**: `add-filters` and `update-filters` both accept arrays, enabling LLMs to create or modify multiple filters in a single tool call instead of looping

## Demo

https://github.com/user-attachments/assets/f18fd739-6834-405b-b40a-1139a74c51e8


scottlovegrove:
@RonenMars Would you mind rebasing this on latest main please. 

Also, re:

> Sync API color mismatch (bug fix)
> The Sync API's `filter_add`/`filter_update` endpoints reject the REST API color strings `turquoise` and `gray`, expecting `teal` and `grey` respectively. A new `src/utils/filter-colors.ts` module provides bidirectional remaps applied transparently on write and read so users and LLMs see consistent `ColorKey` values throughout.

This has been fixed in the SDK, so the workaround shouldn't be required now.

scottlovegrove:
> * **`delete-filter`** — remove a filter by ID (idempotent)

Also, just a note about this, we shouldn't have a new tool for this, we should make use of the `delete-object` tool and expand that to support filters.

PR #359, opened 2026-03-06
feat: add file attachment support to add-comments

arndjan:
## Summary
- Adds optional `filePath` parameter to the `add-comments` tool
- When provided, the file is uploaded via the existing `uploadFile()` API and attached to the comment
- The `@doist/todoist-api-typescript` client already supports both `uploadFile()` and the `attachment` field on `AddCommentArgs` — this simply exposes that capability through the MCP tool

## Use case
AI agents managing tasks (e.g. filing invoices, attaching documents) can now attach files directly to Todoist comments, rather than only providing text references to files stored elsewhere.

## Changes
- `src/tools/add-comments.ts` — added `filePath` field to `CommentSchema`, upload logic in `execute`
- `src/tools/__tests__/add-comments.test.ts` — 2 new tests for attachment upload + no-attachment case

## Test plan
- [x] All 468 existing tests pass
- [x] 2 new tests for file attachment flow
- [x] Type checking passes
- [x] Build succeeds

### DIFF
--- a/src/tools/__tests__/add-comments.test.ts
+++ b/src/tools/__tests__/add-comments.test.ts
@@ -8,6 +8,7 @@ import { addComments } from '../add-comments.js'
 const mockTodoistApi = {
     addComment: vi.fn(),
     getUser: vi.fn(),
+    uploadFile: vi.fn(),
 } as unknown as Mocked<TodoistApi>
 
 const { ADD_COMMENTS } = ToolNames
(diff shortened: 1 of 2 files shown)


scottlovegrove:
@arndjan Thanks for creating this PR. I'm going to close it without merging though as we already have #191 open for this and is still in discussion.

PR #376, opened 2026-03-15
fix: use wildcard queries for substring matching in search

### LINKED ISSUE #358: `searchProjects` does not return projects that have a `parentId`
## Description

The `searchProjects` endpoint used by the `find-projects` MCP tool does not return any projects that have a `parentId` set (i.e., sub-projects nested under a parent). The non-search path (`getProjects` with pagination) correctly returns sub-projects.

## Steps to Reproduce

1. Create a project nested under a parent project (e.g., "AI Factory" under top-level project "HBS")
2. Call `find-projects` with `search: "AI Factory"`
3. Observe: zero results returned
4. Call `get-overview` (which uses `getProjects` internally) and confirm the project exists with a `parentId`
5. Call 'find-projects` with  `search: "HBS"` and project is correctly returned (in this case, "HBS" has no `parentId`)

## Expected Behavior

`searchProjects` should return matching projects regardless of whether they have a `parentId`.

## Actual Behavior

Projects with a `parentId` are silently excluded from search results. Only top-level projects (no parent) are returned.

## Investigation

The search path calls `searchProjects({ query })` which hits a dedicated Todoist API endpoint. This is not client-side filtering -- the issue appears to be in the Todoist API itself or in how the MCP server calls it.

The non-search code path (listing all projects via `getProjects` with pagination) does return sub-projects correctly.

## Workaround

- Use `fetch-object` with the direct project ID if the project ID is known
- Use `get-overview` (no `projectId` param) to list all projects including sub-projects

## Environment

- MCP Server: `@doist/todoist-ai`
- Client: Claude Code

scottlovegrove:
## Summary

- Wraps search queries with `*` wildcards for project, label, and section searches, enabling substring matching instead of the API's default literal/prefix matching
- Adds `toWildcardQuery` helper that escapes literal backslashes and asterisks before wrapping
- Fixes project search not returning sub-projects or partial name matches

Closes #358

## Test plan

- [x] All 550 tests pass
- [x] Manual verification: `find-projects '{"search":"Test"}'` returns both "Test" and "Sub-test"
- [x] Manual verification: `find-projects '{"search":"sub"}'` returns "Sub-test"
- [x] Manual verification: `find-projects '{"search":"Photography"}'` returns "Photography 📸"

### DIFF
--- a/src/tools/__tests__/__snapshots__/find-labels.test.ts.snap
+++ b/src/tools/__tests__/__snapshots__/find-labels.test.ts.snap
@@ -44,7 +44,7 @@ No shared labels."
 exports[`find-labels tool > searching labels > should return empty list when no labels match exact search 1`] = `
 "All labels matching "nonexistent": 0.
 Filter: search: "nonexistent".
-No results. Search requires the exact label name; Check spelling; Remove search to see all labels.
+No results. Try broader search terms; Check spelling; Remove search to see all labels.
 No shared labels."
 `;
 
--- a/src/tools/find-labels.ts
+++ b/src/tools/find-labels.ts
@@ -11,7 +11,9 @@ const ArgsSchema = {
     search: z
         .string()
         .optional()
-        .describe('Search for a label by exact name. If omitted, all labels are returned.'),
+        .describe(
+            'Search for a label by name (partial and case insensitive match). If omitted, all labels are returned.',
+        ),
     limit: z
         .number()
         .int()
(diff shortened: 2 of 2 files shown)


src/tool-helpers.ts
```diff
@@ -128,6 +128,15 @@ export async function fetchAllPages<
     return allResults
 }
 
+/**
+ * Wraps a search query with wildcards for substring matching.
+ * Escapes any literal backslashes and asterisks in the query first.
+ */
+export function toWildcardQuery(query: string): string {
```
henningmu, line 135: Probably fine for a first non-over-engineered version. However, if a user used a wildcard deliberately (e.g. to match prefixes) we would break it here. Maybe we could scan if a wildcard is present, if not, append it, else do nothing? 🤔

PR #410, opened 2026-03-25
fix: move .optional() outside z.preprocess() for dueString/deadlineDate

### LINKED ISSUE #409: update-tasks schema incorrectly marks dueString and deadlineDate as required due to z.preprocess() wrapper
## Bug

Since commit f05bd11 (`fix: make due/deadline clearing explicit in update-tasks`, v7.13.0), the `update-tasks` tool's emitted JSON Schema incorrectly marks `dueString` and `deadlineDate` as `required` fields. This causes LLM agents to always provide these values — even when they only intend to update labels or priority — which destroys recurring task schedules.

## Root Cause

The commit wrapped both fields in `z.preprocess()`:

```ts
dueString: z.preprocess(
    (value) => (value === null ? 'remove' : value),
    z.string().optional().describe("..."),
),
```

The `.optional()` lives **inside** the `z.preprocess()` wrapper. When the MCP SDK converts the Zod schema to JSON Schema (via `zod-to-json-schema`), `z.preprocess()` creates a `ZodEffects` type that does not propagate the inner `.optional()` flag. The resulting JSON Schema puts both fields in the `required` array:

```json
"required": ["id", "dueString", "deadlineDate"]
```

At runtime validation, Zod correctly treats the fields as optional. But the **emitted JSON Schema** — which is what LLM agents see — says they're required.

## Impact

When an agent calls `update-tasks` to add a label to an existing recurring task (e.g., tagging it with a `decision` label), it is forced by the schema to supply a `dueString`. The Todoist API then replaces the entire due string — including any recurrence pattern like "every weekday at 9am" — with whatever flat date string the agent provided. The task silently loses its recurring status.

## Suggested Fix

Move `.optional()` to wrap the `z.preprocess()` call rather than living inside it:

```ts
dueString: z.preprocess(
    (value) => (value === null ? 'remove' : value),
    z.string().describe("..."),
).optional(),
```

This preserves the legacy-null-to-"remove" coercion while correctly emitting the field as optional in JSON Schema.

## Versions Affected

v7.13.0 through current (v8.4.0+). The remote server at `ai.todoist.net/mcp` serves the incorrect schema to all MCP clients.

scottlovegrove:
## Summary

- Fixes #409: `z.preprocess()` wrapping `.optional()` inside it causes Zod v4's `toJSONSchema()` to emit `dueString` and `deadlineDate` as `required` fields in the JSON Schema
- LLM agents are forced to always supply these values, which destroys recurring task schedules when they only intend to update other fields (e.g. labels, priority)
- Moves `.optional()` from inside `z.preprocess()` to outside it, so the emitted schema correctly marks the fields as optional while preserving the legacy null-to-"remove" coercion

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] All 692 tests pass
- [x] Verified emitted JSON Schema no longer includes `dueString`/`deadlineDate` in `required` array
- [x] Verified runtime parsing still works: omitting fields, providing values, legacy `null`, and `"remove"` all behave identically

### DIFF
--- a/src/tools/update-tasks.ts
+++ b/src/tools/update-tasks.ts
@@ -37,26 +37,28 @@ const TasksUpdateSchema = z.object({
     parentId: z.string().optional().describe('The new parent task ID (for subtasks).'),
     order: z.number().optional().describe('The new order of the task within its parent/section.'),
     priority: PrioritySchema.optional().describe(PRIORITY_INPUT_DESCRIPTION),
-    dueString: z.preprocess(
-        // Keep accepting legacy null while exposing a Gemini-compatible string schema.
-        (value) => (value === null ? 'remove' : value),
-        z
-            .string()
-            .optional()
-            .describe(
-                'The new due date for the task in natural language (e.g., "tomorrow at 5pm"). Use "remove" to clear the due date.',
-            ),
-    ),
-    deadlineDate: z.preprocess(
-        // Keep accepting legacy null while exposing a Gemini-compatible string schema.
-        (value) => (value === null ? 'remove' : value),
-        z
-            .string()
-            .optional()
-            .describe(
-                'The new deadline date for the task in ISO 8601 format (YYYY-MM-DD, e.g., "2025-12-31"). Deadlines are immovable constraints shown with a different indicator than due dates. Use "remove" to clear the deadline.',
-            ),
-    ),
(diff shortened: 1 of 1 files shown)


pawelgrimm:
🚀

PR #414, opened 2026-03-26
feat: introduce MCP Apps

henningmu:
## 🗺️ Overview

- removes the existing ChatGPT app/widget pipeline and old widget assets
- introduces the MCP Apps task-list app, server/resource wiring, and build pipeline
- slightly improves the task-list UI

## 🧪 Testing

- [x] `npm run build`, `npm run start` -> Use the app in the MCP Inspector
- [ ] `npm run start:http`, expose with ngrok -> Use the app in Claude

## 📸 Demoes

MCP Inspector: https://cleanshot.com/share/0ykvNm0y
Claude: https://cleanshot.com/share/rf40MCRm

### DIFF
--- a/src/mcp-apps/task-list/app.tsx
+++ b/src/mcp-apps/task-list/app.tsx
@@ -1,14 +1,19 @@
 import { useApp, useHostStyles } from '@modelcontextprotocol/ext-apps/react'
 import { useCallback, useMemo, useState } from 'react'
+import type { z } from 'zod'
+import { ArgsSchema } from '../../tools/find-tasks-by-date.js'
+import { ToolNames } from '../../utils/tool-names.js'
 import { Empty } from './empty'
 import { Loading } from './loading'
 import { TaskList } from './task-list'
 import styles from './task-list.module.css'
 import type { Task } from './types'
 
+type FindTasksByDateArgs = z.input<z.ZodObject<typeof ArgsSchema>>
+
 type ToolOutput = {
     tasks: Task[]
-    appliedFilters?: Record<string, unknown>
+    appliedFilters?: FindTasksByDateArgs
 }
 
 type ToolResult = {
@@ -19,7 +24,7 @@ type ToolResult = {
 
 const TODOIST_APP_URL = 'https://app.todoist.com/app'
 
-function getInvocationArgs(args: Record<string, unknown> | null | undefined) {
+function getInvocationArgs(args: FindTasksByDateArgs | null | undefined) {
     if (!args) {
         return null
     }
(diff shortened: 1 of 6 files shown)


src/mcp-apps/task-list/app.tsx
```diff
@@ -0,0 +1,180 @@
...
+            if (!app || !tasks) return
+
+            const previousTasks = tasks
+
+            setTasks((previous) => previous?.filter((task) => task.id !== taskId) ?? [])
+            setToolError(null)
+
+            try {
+                const result = await app.callServerTool({
+                    name: 'complete-tasks',
```
scottlovegrove: Maybe we should call the constant from tool-names, since the `callServerTool` isn't strongly typed.

src/mcp-apps/task-list/app.tsx
```diff
@@ -0,0 +1,180 @@
...
+            Open in Todoist
+            <OpenInTodoistIcon />
+        </button>
+    )
+}
+
+export function App() {
+    const [tasks, setTasks] = useState<Task[] | null>(null)
+    const [loading, setLoading] = useState(true)
+    const [lastArgs, setLastArgs] = useState<Record<string, unknown> | null>(null)
```
scottlovegrove: Is there a way we could type this a bit better than just `Record<string, unknown>`? We already know what all the potential args will be, right?

src/mcp-apps/task-list/task-list.tsx
```diff
@@ -0,0 +1,36 @@
+import { TaskListItem } from './task-item'
+import styles from './task-list.module.css'
+import type { PriorityLevel, Task } from './types'
+
+type Props = {
+    tasks: Task[]
+    onCompleteTask: (taskId: string) => void
+}
+
+const PRIORITY_ORDER: Record<PriorityLevel, number> = { p1: 4, p2: 3, p3: 2, p4: 1 }
```
scottlovegrove: I wonder if we could borrow from the Todoist SDK as we have these kinds of mappings in there. We even have a helper in this project in the `priorites.ts` (added before I put it in the SDK).

src/mcp-apps/task-list/types.ts
```diff
@@ -0,0 +1,7 @@
+export type PriorityLevel = 'p1' | 'p2' | 'p3' | 'p4'
```
scottlovegrove: Why do we need this type? Can't we just use `Priority` from `priorities.ts`?

src/mcp-helpers.ts
```diff
@@ -48,9 +45,9 @@ type Feature = {
...
 
-type McpTextResource = {
-    name: string
-} & TextResourceContents
+type AppToolMeta =
+    | ({ ui: Record<string, unknown> } & Record<string, unknown>)
+    | ({ 'ui/resourceUri': string } & Record<string, unknown>)
 
 /**
  * Wether to return the structured content directly, vs. in the `content` part of the output.
```
scottlovegrove:
```suggestion
 * Whether to return the structured content directly, vs. in the `content` part of the output.
```
Not your PR change, but if you fancy it, saves a separate PR to fix 😄

PR #421, opened 2026-03-29
fix: remove biome format step from release pipeline

scottlovegrove:
## Summary
- Removes the `@semantic-release/exec` step that ran `npx @biomejs/biome format --write CHANGELOG.md` during releases
- Biome doesn't support formatting Markdown files, causing the release to fail with exit code 1
- Removes the now-unused `@semantic-release/exec` devDependency

## Test plan
- [ ] Verify the release workflow passes on merge

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,6 @@
             "devDependencies": {
                 "@biomejs/biome": "2.4.7",
                 "@semantic-release/changelog": "6.0.3",
-                "@semantic-release/exec": "7.1.0",
                 "@semantic-release/git": "10.0.1",
                 "@types/dompurify": "3.2.0",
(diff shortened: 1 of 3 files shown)


rfgamaral:
FYI, this was added to Typist because `CHANGELOG.md` was changed by `semantic-release` in a format not compatible with prettier, which would fail PR pipelines after a release has been made. I'm guessing this repo does not validate the format of `.md` files or the `CHANGELOG.md` file? But if we ever want to do that, we'll run into this problem, and we'll need to put the exec step back. Other than that ✅

scottlovegrove:
@rfgamaral I hit that in another repo ages ago and just added the changelog.md to the prettier/biome ignore file 🤷🏻‍♂️

Issue #426, opened 2026-03-31
`add-sections` always fails; `delete-object` silently fails for workspace projects; `list-workspaces` crashes with Zod errors


keyboardAnt:
## Bug Summary

Three tools in the official Todoist MCP (`https://ai.todoist.net/mcp`) are broken when used via Claude.ai’s built-in Todoist connector. Two additional bugs were discovered during cleanup attempts.

## Bug 1: `add-sections` always fails (total blocker)

Every call returns a generic error with no diagnostic detail:

```json
{"error": "Error occurred during tool execution"}
```

**Tested inputs (all fail identically):**

- Personal project: `{"sections": [{"name": "Test Section", "projectId": "<PERSONAL_PROJECT_ID>"}]}`
- Workspace project: `{"sections": [{"name": "Test Section", "projectId": "<WORKSPACE_PROJECT_ID>"}]}`
- Pre-existing project: `{"sections": [{"name": "Test Section", "projectId": "<EXISTING_PROJECT_ID>"}]}`

No input validation errors — the schema accepts the call, it fails server-side.

## Bug 2: `list-workspaces` fails with Zod validation errors

Calling `list-workspaces` (no parameters) returns:

```json
[
  {"expected": "string", "code": "invalid_type", "path": ["id"], "message": "Invalid input: expected string, received number"},
  {"code": "invalid_value", "values": ["ADMIN","MEMBER","GUEST"], "path": ["role"], "message": "Invalid option: expected one of \"ADMIN\"|\"MEMBER\"|\"GUEST\""},
  {"expected": "object", "code": "invalid_type", "path": ["limits"], "message": "Invalid input: expected object, received undefined"},
  {"expected": "string", "code": "invalid_type", "path": ["createdAt"], "message": "Invalid input: expected string, received undefined"},
  {"expected": "string", "code": "invalid_type", "path": ["creatorId"], "message": "Invalid input: expected string, received number"}
]
```

Root cause appears to be: the Todoist API returns numeric `id` and `creatorId` but the Zod schema expects strings, and `limits`/`createdAt` fields are missing from the API response.

## Bug 3: `add-projects` with `workspace` parameter triggers same Zod errors

Creating a project directly into a workspace:

```json
{"projects": [{"name": "My Project", "color": "teal", "workspace": "<WORKSPACE_NAME>"}]}
```

…or using the workspace ID (`"workspace": "<WORKSPACE_ID>"`) — both trigger the identical Zod validation errors as Bug 2.

[...]

PR #427, opened 2026-03-31
fix: update SDK to v7.9.1 and handle optional workspace fields

scottlovegrove:
## Summary

- Updates `@doist/todoist-api-typescript` to v7.9.1, which fixes Zod validation errors in `getWorkspaces()` — the REST API returns numeric IDs and omits `role`, `limits`, and `createdAt` fields
- Updates the `list-workspaces` tool to handle the now-optional `role` and `createdAt` fields

This also fixes `add-projects` with a `workspace` parameter (Bug 3 in the issue), since workspace resolution calls `getWorkspaces()` internally.

Related: #426

## Test plan

- [x] `list-workspaces` returns all workspaces via `run-tool.ts` against live API
- [x] `add-projects` with workspace parameter works against live API
- [x] Typecheck passes
- [x] All 755 tests pass

### DIFF
--- a/src/tools/list-workspaces.ts
+++ b/src/tools/list-workspaces.ts
@@ -15,12 +15,18 @@ const WorkspaceOutputSchema = {
     id: z.string().describe('The unique identifier for the workspace.'),
     name: z.string().describe('The name of the workspace.'),
     plan: z.enum(WORKSPACE_PLANS).describe('The workspace plan type.'),
-    role: z.enum(WORKSPACE_ROLES).describe("The user's role in the workspace."),
+    role: z
+        .enum(WORKSPACE_ROLES)
+        .optional()
+        .describe("The user's role in the workspace, if available."),
     isLinkSharingEnabled: z
         .boolean()
         .describe('Whether link sharing is enabled for the workspace.'),
     isGuestAllowed: z.boolean().describe('Whether guests are allowed in the workspace.'),
-    createdAt: z.string().describe('The ISO 8601 timestamp when the workspace was created.'),
+    createdAt: z
+        .string()
+        .optional()
+        .describe('The ISO 8601 timestamp when the workspace was created.'),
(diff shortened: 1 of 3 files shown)


src/tools/list-workspaces.ts
```diff
@@ -15,12 +15,18 @@ const WorkspaceOutputSchema = {
...
+        .describe("The user's role in the workspace, if available."),
     isLinkSharingEnabled: z
         .boolean()
         .describe('Whether link sharing is enabled for the workspace.'),
     isGuestAllowed: z.boolean().describe('Whether guests are allowed in the workspace.'),
-    createdAt: z.string().describe('The ISO 8601 timestamp when the workspace was created.'),
+    createdAt: z
+        .string()
+        .optional()
+        .describe('The ISO 8601 timestamp when the workspace was created.'),
```
doistbot, line 29: [P3] Consistent field description

Consider appending ", if available." to this description for consistency with the `role` field description above.

```suggestion
        .describe('The ISO 8601 timestamp when the workspace was created, if available.'),
```

src/tools/list-workspaces.ts
```diff
@@ -78,12 +84,16 @@ async function generateWorkspacesList(
             lines.push(`## ${workspace.name}`)
             lines.push(`- **ID:** ${workspace.id}`)
             lines.push(`- **Plan:** ${workspace.plan}`)
-            lines.push(`- **Your Role:** ${workspace.role}`)
+            if (workspace.role) {
```
doistbot, line 87: [P2] Add regression coverage for the omitted-field shape this branch is handling. `list-workspaces.test.ts` still only exercises workspaces with `role` and `createdAt` populated, so the exact live API response that motivated this change is not protected against regressions. Since the same `getWorkspaces()` payload also feeds workspace resolution, a case with those keys omitted here—and ideally one `add-projects` workspace test—would lock in both fixes.

frankieyan:
✅

PR #428, opened 2026-03-31
feat: add isUrgent flag support for reminders

scottlovegrove:
## Summary
- Adds `isUrgent` optional boolean to add-reminders and update-reminders tools for relative and absolute reminder types
- Adds `isUrgent` to the output schema and `mapReminder` helper, so it also appears in find-reminders responses
- Extracts shared `ReminderIsUrgentSchema` in `reminder-schemas.ts` for consistent descriptions across tools

## Test plan
- [x] 4 new tests covering isUrgent on add (relative + absolute) and update (relative + absolute)
- [x] All 759 tests pass
- [x] TypeScript type check passes

### DIFF
--- a/src/tools/__tests__/update-reminders.test.ts
+++ b/src/tools/__tests__/update-reminders.test.ts
@@ -117,6 +117,29 @@ describe(`${UPDATE_REMINDERS} tool`, () => {
             )
         })
 
+        it('should clear isUrgent with false for relative reminder', async () => {
+            const updatedReminder = createMockRelativeReminder({
+                isUrgent: false,
+            } as Partial<Reminder>)
+            mockTodoistApi.updateReminder.mockResolvedValue(updatedReminder)
+
+            const result = await updateReminders.execute(
+                {
+                    reminders: [{ type: 'relative', id: 'reminder-1', isUrgent: false }],
+                },
+                mockTodoistApi,
+            )
+
+            expect(mockTodoistApi.updateReminder).toHaveBeenCalledWith('reminder-1', {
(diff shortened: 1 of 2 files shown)


src/tools/add-reminders.ts
```diff
@@ -24,6 +25,7 @@ const RelativeReminderInputSchema = z.object({
     service: ReminderServiceSchema.optional().describe(
         'Delivery method: "email" or "push" notification. Defaults to push.',
     ),
+    isUrgent: ReminderIsUrgentSchema,
```
doistbot, line 28: [P2] Update the MCP server instructions for this new flag. This adds a user-facing reminder option, but the reminders section in `src/mcp-server.ts` still only documents reminder types and IDs. Per the repo guidance, new tool features should be reflected there as well; otherwise `isUrgent` is much less discoverable to the model on add/update reminders.

src/tools/__tests__/update-reminders.test.ts
```diff
@@ -94,6 +94,29 @@ describe(`${UPDATE_REMINDERS} tool`, () => {
             )
         })
 
+        it('should update isUrgent for relative reminder', async () => {
```
doistbot, line 97: [P2] Add coverage for clearing urgency with `false`. `isUrgent: false` is the path users will use to remove urgency from an existing reminder, and boolean fields are easy to regress via truthiness/spread handling. The new tests only lock down the `true` case, so please add at least one update assertion that sends `false` and verifies it reaches `updateReminder` unchanged.

PR #429, opened 2026-03-31
fix: Add OpenAI specific meta data to MCP Apps

henningmu:
## Short description

To submit our MCP App to OpenAI, we need to add some proprietary metadata to the app. This metadata is simply ignored by the MCP inspector and Claude, making the change safe.

### Reference

CleanShot 2026-03-31 at 19 11 47@2x

### DIFF
--- a/src/mcp-apps/resources.ts
+++ b/src/mcp-apps/resources.ts
@@ -38,7 +38,7 @@ const taskListHash = createHash('sha256').update(taskListHtml).digest('hex').sli
 const taskListResourceUri = `ui://todoist/task-list@${taskListHash}`
 const taskListResourceDescription = 'Interactive task list widget'
 // Apps SDK expects an origin here, so use the hosted MCP origin rather than the full /mcp URL.
-const taskListWidgetDomain = new URL('https://ai.todoist.net/mcp').origin
+const taskListWidgetDomain = 'https://ai.todoist.net'
 const taskListResourceMeta = {
     ui: {
         prefersBorder: true,


src/mcp-apps/resources.ts
```diff
@@ -36,6 +36,27 @@ function loadTaskListHtml() {
 const taskListHtml = loadTaskListHtml()
 const taskListHash = createHash('sha256').update(taskListHtml).digest('hex').slice(0, 12)
 const taskListResourceUri = `ui://todoist/task-list@${taskListHash}`
+const taskListResourceDescription = 'Interactive task list widget'
+// Apps SDK expects an origin here, so use the hosted MCP origin rather than the full /mcp URL.
+const taskListWidgetDomain = new URL('https://ai.todoist.net/mcp').origin
```
doistbot: [P3] Unnecessary runtime URL parsing

Since the URL string is statically known, dynamically parsing it at runtime to extract the origin is unnecessary. You can safely hardcode the origin string directly to simplify the code.

```suggestion
const taskListWidgetDomain = 'https://ai.todoist.net'
```

src/mcp-apps/resources.ts
```diff
@@ -36,6 +36,27 @@ function loadTaskListHtml() {
...
+    ui: {
+        prefersBorder: true,
+        csp: {
+            // The widget is bundled into one HTML file and does not fetch or load external assets.
+            connectDomains: [] as string[],
+            resourceDomains: [] as string[],
+        },
+        domain: taskListWidgetDomain,
+    },
+    'openai/widgetDescription': taskListResourceDescription,
```
doistbot, line 52: [P2] Avoid maintaining two parallel copies of the widget config. The description, border preference, CSP, and domain now live once under `ui` and again under the `openai/widget*` keys, which makes it easy for the OpenAI and non-OpenAI paths to drift on the next change. It would be safer to keep one canonical widget config and derive the OpenAI aliases from it in one place.

PR #436, opened 2026-04-02
fix: add retry with exponential backoff for transient 5xx errors

craigcarlyle:
## Summary

- Adds automatic retry logic (up to 2 retries with exponential backoff) for transient HTTP 502/503/504 errors from the Todoist REST API
- Wraps all tool executions in `executeWithRetry()` so failures are transparently retried before surfacing to MCP clients
- Non-retryable errors (400, 401, 404, 500, etc.) continue to fail immediately with no behavior change

## Context

Datadog investigation of [Doist/Issues#19954](https://github.com/Doist/Issues/issues/19954) revealed:

- The Todoist REST API returns intermittent **503 Service Unavailable** errors during periodic load spikes, clustering heavily at the top of each hour (~262 logged 503s from MCP in 7 days)
- The `agentist` service also encounters **502 Bad Gateway** when connecting to `ai.todoist.net/mcp` (1–4 times daily since mid-March)
- Neither `todoist-ai` nor `@doist/todoist-sdk` had retry logic for HTTP 5xx responses — the SDK only retries on network-level errors (DNS failures, timeouts)

## Changes

| File | Change |
|------|--------|
| `src/utils/retry.ts` | New retry utility: `executeWithRetry()` with configurable exponential backoff (defaults: 2 retries, 500ms base delay, 2s max) |
| `src/mcp-helpers.ts` | Wraps `tool.execute()` with `executeWithRetry()` in `registerTool` |
| `src/utils/retry.test.ts` | 35 tests covering status code extraction, transient error detection, retry behavior, backoff timing, and configuration |

## Test plan

- [x] All 795 tests pass (including 35 new retry tests)
- [x] TypeScript type check passes
- [x] Biome lint/format passes
- [ ] Deploy to staging and verify 503 errors are retried transparently
- [ ] Monitor Datadog `todoist-ai-integrations` error logs for reduction in 503-related tool failures

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/utils/retry.ts
+++ b/src/utils/retry.ts
@@ -1,0 +1,90 @@
+const DEFAULT_MAX_RETRIES = 2
+const DEFAULT_BASE_DELAY_MS = 500
+const DEFAULT_MAX_DELAY_MS = 2000
+
+const RETRYABLE_STATUS_CODES = new Set([502, 503, 504])
+
+type RetryConfig = {
+    maxRetries?: number
+    baseDelayMs?: number
+    maxDelayMs?: number
+}
+
+function extractHttpStatusCode(error: unknown): number | undefined {
+    if (error === null || error === undefined || typeof error !== 'object') {
+        return undefined
+    }
+
+    const record = error as Record<string, unknown>
+
+    if (typeof record.httpStatusCode === 'number') {
+        return record.httpStatusCode
+    }
+
+    if (typeof record.statusCode === 'number') {
+        return record.statusCode
+    }
+
+    if (typeof record.status === 'number') {
+        return record.status
+    }
+
+    if (error instanceof Error) {
+        const match = error.message.match(/\bHTTP\s+(\d{3})\b/i)
+        if (match?.[1]) {
+            return Number(match[1])
+        }
+    }
+
+    return undefined
+}
+
+function isTransientError(error: unknown): boolean {
+    const statusCode = extractHttpStatusCode(error)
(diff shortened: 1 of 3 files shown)


src/utils/retry.ts
```diff
@@ -0,0 +1,90 @@
...
+    attempt,
+    baseDelayMs,
+    maxDelayMs,
+}: {
+    attempt: number
+    baseDelayMs: number
+    maxDelayMs: number
+}): number {
+    const delay = baseDelayMs * 2 ** attempt
+    return Math.min(delay, maxDelayMs)
```
doistbot, line 57: [P2] Lack of jitter in backoff strategy

The PR context mentions intermittent 503s clustering heavily at the top of each hour. Using pure exponential backoff without jitter means all failed clients will retry at the exact same intervals, causing synchronized spikes (a "thundering herd") that could exacerbate server load.

Consider adding randomized jitter (e.g., picking a random value between 0 and the computed delay) before returning the value. Note that doing so will require adjusting the strict delay assertions in `src/utils/retry.test.ts` to expect numeric ranges rather than exact match values.

scottlovegrove (commented):
Whilst I'm not wholly against this (retry is good), it is worth noting that at the SDK level, we already support [`fetchWithRetry`](https://github.com/Doist/todoist-sdk-typescript/blob/main/src/transport/fetch-with-retry.ts)

craigcarlyle:
The SDK's `fetchWithRetry` only retries on network errors, not HTTP 5xx responses 🤔

PR #450, opened 2026-04-10
fix: return HTTP 401 for invalid Todoist API tokens in HTTP transport

scottlovegrove:
## Summary

- The [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization) is explicit that **"Invalid or expired tokens MUST receive a HTTP 401 response"** (Token Handling section), and that 401 responses **MUST** include a `WWW-Authenticate` header with `resource_metadata` URL (RFC 9728) so clients can discover the authorization server and re-authenticate
- We were violating this: 401 errors from the Todoist API were caught in the tool callback and returned as **HTTP 200** JSON-RPC responses with `isError: true` and the 401 buried in text content — MCP clients expecting HTTP 401 to trigger OAuth token refresh never saw it
- The root cause is architectural: the Streamable HTTP transport commits `HTTP 200` with SSE headers before tool execution begins, so the status code cannot be changed after the fact
- Fix: validate the Todoist API token **before** entering the MCP layer via Express middleware, returning a spec-compliant HTTP 401 with `WWW-Authenticate: Bearer resource_metadata="..."` if validation fails

## What changed

- **`src/utils/validate-todoist-token.ts`** — Validates a token by calling `TodoistApi.getUser()`. Returns `true` if valid, `false` on 401/403, throws on transient errors
- **`src/middleware/require-valid-todoist-token.ts`** — Express middleware with TTL-based caching, static/dynamic token source modes, spec-compliant 401 responses. Fails open on transient errors (network/5xx) to avoid blocking all requests during a Todoist API blip
- **`src/main-http.ts`** — Middleware integrated before the MCP POST handler
- **`src/index.ts`** — Exports `validateTodoistToken` and `requireValidTodoistToken` for use by the hosted service (`todoist-ai-integrations`)

## MCP spec references

From the [Authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization):

> **Token Handling**: "MCP servers MUST validate access tokens... If validation fails, servers MUST respond according to OAuth 2.1 Section 5.3 error handling requirements. Invalid or expired tokens MUST receive a HTTP 401 response."
>
> **Error Handling table**: `401 Unauthorized` — "Authorization required or token invalid"
>
> **Protected Resource Metadata Discovery**: "Include the resource metadata URL in the WWW-Authenticate HTTP header under resource_metadata when returning 401 Unauthorized responses"

## Test plan

- [ ] `npm run check` passes (types, formatting, schema validation)
- [ ] All 812 tests pass (including 16 new tests for validation and middleware)
- [ ] Manual test: start `main-http.ts` with invalid `TODOIST_API_KEY` → HTTP 401 with `WWW-Authenticate` header
- [ ] Manual test: start `main-http.ts` with valid key → MCP requests work normally
- [ ] Verify cached validation doesn't call Todoist API on every request

### DIFF
--- a/src/middleware/require-valid-todoist-token.ts
+++ b/src/middleware/require-valid-todoist-token.ts
@@ -38,22 +38,43 @@ type CacheEntry = {
 
 const DEFAULT_STATIC_TTL_MS = 300_000 // 5 minutes
 const DEFAULT_DYNAMIC_TTL_MS = 120_000 // 2 minutes
+const MAX_CACHE_SIZE = 10_000
 
 const cache = new Map<string, CacheEntry>()
 
-function getCachedResult(apiKey: string): boolean | undefined {
-    const entry = cache.get(apiKey)
+function makeCacheKey(apiKey: string, baseUrl?: string): string {
+    return baseUrl ? `${apiKey}::${baseUrl}` : apiKey
+}
+
+function pruneExpiredEntries(): void {
+    const now = Date.now()
+    for (const [key, entry] of cache) {
+        if (now >= entry.expiresAt) {
+            cache.delete(key)
+        }
+    }
+}
+
+function getCachedResult(cacheKey: string): boolean | undefined {
+    const entry = cache.get(cacheKey)
     if (entry && Date.now() < entry.expiresAt) {
         return entry.valid
     }
     if (entry) {
-        cache.delete(apiKey)
+        cache.delete(cacheKey)
     }
     return undefined
 }
 
-function setCachedResult(apiKey: string, valid: boolean, ttlMs: number): void {
-    cache.set(apiKey, { valid, expiresAt: Date.now() + ttlMs })
+function setCachedResult(cacheKey: string, valid: boolean, ttlMs: number): void {
+    if (cache.size >= MAX_CACHE_SIZE) {
+        pruneExpiredEntries()
+    }
+    // If still at capacity after pruning, clear everything
+    if (cache.size >= MAX_CACHE_SIZE) {
+        cache.clear()
+    }
+    cache.set(cacheKey, { valid, expiresAt: Date.now() + ttlMs })
 }
 
 function buildWwwAuthenticateHeader(resourceMetadataUrl?: string): string {
@@ -89,7 +110,8 @@ function requireValidTodoistToken(options: RequireValidTodoistTokenOptions): Req
             return
         }
 
-        const cached = getCachedResult(apiKey)
+        const cacheKey = makeCacheKey(apiKey, options.baseUrl)
+        const cached = getCachedResult(cacheKey)
         if (cached !== undefined) {
             if (cached) {
                 next()
@@ -101,7 +123,7 @@ function requireValidTodoistToken(options: RequireValidTodoistTokenOptions): Req
 
         try {
             const valid = await validateTodoistToken(apiKey, options.baseUrl)
-            setCachedResult(apiKey, valid, ttlMs)
+            setCachedResult(cacheKey, valid, ttlMs)
 
             if (valid) {
                 next()
--- a/src/middleware/require-valid-todoist-token.ts
+++ b/src/middleware/require-valid-todoist-token.ts
@@ -1,0 +1,127 @@
+import type { NextFunction, Request, RequestHandler, Response } from 'express'
+
+import { validateTodoistToken } from '../utils/validate-todoist-token.js'
+
+type StaticTokenSource = {
+    type: 'static'
+    apiKey: string
+    baseUrl?: string
+}
+
+type DynamicTokenSource = {
+    type: 'dynamic'
+    getApiKey: (req: Request) => string | undefined
+    baseUrl?: string
+}
+
+type TokenSource = StaticTokenSource | DynamicTokenSource
+
+type RequireValidTodoistTokenOptions = TokenSource & {
+    /**
+     * Cache TTL in milliseconds. Defaults to 300_000 (5 minutes) for static,
+     * 120_000 (2 minutes) for dynamic.
+     */
+    cacheTtlMs?: number
+
+    /**
+     * URL for the Protected Resource Metadata document (RFC 9728).
+     * Included in the WWW-Authenticate header when returning 401.
+     * If omitted, the header is sent without resource_metadata.
+     */
+    resourceMetadataUrl?: string
+}
+
+type CacheEntry = {
+    valid: boolean
+    expiresAt: number
+}
+
+const DEFAULT_STATIC_TTL_MS = 300_000 // 5 minutes
+const DEFAULT_DYNAMIC_TTL_MS = 120_000 // 2 minutes
+
+const cache = new Map<string, CacheEntry>()
+
+function getCachedResult(apiKey: string): boolean | undefined {
+    const entry = cache.get(apiKey)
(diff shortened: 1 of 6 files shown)


src/middleware/require-valid-todoist-token.ts
```diff
@@ -0,0 +1,127 @@
...
+        return entry.valid
+    }
+    if (entry) {
+        cache.delete(apiKey)
+    }
+    return undefined
+}
+
+function setCachedResult(apiKey: string, valid: boolean, ttlMs: number): void {
+    cache.set(apiKey, { valid, expiresAt: Date.now() + ttlMs })
```
doistbot: [P1] Unbounded memory growth in token cache

The `cache` Map has no size limit. Because invalid tokens are also cached (`valid: false`), a malicious client or buggy script repeatedly sending requests with unique, random invalid tokens (especially when used in `dynamic` mode by `todoist-ai-integrations`) will cause the cache to grow indefinitely until it causes an Out-of-Memory (OOM) crash. Expired entries are only removed if they are looked up again, which won't happen for unique random tokens.

Adding a simple size limit check before setting a new entry is an easy way to prevent this without needing an external LRU cache dependency.

```suggestion
    if (cache.size >= 10000) {
        cache.clear()
    }
    cache.set(apiKey, { valid, expiresAt: Date.now() + ttlMs })
```

src/utils/validate-todoist-token.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @returns `true` if the token is valid, `false` if it's an auth error (401/403).
+ * @throws On transient errors (network failures, 5xx) so the caller can decide.
+ */
+async function validateTodoistToken(apiKey: string, baseUrl?: string): Promise<boolean> {
+    const client = new TodoistApi(apiKey, { baseUrl })
+    try {
+        await client.getUser()
+        return true
+    } catch (error) {
+        const statusCode = extractHttpStatusCode(error)
```
doistbot: [P1] `extractHttpStatusCode()` only looks at top-level status fields or messages matching `HTTP 401`, so it misses common shapes like `{ response: { status: 401 } }` / `Request failed with status code 401`. In that case this function rethrows, the middleware treats it as a transient failure, and an actually invalid token still reaches the MCP handler as HTTP 200. Please use the broader status extraction logic already used for Todoist API errors, or extend this helper before classifying auth failures.

src/middleware/require-valid-todoist-token.ts
```diff
@@ -0,0 +1,127 @@
...
+
+type CacheEntry = {
+    valid: boolean
+    expiresAt: number
+}
+
+const DEFAULT_STATIC_TTL_MS = 300_000 // 5 minutes
+const DEFAULT_DYNAMIC_TTL_MS = 120_000 // 2 minutes
+
+const cache = new Map<string, CacheEntry>()
```
doistbot, line 43: [P2] This cache is process-global and expired entries are only removed when the same token is looked up again. In dynamic mode, a long-lived service that sees many one-off user tokens will retain one `Map` entry per token forever, so memory usage grows monotonically. Please prune expired entries opportunistically on insert/lookup or use a size-bounded/LRU cache.

src/middleware/require-valid-todoist-token.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    return async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+        const apiKey = options.type === 'static' ? options.apiKey : options.getApiKey(req)
+
+        if (!apiKey) {
+            sendUnauthorized(res, resourceMetadataUrl)
+            return
+        }
+
+        const cached = getCachedResult(apiKey)
```
doistbot: [P2] The cache key is just `apiKey`, but validation also depends on `options.baseUrl`. If the same token string is checked against two different Todoist base URLs in one process, the first result is reused for the second and can produce false 401s or false passes. Include `baseUrl` in the cache key, or scope the cache per middleware instance.

doistbot (commented):
This PR introduces a well-structured architectural fix to ensure invalid Todoist API tokens correctly return an HTTP 401 response before reaching the MCP transport layer. By moving token validation to Express middleware, this significantly improves compliance with the MCP authorization specification. However, a few edge cases in the caching and error-handling logic need refinement, specifically addressing unbounded memory growth in the token cache, incorporating the base URL into the cache key, and broadening the status code extraction to catch nested authorization errors.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-ai&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-ai%2Fpull%2F450&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0324f440-3508-11f1-9bd4-824c678924c8%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0324f440-3508-11f1-9bd4-824c678924c8&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/middleware/require-valid-todoist-token.ts
gnapse: I share this concern as well. LRU could be the way to go.

src/utils/validate-todoist-token.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @throws On transient errors (network failures, 5xx) so the caller can decide.
+ */
+async function validateTodoistToken(apiKey: string, baseUrl?: string): Promise<boolean> {
+    const client = new TodoistApi(apiKey, { baseUrl })
+    try {
+        await client.getUser()
+        return true
+    } catch (error) {
+        const statusCode = extractAuthStatusCode(error)
+        if (statusCode !== undefined && AUTH_ERROR_CODES.has(statusCode)) {
```
gnapse, line 38: Seems simpler to just do this:


```suggestion
        if (statusCode === 401 || statusCode === 403) {
```

We would not need the `AUTH_ERROR_CODES` set. And the condition reads more clearly. Maybe just keep constants for 401 and 403 but TBH, these numbers are so ubiquitous for developers that I would not even bother.

gnapse (commented):
nit picky comment below. feel free to ignore since this is already merged.

src/middleware/require-valid-todoist-token.ts
scottlovegrove: This is only for when you are running the local http module, it doesn't affect the hosted MCP server.

PR #454, opened 2026-04-14
feat: make find-project-collaborators discoverable as user-lookup

scottlovegrove:
## Summary

- Rewrite the `find-project-collaborators` tool description with user/person/teammate language and explicit example phrasings so LLM agents match it on "find a user" / "who is X" style prompts.
- Make `projectId` optional; when omitted, search across all shared projects via `userResolver.getAllCollaborators` (promoted from `private` to `public`) — no more forcing the agent to pick a project first.
- Update `mcp-server.ts` instructions to frame this as the user-lookup tool and add a **User Lookup** common workflow.
- Keep the tool name unchanged for backward compatibility (10+ call sites incl. tests, scripts, helpers would otherwise churn, and existing MCP clients would break).

Fixes the agent-discoverability problem raised in Doist/automations#2332 — Ernesto (gnapse) flagged that the name "collaborator" and the project-scoped description were misleading to LLMs looking up a "user".

## Test plan

- [x] `tsc --noEmit` clean
- [x] Full test suite passes (824/824)
- [x] `npm run format:check` clean
- [x] New `src/tools/__tests__/find-project-collaborators.test.ts` covers both branches (per-project and workspace-wide), searchTerm filtering, empty/non-shared cases
- [ ] Manual MCP probe: `npx tsx scripts/run-tool.ts find-project-collaborators '{"searchTerm":"Ernesto"}'`

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -97,7 +97,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 **Collaboration & Comments:**
 - **add-comments/update-comments/find-comments**: Manage task and project discussions
 - **view-attachment**: View file attachments from comments. Pass the fileUrl from a comment's fileAttachment. Returns images inline, text files as text, and binary files as embedded resources.
-- **find-project-collaborators**: Look up Todoist users (workspace members, collaborators, teammates) by name or email to get their user ID — use for "find/who is X" questions or any time you need to resolve a person's name to an ID. Searches the whole workspace by default; pass projectId to scope to a single project
+- **find-project-collaborators**: Look up Todoist users (collaborators, teammates) by name or email to get their user ID — use for "find/who is X" questions or any time you need to resolve a person's name to an ID. By default searches collaborators of every shared project the authenticated user can access (plus the authenticated user themselves). An empty result means the person is not a collaborator on any shared project, not that they do not exist. Pass projectId to scope to a single project
 - **manage-assignments**: Bulk assign/unassign/reassign up to 50 tasks with atomic operations and dry-run validation
 
 **Filters:**
@@ -127,7 +127,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 
 2. **Search Strategy**: Use specific search queries combining multiple filters for precise results. When searching for tasks, start with broader queries and narrow down as needed.
 
-3. **Assignments & user lookup**: Always validate project collaborators exist before assigning tasks. Use find-project-collaborators to verify user access. Also use find-project-collaborators (with just a searchTerm and no projectId) to resolve a user's ID whenever the user references a person by name or email — it searches across the whole workspace.
+3. **Assignments & user lookup**: Always validate project collaborators exist before assigning tasks. Use find-project-collaborators to verify user access. Also use find-project-collaborators (with just a searchTerm and no projectId) to resolve a user's ID whenever the user references a person by name or email — it searches collaborators of all shared projects you can access, plus yourself.
 
 4. **Bulk Operations**: When working with multiple items, prefer bulk tools (complete-tasks, manage-assignments) over individual operations for better performance.
 
@@ -143,7 +143,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 
 - **Daily Planning**: Use find-tasks-by-date with 'today' and get-overview for project status
 - **Team Assignment**: find-project-collaborators → add-tasks with responsibleUser → manage-assignments for bulk changes
-- **User Lookup**: find-project-collaborators with just a searchTerm (no projectId) to resolve a name or email to a Todoist user ID across the entire workspace
+- **User Lookup**: find-project-collaborators with just a searchTerm (no projectId) to resolve a name or email to a Todoist user ID across all shared-project collaborators you can access
 - **Task Search**: find-tasks with multiple filters → update-tasks or complete-tasks based on results
 - **Project Organization**: add-projects → add-sections → add-tasks with projectId and sectionId
 - **Progress Reviews**: find-completed-tasks (defaults to last 7 days; optionally use explicit date ranges) → get-overview for project summaries
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -97,7 +97,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 **Collaboration & Comments:**
 - **add-comments/update-comments/find-comments**: Manage task and project discussions
 - **view-attachment**: View file attachments from comments. Pass the fileUrl from a comment's fileAttachment. Returns images inline, text files as text, and binary files as embedded resources.
-- **find-project-collaborators**: Find team members by name or email for assignments
+- **find-project-collaborators**: Look up Todoist users (workspace members, collaborators, teammates) by name or email to get their user ID — use for "find/who is X" questions or any time you need to resolve a person's name to an ID. Searches the whole workspace by default; pass projectId to scope to a single project
 - **manage-assignments**: Bulk assign/unassign/reassign up to 50 tasks with atomic operations and dry-run validation
 
 **Filters:**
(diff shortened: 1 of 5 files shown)


src/tools/find-project-collaborators.ts
```diff
@@ -173,26 +240,29 @@ function generateTextContent({
...
         if (searchTerm) {
-            zeroReasonHints.push(`No collaborators match "${searchTerm}"`)
+            zeroReasonHints.push(`No users match "${searchTerm}"`)
             zeroReasonHints.push('Try a broader search term or check spelling')
             if (totalAvailable > 0) {
-                zeroReasonHints.push(`${totalAvailable} collaborators available without filter`)
+                zeroReasonHints.push(`${totalAvailable} users available without filter`)
             }
-        } else {
+        } else if (projectName) {
```
doistbot: [P2] Dead code: The `else if (projectName)` and `else` blocks are unreachable.

Both `findProjectCollaborators.execute()` and `executeWorkspaceSearch()` return early when no collaborators are found in the project or workspace (e.g., via `generateEmptyProjectResponse`). This means `generateTextContent` is only called when the initial list of collaborators is non-empty. If `collaborators.length === 0` inside this function, it is exclusively because the `searchTerm` filtered everyone out. Therefore, `searchTerm` is guaranteed to be truthy here.

You can simplify this block by removing the unreachable branches:

```typescript
    const zeroReasonHints: string[] = []
    if (collaborators.length === 0 && searchTerm) {
        zeroReasonHints.push(`No users match "${searchTerm}"`)
        zeroReasonHints.push('Try a broader search term or check spelling')
        if (totalAvailable > 0) {
            zeroReasonHints.push(`${totalAvailable} users available without filter`)
        }
    }
```

src/utils/user-resolver.ts
```diff
@@ -217,9 +217,9 @@ export class UserResolver {
     }
 
     /**
-     * Get all collaborators from all shared projects
+     * Get all collaborators from all shared projects, deduplicated by user ID.
      */
-    private async getAllCollaborators(client: TodoistApi): Promise<ProjectCollaborator[]> {
+    async getAllCollaborators(client: TodoistApi): Promise<ProjectCollaborator[]> {
```
doistbot, line 223: [P1] `getAllCollaborators()` now backs the new workspace-wide lookup, but it still does a single `client.getProjects({})` call internally. `getProjects` is paginated elsewhere in this codebase, so any account with more than one page of projects will silently miss collaborators from later pages and fail some "find user" lookups. Please page through all projects here (for example with the existing pagination helper) before collecting shared-project collaborators.

src/tools/find-project-collaborators.ts
```diff
@@ -136,26 +140,89 @@ const findProjectCollaborators = {
...
+async function executeWorkspaceSearch({
+    searchTerm,
+    client,
+    appliedFilters,
+}: {
+    searchTerm: string | undefined
+    client: TodoistApi
+    appliedFilters: Record<string, unknown>
+}) {
+    const allCollaborators = await userResolver.getAllCollaborators(client)
```
doistbot: [P2] This new user-lookup branch is narrower than the existing resolver logic. `resolveUser()` prepends `client.getUser()` so the authenticated user can still be found even when there are no shared projects, but this path only uses `getAllCollaborators()`. On personal accounts or any setup without shared projects, asking for your own user ID by name/email now returns an empty result. Consider adding the current user here as well, or reusing the same source of truth as `resolveUser()`.

src/mcp-server.ts
```diff
@@ -97,7 +97,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 **Collaboration & Comments:**
 - **add-comments/update-comments/find-comments**: Manage task and project discussions
 - **view-attachment**: View file attachments from comments. Pass the fileUrl from a comment's fileAttachment. Returns images inline, text files as text, and binary files as embedded resources.
-- **find-project-collaborators**: Find team members by name or email for assignments
+- **find-project-collaborators**: Look up Todoist users (workspace members, collaborators, teammates) by name or email to get their user ID — use for "find/who is X" questions or any time you need to resolve a person's name to an ID. Searches the whole workspace by default; pass projectId to scope to a single project
```
doistbot: [P2] The prompt now promises broader behavior than the code implements. The new branch only aggregates collaborators from shared projects, but these instructions say the tool searches the "whole workspace" / "entire workspace". That makes an empty result look authoritative to the model when it may only mean "not found in shared-project collaborators". Please keep the wording aligned with the actual scope.

doistbot (commented):
This PR effectively repurposes the project collaborators tool to act as a more discoverable, general user-lookup by making the project ID optional and updating the LLM instructions. This is a thoughtful approach to improving agent discoverability without breaking backward compatibility for existing clients. However, a few adjustments are needed to ensure completeness, including adding pagination to the underlying project fetch, including the authenticated user for isolated accounts, pruning some unreachable filtering logic, and aligning the prompt wording more accurately with the tool's actual shared-project scope.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-ai&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-ai%2Fpull%2F454&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A082a01d0-3817-11f1-9a1a-449d221d782e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A082a01d0-3817-11f1-9a1a-449d221d782e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/utils/user-resolver.ts
scottlovegrove: Fixed in 5cff8ba — `getAllCollaborators` now uses the shared `fetchAllPages` helper to walk every page of `getProjects`. Added a pagination test in `user-resolver.test.ts`.

src/tools/find-project-collaborators.ts
scottlovegrove: Fixed in 5cff8ba — `executeWorkspaceSearch` now prepends the authenticated user (via a local `prependCurrentUser` helper) before filtering, matching `resolveUser()`'s behaviour. Added tests covering the personal-account lookup and the no-duplicate case when the user is already in the collaborator list.

src/tools/find-project-collaborators.ts
scottlovegrove: Fixed in 5cff8ba — collapsed the dead branches into a single `if (collaborators.length === 0 && searchTerm)` and added a comment explaining that upstream callers handle the truly-empty case.

src/mcp-server.ts
scottlovegrove: Fixed in 5cff8ba — reworded the mcp-server bullet, the tool description, and the Best Practice / Common Workflow entries to say "collaborators of all shared projects you can access (plus yourself)" instead of "whole/entire workspace", and added an explicit note that an empty result means "not a collaborator" rather than "does not exist".

craigcarlyle:
✅

PR #458, opened 2026-04-22
docs: Add one-click "Add to Kiro" install badge

ykethan:
# Pull Request

Closes #

## Short description

Adds an "Add to Kiro" one-click install badge to the README for [Kiro](https://kiro.dev), an AI-powered IDE.

The badge auto-configures the Todoist MCP server in Kiro with no manual JSON editing. A dedicated Kiro section is added under the Setup Guide. Both remote HTTP and npx transport configs are included as manual fallbacks.

Only the README is modified — no code or tool changes.

## PR Checklist

- [ ] Added tests for bugs / new features

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,42 @@ This will take you through a wizard to authenticate using your browser with Todo
 }
 ```
 
+#### Kiro
+
+[![Add to Kiro](https://kiro.dev/images/add-to-kiro.svg)](https://kiro.dev/launch/mcp/add?name=todoist&config=%7B%22url%22%3A%22https%3A%2F%2Fai.todoist.net%2Fmcp%22%7D)
+
+Create or edit a configuration file:
+
+- **Global:** `~/.kiro/settings/mcp.json`
+- **Project-specific:** `.kiro/settings/mcp.json`
+
+Using the remote URL (recommended):
+
+```json
+{
+    "mcpServers": {
+        "todoist": {
+            "url": "https://ai.todoist.net/mcp"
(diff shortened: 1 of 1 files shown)


gnapse:
Thanks for your contribution.

However, I'm reluctant to add this. We cannot afford to put specific instructions about any conceivable MCP client app, plus these instructions are pretty much the same as for other client apps, except for the path to the config file being specific about kiro. It is each client app’s responsibility (in this case Kiro's responsibility) to document how to configure MCP servers on it, and I'm sure they do document it for their users.

We just provide a few for the common ones, for illustrative purposes.

ykethan:
The main value-add here is actually the one-click badge itself rather than the setup instructions. When a user clicks "Add to Kiro," the server is auto-configured with no manual steps at all — similar to how the Cursor and Smithery badges work at the top of the README.

Would you be open to a slimmed-down version that just adds the badge inline with the existing ones (Cursor, Smithery, etc.) without a dedicated Kiro section? Happy to update the PR if so, or close it if you'd rather not — no hard feelings either way.

gnapse:
> adds the badge inline with the existing ones (Cursor, Smithery, etc.)

What badges do you mean? I see no badges in our README for Cursor, Smithery, etc. 🤔

scottlovegrove:
I'm going to close this. Having individual badges for each and every MCP client will become noisy and we'll have to decide at what point should we not have them included. Better to just not include any. 

Thanks for raising the PR though @ykethan

PR #467, opened 2026-05-03
fix: surface view-attachment file body in content blocks

### LINKED ISSUE #466: view-attachment returns metadata only — file content missing from MCP response
The view-attachment tool authenticates successfully and returns a JSON metadata envelope (with a contentDelivery field correctly set to "text" for HTML files), but no actual file content is included in the MCP response. The fix in #398 resolved the auth problem, but the content packaging step appears to be broken in the deployed server.
Steps to reproduce

- Forward an email into Todoist (creates a text/html file attachment on the task)
- Call view-attachment with the fileUrl from the comment's fileAttachment field
- Observe the response

**Expected result**
A text content block containing the file's content, as described in the PR #398 commit message:

> "Returns images inline (ImageContent), text files as text, and binary files like PDFs as embedded resources."

**Actual result**
`{"fileName":"file.html","fileType":"text/html","fileSize":17244,"contentDelivery":"text"}`

scottlovegrove:
## Summary

- Fixes #466. `view-attachment` was returning the file body in `content[]` blocks **and** a structured-content metadata envelope; MCP clients that prefer `structuredContent` (e.g. ChatGPT, Claude Code) surfaced only the metadata and the file body looked missing.
- Drop `outputSchema`/`structuredContent` from `view-attachment` so the typed image/text/resource content blocks become the canonical answer.
- Make `outputSchema` optional on the `TodoistTool` interface (no behavior change for other tools, all already declare one).
- `registerTool()` now omits `outputSchema` when undefined.
- Replace deprecated `z.string().url()` with `z.url()` in the args schema.

## Why content blocks, not a richer structured schema

`content[]` is the MCP-spec'd carrier for media: `image` (inline render), `resource` (embedded with mimeType/uri), `text` (raw body). Cramming base64 image/binary into `structuredContent` loses the rendering hint and bloats JSON. Text-only would split the response shape across two carriers.

## Test plan

- [x] `npm run type-check` — clean
- [x] `npm test` — 841/841 pass
- [x] `npm run format:check` — clean
- [x] Local repro via `scripts/run-tool.ts view-attachment` — HTML body present in `contentItems[0]`, no structured envelope
- [x] Raw HTTP `tools/call` against deployed server — confirmed prior shape included redundant metadata text + structuredContent; new shape leaves only header text + file body
- [ ] Post-deploy: re-run the original failing case from #466 (forwarded-email HTML attachment) through a real MCP client; confirm the body is rendered

### DIFF
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -260,7 +260,7 @@ function registerTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape
     const toolConfig = {
         description: tool.description,
         inputSchema: tool.parameters,
-        outputSchema: tool.outputSchema as Output,
+        ...(tool.outputSchema ? { outputSchema: tool.outputSchema as Output } : {}),
         annotations: getMcpAnnotations(tool),
         ...(tool._meta ? { _meta: tool._meta } : {}),
     }
--- a/src/todoist-tool.ts
+++ b/src/todoist-tool.ts
@@ -17,7 +17,7 @@ type RequiredToolAnnotations = ToolAnnotations & {
 /**
  * A Todoist tool that can be used in an MCP server or other conversational AI interfaces.
  */
-type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape> = {
+type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape = z.ZodRawShape> = {
     /**
      * The name of the tool.
      */
@@ -42,7 +42,7 @@ type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape> = {
      *
      * This is used to describe the structured output format that the tool will return.
      */
-    outputSchema: Output
+    outputSchema?: Output
 
     /**
      * MCP ToolAnnotations hints for this tool.
--- a/src/todoist-tool.ts
+++ b/src/todoist-tool.ts
@@ -17,7 +17,10 @@ type RequiredToolAnnotations = ToolAnnotations & {
 /**
  * A Todoist tool that can be used in an MCP server or other conversational AI interfaces.
  */
-type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape = z.ZodRawShape> = {
+type TodoistTool<
+    Params extends z.ZodRawShape,
+    Output extends z.ZodRawShape = Record<string, never>,
+> = {
     /**
      * The name of the tool.
      */
--- a/src/tools/__tests__/view-attachment.test.ts
+++ b/src/tools/__tests__/view-attachment.test.ts
@@ -243,6 +243,7 @@ describe('view-attachment tool', () => {
                 type: 'text',
                 text: '<h1>Hello</h1>',
             })
+            expect((result as Record<string, unknown>).structuredContent).toBeUndefined()
         })
     })
 
@@ -418,7 +419,7 @@ describe('view-attachment tool', () => {
                 type: 'text',
                 text: 'hello',
             })
-            expect(result.textContent).toContain('KB')
+            expect(result.textContent).toContain('0.0KB')
         })
     })
 
--- a/src/tools/__tests__/view-attachment.test.ts
+++ b/src/tools/__tests__/view-attachment.test.ts
@@ -100,9 +100,8 @@ describe('view-attachment tool', () => {
                 data: Buffer.from(imageData).toString('base64'),
                 mimeType: 'image/png',
             })
-            expect(result.structuredContent.contentDelivery).toBe('image')
-            expect(result.structuredContent.fileType).toBe('image/png')
-            expect(result.structuredContent.fileName).toBe('file.png')
+            expect(result.textContent).toContain('file.png')
+            expect(result.textContent).toContain('image/png')
         })
 
         it('should return ImageContent for JPEG files', async () => {
@@ -123,7 +122,6 @@ describe('view-attachment tool', () => {
                 type: 'image',
                 mimeType: 'image/jpeg',
             })
-            expect(result.structuredContent.contentDelivery).toBe('image')
         })
 
         it('should return ImageContent for GIF files', async () => {
@@ -185,8 +183,7 @@ describe('view-attachment tool', () => {
                 type: 'text',
                 text: textContent,
             })
-            expect(result.structuredContent.contentDelivery).toBe('text')
-            expect(result.structuredContent.fileType).toBe('text/plain')
+            expect(result.textContent).toContain('text/plain')
         })
 
         it('should return TextContent for JSON files', async () => {
@@ -207,7 +204,6 @@ describe('view-attachment tool', () => {
                 type: 'text',
                 text: jsonContent,
             })
-            expect(result.structuredContent.contentDelivery).toBe('text')
         })
 
         it('should return TextContent for CSV files', async () => {
@@ -243,7 +239,10 @@ describe('view-attachment tool', () => {
                 mockTodoistApi,
             )
 
-            expect(result.structuredContent.contentDelivery).toBe('text')
+            expect(result.contentItems?.[0]).toEqual({
+                type: 'text',
+                text: '<h1>Hello</h1>',
+            })
         })
     })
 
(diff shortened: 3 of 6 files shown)


src/tools/view-attachment.ts
```diff
@@ -77,28 +77,17 @@ function getFileNameFromUrl(url: string): string | undefined {
 
 const ArgsSchema = {
     fileUrl: z
-        .string()
```
doistbot, line 80: [P0] Do not remove `.string()`. Zod does not expose a `.url()` method directly on the `z` object (unlike Joi or Yup). URL validation must be chained off a string type via `z.string().url()`. Removing this will throw a `TypeError: z.url is not a function` when `ArgsSchema` is evaluated at module load time, crashing the server on startup.

src/todoist-tool.ts
```diff
@@ -17,7 +17,7 @@ type RequiredToolAnnotations = ToolAnnotations & {
 /**
  * A Todoist tool that can be used in an MCP server or other conversational AI interfaces.
  */
-type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape> = {
+type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape = z.ZodRawShape> = {
```
doistbot: [P2] Defaulting `Output` to `z.ZodRawShape` creates a subtle type trap. If a developer adds a new tool with an `outputSchema` but forgets to explicitly pass its type (e.g., `satisfies TodoistTool<typeof ArgsSchema>`), `Output` silently falls back to `z.ZodRawShape`.

This causes `outputSchema` to accept the schema without errors, but `structuredContent` in the `execute` method will loosely typecheck as `{ [x: string]: any }`—completely bypassing output type safety.

Changing the default to `Record<string, never>` strictly prevents this:
```typescript
type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape = Record<string, never>> = {
```
If the generic is omitted while an `outputSchema` is provided, TypeScript will throw a type error (e.g., `Type 'ZodString' is not assignable to type 'never'`), forcing the developer to provide the `Output` generic and preserving type safety.

src/todoist-tool.ts
```diff
@@ -42,7 +42,7 @@ type TodoistTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape> = {
      *
      * This is used to describe the structured output format that the tool will return.
      */
-    outputSchema: Output
+    outputSchema?: Output
```
doistbot, line 48: [P2] This makes `outputSchema` optional even when a tool supplies an `Output` generic, so a structured-output tool can now omit the schema while still returning typed `structuredContent`; `registerTool` will silently publish it without `outputSchema`. Preserve the invariant with a union/conditional type where schema-less tools have `outputSchema?: never` and `structuredContent?: never`, while schema-backed tools still require `outputSchema: Output`.

src/tools/__tests__/view-attachment.test.ts
```diff
@@ -402,11 +398,10 @@ describe('view-attachment tool', () => {
             )
 
             expect(result.contentItems).toBeUndefined()
-            expect(result.structuredContent.contentDelivery).toBe('metadata_only')
             expect(result.textContent).toContain('too large')
         })
 
-        it('should use actual body size for fileSize in structured content', async () => {
+        it('should use actual body size for fileSize in text content', async () => {
```
doistbot, line 405: [P2] With the removal of the `result.structuredContent.fileSize` assertion, this test no longer verifies its description ("should use actual body size"). The remaining `toContain('KB')` assertion on line 421 is too broad. Consider adding `expect(result.textContent).toContain('0.0KB')` to precisely verify the size calculation for the 5-byte `'hello'` string.

src/mcp-helpers.ts
```diff
@@ -260,7 +260,7 @@ function registerTool<Params extends z.ZodRawShape, Output extends z.ZodRawShape
     const toolConfig = {
         description: tool.description,
         inputSchema: tool.parameters,
-        outputSchema: tool.outputSchema as Output,
+        ...(tool.outputSchema ? { outputSchema: tool.outputSchema as Output } : {}),
```
doistbot, line 263: [P2] The new no-`outputSchema` registration path is untested. Add a `registerTool` test with a tool that omits `outputSchema` and assert the config passed to `server.registerTool` does not have an own `outputSchema` property, so this doesn't regress back to advertising a structured schema for content-block-only tools.

src/tools/__tests__/view-attachment.test.ts
```diff
@@ -243,7 +239,10 @@ describe('view-attachment tool', () => {
                 mockTodoistApi,
             )
 
-            expect(result.structuredContent.contentDelivery).toBe('text')
+            expect(result.contentItems?.[0]).toEqual({
```
doistbot, line 242: [P2] This representative content-block assertion would still pass if `viewAttachment.execute()` returned both `contentItems` and the old `structuredContent`, which is the regression this PR fixes for MCP clients. Add an assertion here such as `expect(result.structuredContent).toBeUndefined()` so the test fails if the structured envelope is reintroduced.

src/todoist-tool.ts
doistbot: [P2] Modifying the `TodoistTool` contract (making `outputSchema` optional) is an explicit trigger in `CLAUDE.md` for keeping `CODEBASE.md` accurate. Please update the `TodoistTool` code block example in `CODEBASE.md` to reflect this new signature.

doistbot (commented):
This PR successfully addresses the issue with hidden file bodies in `view-attachment` by removing `structuredContent` and adjusting the `TodoistTool` interface to allow optional output schemas. Relying purely on standard content blocks is a great approach that will significantly improve rendering compatibility across different MCP clients. A few adjustments have been flagged for review, including reverting the Zod URL syntax to avoid runtime crashes, tightening the TypeScript generics to preserve strict type safety, adding test assertions for the new schema-less behavior, and keeping the `TodoistTool` documentation up to date in CODEBASE.md.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-ai&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-ai%2Fpull%2F467&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A65b2f780-471c-11f1-9248-9a1833679308%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A65b2f780-471c-11f1-9248-9a1833679308&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/view-attachment.ts
scottlovegrove: Skipping this revert — the claim is incorrect for Zod v4. The repo is on `zod@4.3.6` (upgraded in #236), and `z.url()` is a top-level schema constructor in Zod 4 (`typeof z.url === 'function'`). The deprecated form is actually `z.string().url()`, which is exactly the source of the deprecation warning the IDE flagged on line 81. Verified locally: `npm run type-check` and the full test suite (843/843) both pass with `z.url()`. See the [Zod v4 migration notes](https://zod.dev/v4/changelog#zstringurl) — `z.string().url()` was moved out into top-level schemas like `z.url()`.

craigcarlyle:
✅

PR #470, opened 2026-05-06
fix: strip section response fields to match output schema

scottlovegrove:
## Summary

- `add-sections`, `update-sections`, and `get-overview` declare section objects in their output schemas as `{id, name}` (via `SectionSchema`), but the structured responses included the full SDK `Section` (`userId`, `projectId`, `sectionOrder`, `isArchived`, `isDeleted`, `isCollapsed`, `url`).
- Strict MCP clients (e.g. Perplexity's structured-content validation with `additionalProperties: false`) rejected the responses.
- Map sections to `{id, name}` before returning so the runtime payload matches the declared schema.
- In `get-overview`, also collapse the duplicated inline section schemas to the shared `SectionSchema` and infer `SectionSummary` from it (single source of truth).
- `find-sections` was already stripping correctly — left untouched.

## Test plan

- [x] `npx tsc --noEmit`
- [x] Full vitest run (853 tests)
- [ ] Verify in a strict MCP client that section responses no longer fail structured-content validation

### DIFF
--- a/src/tools/add-sections.ts
+++ b/src/tools/add-sections.ts
@@ -2,7 +2,7 @@ import type { Section } from '@doist/todoist-sdk'
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { isInboxProjectId, resolveInboxProjectId } from '../tool-helpers.js'
-import { SectionSchema as SectionOutputSchema } from '../utils/output-schemas.js'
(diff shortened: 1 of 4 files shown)


scottlovegrove:
@doistbot /review

henningmu (commented):
✅

PR #477, opened 2026-05-13
fix: forward arrayBuffer on tracked-fetch wrapper for binary attachments

### LINKED ISSUE #476: [BUG] Claude plugin: `view-attachment` throws 400 error
## tl;dr from a human

I installed the Claude plugin but it throws a 400 error when trying to fetch images with `view-attachment`. Looks like `ai.todoist.net/mcp` might have an encoding issue. 

I spent some time investigating with Claude. His summary is below in case it's helpful. If there's anything else you need, please let me know.

## Claude's summary of our investigation

The `view-attachment` tool reliably crashes my conversation with `400 Could not process image` whenever it's called against a real file attachment. Below is what I found, plus the concrete decoded-bytes evidence — happy to provide more if useful.

### Repro

Against a Todoist task with at least one image attachment, call `view-attachment` with a `fileUrl` pointing at `files.todoist.com/user_upload/v2/.../file.png` (or `.jpeg`, or `.pdf`). The tool returns an `image` content block whose `data` is valid base64, but the bytes that base64 decodes to are not the original file.

I reproduced it on PNG, JPEG, and PDF attachments. Three test files:

| Type | Source size | fileUrl |
| --- | --- | --- |
| PNG | 7,049 B | `https://files.todoist.com/user_upload/v2/2242230082/file.png` |
| PDF | 620,876 B | `https://files.todoist.com/user_upload/v2/2242230095/file.pdf` |
| JPEG | 876,479 B | `https://files.todoist.com/user_upload/v2/2242230115/file.jpeg` |

### The corruption: textbook UTF-8 lossy round-trip on binary

For the tiny PNG, the base64 string the MCP returned starts with:

```
77+9UE5HDQoaCgAAAA1JSERSAAAAWgAAAFAIBgAAAO+/vTNiZAAAAAFzUkdCAO+/
```

That base64 decodes to:

```
EF BF BD 50 4E 47 0D 0A 1A 0A 00 00 00 0D 49 48 44 52 ...
00 EF BF BD 33 62 ...
```

A real PNG starts with `89 50 4E 47 0D 0A 1A 0A`. The leading `0x89` (and every other byte ≥ `0x80` throughout the file) has been replaced with `EF BF BD`, which is the 3-byte UTF-8 encoding of U+FFFD (the Unicode replacement character).

This is the unmistakable signature of running the file's binary bytes through `bytes.decode("utf-8", errors="replace")` (or any equivalent lossy-text decode), then re-encoding to bytes, *before* base64-encoding.

Size confirms it: the returned base64 is 16,984 chars → 12,736 decoded bytes vs. the 7,049-byte source. Every byte ≥ `0x80` that doesn't form part of a valid UTF-8 sequence has been replaced with `EF BF BD` (net +2 bytes each); bytes that happen to form valid UTF-8 sequences (e.g. `C2 A0`) round-trip unchanged. That's why the size gap is consistent with thousands of such substitutions.

### Why this kills downstream consumers

Claude's API receives a content block with `media_type: image/png` and `data: <valid base64>`. The base64 *is* valid; it just decodes to non-PNG bytes. The image processor rejects the block with `400 Could not process image`, which surfaces in Claude Code as an unrecoverable session error — every retry from the existing session re-sends the same corrupt block and crashes again. (Clearing the app cache and restarting didn't help, because the broken bytes are persisted in the session transcript.)

### What I verified

1. **The CDN file is clean.** I downloaded each `fileUrl` directly with `curl -H "Authorization: Bearer $TODOIST_API_TOKEN"`. Magic bytes are canonical (`89 50 4E 47` for PNG, `%PDF-1.7` for PDF, `FF D8 FF E1` for JPEG), sizes match the `fileSize` from `find-comments`. So the corruption isn't upstream of the MCP.
2. **The MCP source at HEAD looks correct.** [`src/tools/view-attachment.ts`](https://github.com/Doist/todoist-ai/blob/main/src/tools/view-attachment.ts) does `Buffer.from(await response.arrayBuffer()).toString('base64')`, which is the right binary read. Only two commits ever touched that file (#398 add, #467 wrapping fix), and I didn't find an open issue tracking this.
3. **The corruption is therefore between the MCP's underlying HTTP read and the bytes that reach `toString('base64')`** — but only in the deployed/hosted server, since the source at HEAD wouldn't produce this.

### Most likely root cause

One of:

- **The hosted server at `ai.todoist.net/mcp` is running an older revision** in which the file body was read with `response.text()` (or equivalent) before being base64-encoded. `response.text()` decodes the body as UTF-8 with `errors=replace`, producing exactly the byte pattern observed.
- **The Doist internal API client behind `client.viewAttachment` reads the body as a string before handing it to the tool callback.** In that case the HEAD source is "correct" in isolation but receives an already-corrupted buffer.

I'd start by grepping the deployed image-fetching path for `response.text(`, `.text()`, `toString('utf8'/'utf-8')`, or `readAsText`.

### Available debugging artifacts (happy to share)

None of these are sensitive — they're all about a throwaway test task:

- The full 16,984-char base64 string the MCP returned (from the Claude Code session JSONL).
- The clean reference PNG (7,049 B, sha256 `d295ef1109fc6d93d30653b177dddc79b895eb4790076064ef105b9f00f8b265`) downloaded directly from `files.todoist.com`.
- The `tool_use` input and `tool_result` content blocks captured from the Claude Code session JSONL for cross-referencing.

[todoist-mcp-debug.zip](https://github.com/user-attachments/files/27730524/todoist-mcp-debug.zip)

### Environment

- Calling `view-attachment` via the Doist-published, Anthropic-hosted MCP plugin (i.e., `ai.todoist.net/mcp`), not a self-hosted local copy.
- All other tools in the plugin I exercised (`find-tasks`, `find-comments`, `add-comments`, `update-tasks`, etc.) worked correctly — only `view-attachment` is affected.

scottlovegrove:
## Summary

- Bumps `@doist/todoist-sdk` to **10.1.5** ([Doist/todoist-sdk-typescript#602](https://github.com/Doist/todoist-sdk-typescript/pull/602)).
- Forwards `arrayBuffer()` from the tracked-fetch wrapper in `src/usage-tracking.ts` so the SDK's binary endpoints (`viewAttachment`, `downloadBackup`) get a binary-safe response instead of falling back to the pre-10.1.5 lossy `text()` → `TextEncoder` round-trip.

## Bug

Reported in #476. Every byte ≥ `0x80` in a Todoist file attachment was replaced with `EF BF BD` (UTF-8 replacement char U+FFFD) by the time it reached Claude's image processor — surfacing as `400 Could not process image` for every PNG/JPEG/PDF returned by `view-attachment` on the hosted `ai.todoist.net/mcp` deployment.

Root cause was a cooperating pair: the SDK's `UploadClient.viewAttachment` customFetch branch read the body via `response.text()` (UTF-8 decode with `errors=replace`) and re-encoded with `TextEncoder` (fixed in [Doist/todoist-sdk-typescript#602](https://github.com/Doist/todoist-sdk-typescript/pull/602)), and this repo's `toCustomFetchResponse` wrapper only forwarded `text/json/headers`, never giving the SDK a chance to read the body as bytes.

## End-to-end verification

Local run against a real 3.38 MB PNG attachment via `createTodoistClient` (the path used by the hosted MCP):

| Metric | Before | After |
| --- | --- | --- |
| Decoded length | 6,267,715 B (1.85×) | 3,382,996 B (exact) |
| Magic bytes | `EF BF BD 50 4E 47 0D 0A …` | `89 50 4E 47 0D 0A 1A 0A` ✅ |
| `EF BF BD` count in payload | 1,472,311 | 0 ✅ |

## Test plan

- [x] `npm run type-check` — clean.
- [x] `npm test` — 859/859 pass.
- [x] `npm run format:check` — clean.
- [x] End-to-end: `view-attachment` against a real PNG comment URL, decoded base64, magic bytes match, no replacement chars.
- [ ] Post-deploy: rerun the `tools/call view-attachment` curl from #476 against `ai.todoist.net/mcp` and verify `EF BF BD` count is 0.

Closes #476.

### DIFF
--- a/src/usage-tracking.ts
+++ b/src/usage-tracking.ts
@@ -195,6 +195,7 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
         headers: Object.fromEntries(response.headers.entries()),
         text: () => response.text(),
         json: () => response.json(),
+        arrayBuffer: () => response.arrayBuffer(),
     }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "8.12.2",
             "license": "MIT",
             "dependencies": {
-                "@doist/todoist-sdk": "10.1.1",
+                "@doist/todoist-sdk": "10.1.5",
                 "@modelcontextprotocol/ext-apps": "1.2.2",
                 "date-fns": "4.1.0",
                 "dompurify": "3.3.3",
@@ -461,9 +461,9 @@
             }
         },
         "node_modules/@doist/todoist-sdk": {
-            "version": "10.1.1",
-            "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-10.1.1.tgz",
-            "integrity": "sha512-V60AjoJG5QjiJ2iYB8IIC4hLA4B5PLUybYysc1GU2qC3xjN7Zy3juRukKo8PO+O1VMXeb8EACMri5U/0GPnLkA==",
+            "version": "10.1.5",
+            "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-10.1.5.tgz",
+            "integrity": "sha512-32e1D4qMRaGsQJt0IPM/X2eTUpjMjQRRHXK/RDSIRVH2015pTo0oakzA+bUcoIRYy+BfWLGw0d9Gowg2oP+EkA==",
             "license": "MIT",
             "dependencies": {
                 "camelcase": "6.3.0",
@@ -471,7 +471,7 @@
                 "form-data": "4.0.5",
                 "ts-custom-error": "^3.2.0",
                 "undici": "^7.16.0",
-                "uuid": "11.1.0",
+                "uuid": "11.1.1",
                 "zod": "4.3.6"
             },
             "engines": {
@@ -13966,9 +13966,9 @@
             "license": "MIT"
         },
         "node_modules/uuid": {
-            "version": "11.1.0",
-            "resolved": "https://registry.npmjs.org/uuid/-/uuid-11.1.0.tgz",
-            "integrity": "sha512-0/A9rDy9P7cJ+8w1c9WD9V//9Wj15Ce2MPz8Ri6032usz+NfePxx5AcN3bN+r6ZL6jEo066/yNYB3tn4pQEx+A==",
+            "version": "11.1.1",
+            "resolved": "https://registry.npmjs.org/uuid/-/uuid-11.1.1.tgz",
+            "integrity": "sha512-vIYxrBCC/N/K+Js3qSN88go7kIfNPssr/hHCesKCQNAjmgvYS2oqr69kIufEG+O4+PfezOH4EbIeHCfFov8ZgQ==",
             "funding": [
                 "https://github.com/sponsors/broofa",
                 "https://github.com/sponsors/ctavan"
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
         "prepare": "husky"
     },
     "dependencies": {
-        "@doist/todoist-sdk": "10.1.1",
+        "@doist/todoist-sdk": "10.1.5",
         "@modelcontextprotocol/ext-apps": "1.2.2",
         "date-fns": "4.1.0",
         "dompurify": "3.3.3",
--- a/src/usage-tracking.test.ts
+++ b/src/usage-tracking.test.ts
@@ -102,6 +102,30 @@ describe('usage tracking', () => {
         expect(capturedSessionIds[0]).not.toBe(capturedSessionIds[1])
     })
 
+    it('forwards arrayBuffer so binary attachments are not corrupted', async () => {
+        // PNG magic + a stretch of high bytes that would be replaced with
+        // U+FFFD by any UTF-8 text round-trip on the response body.
+        const binaryBytes = new Uint8Array([
+            0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x00, 0x00, 0x0d, 0xff, 0xfe,
+            0xfd, 0x80,
+        ])
+        const trackedFetch = createTrackedFetch(
+            async () =>
+                new Response(binaryBytes, {
+                    status: 200,
+                    headers: { 'content-type': 'image/png' },
+                }),
+        )
+
+        const response = await trackedFetch('https://files.todoist.com/file.png')
+        const buffer = await response.arrayBuffer?.()
+
+        expect(buffer).toBeDefined()
+        if (buffer) {
+            expect(new Uint8Array(buffer)).toEqual(binaryBytes)
+        }
+    })
+
     it('isolates concurrent tool contexts', async () => {
         const capturedTools: string[] = []
         const trackedFetch = createTrackedFetch(


src/usage-tracking.ts
```diff
@@ -195,6 +195,7 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
         headers: Object.fromEntries(response.headers.entries()),
         text: () => response.text(),
         json: () => response.json(),
+        arrayBuffer: () => response.arrayBuffer(),
```
doistbot, line 198: [P2] This adapter is still a hand-maintained subset of the Fetch `Response` API, which is the abstraction drift that led to this regression in the first place. As long as `toCustomFetchResponse()` cherry-picks methods, the next SDK endpoint that uses another standard member (`blob()`, `formData()`, `bodyUsed`, etc.) will need another fix here. Please preserve the native `Response` surface at this boundary (or proxy it generically) instead of extending the wrapper one method at a time.

doistbot, line 198: [P2] The forwarding of `arrayBuffer` is the core fix for preventing binary payload corruption, but it is entirely untested. Please add a unit test to `usage-tracking.test.ts` to verify that `createTrackedFetch` correctly exposes `arrayBuffer()` and successfully returns binary data without loss.

doistbot (commented):
This PR bumps the Todoist SDK and updates the tracked-fetch wrapper to forward `arrayBuffer()`, resolving a critical bug where binary attachments were corrupted by UTF-8 text encoding. Restoring binary-safe responses is an excellent fix that ensures reliable image processing for the hosted MCP deployment. However, relying on a hand-maintained subset of the Fetch API leaves the adapter vulnerable to future abstraction drift, and adding unit tests for the new array buffer functionality would help ensure long-term stability.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-ai&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-ai%2Fpull%2F477&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5c865540-4f0e-11f1-84ce-fb19b27b2e7b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5c865540-4f0e-11f1-84ce-fb19b27b2e7b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/usage-tracking.ts
scottlovegrove: Pushing back on this one — keeping the enumerated wrapper as-is.

Three alternatives all have meaningful drawbacks:

1. **Proxy + cast through `unknown`** to `Omit<Response, "headers"> & { headers: Record<string, string> }`. The cast is structurally necessary (we swap a typed property at runtime; TS can't follow it) and `as unknown` reintroduces the exact escape hatch we want to avoid.
2. **Return the native `Response` directly.** Requires the SDK 's `CustomFetchResponse.headers` to accept `Headers | Record<string, string>` — a coordinated SDK change for an abstraction win, not a bug fix.
3. **Object.create / property override.** Native `Response` body-reading methods check `this instanceof Response`, so any wrapper that isn't a Proxy with bound methods fails at runtime.

The drift hazard is real, but the cost of mitigating it is real too. The explicit literal keeps full TS safety: any new required field on `CustomFetchResponse` fails to compile here, and adding an optional field is a single-line follow-up — which is what just happened with `arrayBuffer`. Happy to revisit if `CustomFetchResponse` grows substantially.

scottlovegrove: Added in 6e2270d. The new test feeds binary bytes (PNG magic + a run of high bytes that would be replaced with U+FFFD by any UTF-8 text round-trip) through `createTrackedFetch`, calls `arrayBuffer()` on the resulting response, and asserts byte-for-byte equality.

pawelgrimm (commented):
🚀

PR #484, opened 2026-05-18
feat!: reorder MCP tool response content blocks (JSON first, summary last)

scottlovegrove:
## Summary

- Swap legacy `content[]` ordering inside `getToolOutput()` so the stringified JSON of `structuredContent` lives at `content[0]` and the human-readable prose summary moves to the last block. Tools that return additional content items (e.g. `view-attachment` returning an image) keep those tool-authored blocks first, with the JSON `text` item following them.
- Unblocks MCP clients that only surface `content[0]` — notably **OpenAI's Responses API**, which strips `structuredContent` and `_meta` entirely before exposing the result to callers. Today those clients receive only the prose summary; after this change they get the structured payload they actually need.
- No new public API, no feature flag, no query parameter. Single-file behavior change to the wire shape of every tool result.

## Why a default change beats a flag

- The [MCP spec (2025-11-25)](https://modelcontextprotocol.io/specification/2025-11-25/server/tools) is silent on `content[]` ordering. The only normative line is that a server SHOULD also return serialized JSON in a TextContent block when returning `structuredContent` — which we still do.
- Modern clients (Claude Code v2.0.21+, Codex, Cursor) prefer `structuredContent` and largely ignore `content[]` when both are present — unaffected by this change.
- Our own MCP client (`automations`) reads `structuredContent` first and falls back to an order-agnostic scan of `content[]` — verified safe.
- Sibling MCP servers (`memory`, `github`) already lead with JSON; the leading prose summary was a Todoist-only choice.

## Risk

Low. The only consumers that would notice are clients that (a) read only `content[0]`, (b) feed it to an LLM, (c) preferred prose over JSON. LLMs parse JSON cleanly. Any consumer piping `content[0]` straight to a UI without an LLM in the loop would see raw JSON instead of prose — this pattern is rare and not observed in known integrations.

## Refs

- Background discussion: [SEP-1624](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624)

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm test` — all 863 tests pass (includes ordering assertions: legacy mode → JSON at `content[0]`, contentItems-first with image → `[image, JSON, summary]`, structured-only mode → no JSON dup block)
- [x] `npm run build` clean
- [x] `npm run format:check` clean
- [ ] Manual end-to-end against `https://ai.todoist.net/mcp` once `@doist/todoist-mcp` bump lands in `todoist-ai-integrations`
- [ ] Smoke `~/automations` after the AI integrations bump (expected unaffected — `structuredContent`-first consumer)

## BREAKING CHANGE

The order of items inside a tool result's `content[]` array has changed. Previously `content[0]` was a human-readable prose summary and `content[1]` was the stringified JSON of `structuredContent`. From this release, when a tool returns structured content the array order is:

1. Any tool-authored non-text content items (images, embedded resources) — unchanged for tools that use them
2. Stringified JSON of `structuredContent`
3. Human-readable prose summary (always the last `text` item)

Consumers that read `content[]` entries by hard-coded index — for example, expecting `content[0]` to be a prose string suitable for display to a human — will need to either:

- prefer `structuredContent` (recommended for any client implementing MCP spec `2025-03-26` or later), or
- locate the prose summary by taking the last `text` content item rather than `content[0]`, or
- locate the JSON payload by parsing the first `text` content item whose body is valid JSON rather than relying on a fixed index.

The MCP spec is silent on `content[]` ordering, and modern clients (Claude Code v2.0.21+, Codex, Cursor) prefer `structuredContent` and are unaffected. 

### DIFF
--- a/src/mcp-helpers.test.ts
+++ b/src/mcp-helpers.test.ts
@@ -1,5 +1,46 @@
+import type { ContentBlock } from '@modelcontextprotocol/sdk/types.js'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import { registerTool, stripEmailsFromObject, stripEmailsFromText } from './mcp-helpers.js'
+
+type RegisterToolArgs = Parameters<typeof registerTool>[0]
+type ToolFixture = RegisterToolArgs['tool']
+
+/**
+ * Build a minimal `TodoistTool` fixture for `registerTool` tests. Returns a
+ * shared scaffold (parameters, annotations, server/client casts) so individual
+ * tests only have to specify the bits they actually exercise.
+ */
+function buildToolFixture(overrides: {
+    name?: string
+    description?: string
+    outputSchema?: Record<string, unknown>
+    execute: ToolFixture['execute']
+}): ToolFixture {
+    const { name = 'test-tool', description = 'Test tool', outputSchema, execute } = overrides
+
+    return {
+        name,
+        description,
+        parameters: {},
+        ...(outputSchema ? { outputSchema } : {}),
+        annotations: {
+            readOnlyHint: true,
+            destructiveHint: false,
+            idempotentHint: true,
+        },
+        execute,
+    } as unknown as ToolFixture
+}
+
+/**
+ * Capture the callback `registerTool` registers with the MCP server.
+ */
+function captureRegisterToolMock() {
+    const mock = vi.fn()
+    const server = { registerTool: mock } as unknown as RegisterToolArgs['server']
+    const client = {} as RegisterToolArgs['client']
+    return { mock, server, client }
+}
 
 describe('stripEmailsFromObject', () => {
     it.each([
@@ -54,74 +95,50 @@ describe('stripEmailsFromText', () => {
 
 describe('registerTool config', () => {
     it('omits outputSchema when the tool does not declare one', () => {
-        const registerToolMock = vi.fn()
+        const { mock, server, client } = captureRegisterToolMock()
 
         registerTool({
-            tool: {
+            tool: buildToolFixture({
                 name: 'no-schema-tool',
                 description: 'Tool without output schema',
-                parameters: {},
-                annotations: {
-                    readOnlyHint: true,
-                    destructiveHint: false,
-                    idempotentHint: true,
-                },
                 execute: async () => ({ textContent: 'ok' }),
-            },
-            server: {
-                registerTool: registerToolMock,
-            } as unknown as Parameters<typeof registerTool>[0]['server'],
-            client: {} as Parameters<typeof registerTool>[0]['client'],
+            }),
+            server,
+            client,
         })
 
-        expect(registerToolMock).toHaveBeenCalledTimes(1)
-        const config = registerToolMock.mock.calls[0]?.[1] as Record<string, unknown>
+        expect(mock).toHaveBeenCalledTimes(1)
+        const config = mock.mock.calls[0]?.[1] as Record<string, unknown>
         expect(Object.hasOwn(config, 'outputSchema')).toBe(false)
     })
 
     it('includes outputSchema when the tool declares one', () => {
-        const registerToolMock = vi.fn()
+        const { mock, server, client } = captureRegisterToolMock()
         const outputSchema = {}
 
         registerTool({
-            tool: {
+            tool: buildToolFixture({
                 name: 'schema-tool',
                 description: 'Tool with output schema',
-                parameters: {},
                 outputSchema,
-                annotations: {
-                    readOnlyHint: true,
-                    destructiveHint: false,
-                    idempotentHint: true,
-                },
                 execute: async () => ({ textContent: 'ok' }),
-            },
-            server: {
-                registerTool: registerToolMock,
-            } as unknown as Parameters<typeof registerTool>[0]['server'],
-            client: {} as Parameters<typeof registerTool>[0]['client'],
+            }),
+            server,
+            client,
         })
 
-        const config = registerToolMock.mock.calls[0]?.[1] as Record<string, unknown>
+        const config = mock.mock.calls[0]?.[1] as Record<string, unknown>
         expect(config.outputSchema).toBe(outputSchema)
     })
 })
 
 describe('registerTool error path', () => {
     it('applies centralized API formatting in MCP callback errors', async () => {
-        const registerToolMock = vi.fn()
+        const { mock, server, client } = captureRegisterToolMock()
 
         registerTool({
-            tool: {
-                name: 'test-tool',
-                description: 'Test tool',
-                parameters: {},
+            tool: buildToolFixture({
                 outputSchema: {},
-                annotations: {
-                    readOnlyHint: true,
-                    destructiveHint: false,
-                    idempotentHint: true,
-                },
                 execute: async () => {
                     throw {
                         httpStatusCode: 500,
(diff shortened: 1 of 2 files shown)


src/mcp-helpers.test.ts
```diff
@@ -155,3 +156,91 @@ describe('registerTool error path', () => {
...
+        structuredContent,
+    }: {
+        textContent?: string
+        structuredContent?: Record<string, unknown>
+    }) {
+        vi.resetModules()
+        const { registerTool: registerToolFresh } = await import('./mcp-helpers.js')
+        const registerToolMock = vi.fn()
+
+        registerToolFresh({
```
doistbot, line 195: [P3] This new test path rebuilds the same fake `registerTool` fixture shape already used in the other `registerTool` tests in this file (`parameters`, `outputSchema`, `annotations`, and the server/client casts). Please extract that scaffold into one shared local factory/helper so the test harness does not drift when the tool contract changes.

src/mcp-helpers.ts
```diff
@@ -92,33 +92,31 @@ function getToolOutput<StructuredContent extends Record<string, unknown>>({
 
     const contentArray: Array<Record<string, unknown>> = []
 
-    if (textContent) {
-        contentArray.push({ type: 'text' as const, text: textContent })
+    // Stringified JSON goes first so clients that only surface content[0]
+    // (e.g. OpenAI Responses API, which strips structuredContent) get the
+    // structured payload rather than the prose summary.
+    if (!USE_STRUCTURED_CONTENT && structuredContent) {
```
doistbot, line 104: [P2] This shared helper now prepends the fallback JSON ahead of every `contentItems` block, so any tool that returns both `structuredContent` and rich `contentItems` will have its tool-authored primary block displaced by an implementation-detail text block. Since `execute()` explicitly supports mixed `contentItems`, consider keeping the JSON fallback after non-text `contentItems` (or scoping this reordering to text-only responses) so `getToolOutput()` doesn’t redefine block priority for future tools.

src/mcp-helpers.test.ts
```diff
@@ -155,3 +156,91 @@ describe('registerTool error path', () => {
...
     })
 })
+
+describe('registerTool content ordering', () => {
+    afterEach(() => {
+        vi.unstubAllEnvs()
+        vi.resetModules()
+    })
+
+    async function invokeCallback({
```
doistbot: [P2] `invokeCallback()` only accepts `textContent` and `structuredContent`, so these new tests never cover the other behavior changed in this diff: `textContent` now comes after `contentItems`. Please thread `contentItems` through this helper and add a legacy-mode assertion for `JSON -> contentItems -> summary`; otherwise attachment-style tools can regress without this suite catching it.

src/mcp-helpers.test.ts
```diff
@@ -155,3 +156,91 @@ describe('registerTool error path', () => {
...
+        // read the first block (e.g. OpenAI Responses API).
+        expect(content[0]?.type).toBe('text')
+        expect(JSON.parse(content[0]?.text ?? '')).toEqual(structuredContent)
+
+        // Last block is the prose summary.
+        expect(content[content.length - 1]?.text).toBe('Tasks matching filter: 1.')
+    })
+
+    it('omits the JSON dup block when USE_STRUCTURED_CONTENT mode is on', async () => {
+        vi.stubEnv('USE_STRUCTURED_CONTENT', 'true')
```
doistbot, line 277: [P2] This test still passes even if `process.env.USE_STRUCTURED_CONTENT === 'true'` stops being honored, because Vitest already runs with `NODE_ENV='test'` and `mcp-helpers.ts` treats that as structured-content mode. Set `NODE_ENV='production'` here as well so the assertion actually covers the env-flag branch instead of the test-runner default.

doistbot (commented):
This PR adjusts the MCP tool response ordering to place the stringified JSON payload first, ensuring compatibility with clients like the OpenAI Responses API that only surface the initial content block. This is a thoughtful and impactful update that seamlessly aligns with modern tooling behaviors without requiring new configuration. A few refinements are noted for the test suite, specifically around extracting a shared test fixture, properly isolating environment variables for accurate flag testing, and threading `contentItems` through the helpers to verify that the new fallback placement doesn't unintentionally displace attachment-style tool responses.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F484&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3da89330-52a7-11f1-94eb-c1b7e5997743%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3da89330-52a7-11f1-94eb-c1b7e5997743&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (commented):
_(withdrawn — see inline replies on each comment)_

src/mcp-helpers.test.ts
scottlovegrove: Done in e9e223e — extracted `buildToolFixture()` and `captureRegisterToolMock()` at the top of the file, and rewired the `registerTool config` and `registerTool error path` tests onto them too so the scaffold is in exactly one place.

src/mcp-helpers.ts
scottlovegrove: Good catch — fixed in e9e223e. Reordered to `[contentItems, JSON-dup, textContent]` so tool-authored content blocks keep primacy at `content[0]` (today only `view-attachment` populates `contentItems` and it has no `outputSchema`, but the future case where a tool returns both is now safe). The Responses API still receives the JSON at `content[0]` for the universal case (no `contentItems`).

src/mcp-helpers.test.ts
scottlovegrove: Threaded `contentItems` through `invokeCallback()` in e9e223e and added a new case asserting the `image → JSON → summary` ordering for legacy content mode. Catches both the new reordering rule and the `view-attachment` shape.

src/mcp-helpers.test.ts
scottlovegrove: Right — fixed in e9e223e. The `USE_STRUCTURED_CONTENT=true` test now also stubs `NODE_ENV='production'` so the assertion exercises the env-flag branch directly instead of riding on vitest defaulting `NODE_ENV` to `test`.

craigcarlyle:
✅

PR #491, opened 2026-05-31
fix: resolve project from section when assigning tasks with sectionId

scottlovegrove:
## What

`add-tasks` rejected any task that supplied a `sectionId` and a `responsibleUser` but no `projectId`, throwing:

> When assigning tasks to sections, please also specify projectId

## Why

Assignment validation needs the target project to confirm the assignee is a collaborator on it. The `parentId` branch resolves the project by looking up the parent task, but the `sectionId` branch never did the equivalent lookup, so it bailed out and forced the caller to pass a `projectId` that's fully determined by the section anyway. Callers that only knew the section (a common case) couldn't assign at all.

## Change

A section always belongs to a project, so resolve it via `getSection(sectionId)` and use its `projectId` for assignment validation, mirroring the existing `parentId` handling. A missing section now produces a clear `Section "<id>" not found` error instead of the misleading "specify projectId" message.

## Testing

- New tests: project is resolved from the section (validator receives the section's project, task created with `assigneeId`), and a bad `sectionId` surfaces a clear not-found error.
- Full suite green (906 tests), type-check and lint clean.

### DIFF
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -252,11 +252,13 @@ async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi)
                 throw new Error(`Task "${task.content}": Parent task "${parentId}" not found`)
             }
         } else if (!targetProjectId && sectionId) {
-            // For section tasks, we need to find the project - this is a limitation
-            // For now, we'll require explicit projectId when using assignments with sections
-            throw new Error(
-                `Task "${task.content}": When assigning tasks to sections, please also specify projectId`,
-            )
+            // For section tasks, resolve the project from the section
+            try {
+                const section = await client.getSection(sectionId)
+                targetProjectId = section.projectId
+            } catch (_error) {
+                throw new Error(`Task "${task.content}": Section "${sectionId}" not found`)
+            }
         }
 
         if (!targetProjectId) {
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -97,6 +97,9 @@ const addTasks = {
             }
         })
 
+        // Resolve each section's project at most once across the whole batch
+        const sectionProjectCache = new Map<string, string>()
+
         // Process groups in parallel; within each group, process sequentially
         type IndexedResult = { index: number; result: PromiseSettledResult<Task> }
         const groupResults = await Promise.all(
@@ -104,7 +107,7 @@ const addTasks = {
                 const results: IndexedResult[] = []
                 for (const { task, index } of group) {
                     try {
-                        const created = await processTask(task, client)
+                        const created = await processTask(task, client, sectionProjectCache)
                         results.push({ index, result: { status: 'fulfilled', value: created } })
                     } catch (error) {
                         results.push({
@@ -173,7 +176,11 @@ function destinationKey(task: z.infer<typeof TaskSchema>): string {
     return `${task.projectId ?? ''}|${task.sectionId ?? ''}|${task.parentId ?? ''}`
 }
 
-async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi): Promise<Task> {
+async function processTask(
+    task: z.infer<typeof TaskSchema>,
+    client: TodoistApi,
+    sectionProjectCache: Map<string, string>,
+): Promise<Task> {
     const {
         duration: durationStr,
         projectId,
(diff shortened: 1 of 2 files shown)


src/tools/add-tasks.ts
```diff
@@ -252,11 +252,13 @@ async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi)
...
-            // For section tasks, we need to find the project - this is a limitation
-            // For now, we'll require explicit projectId when using assignments with sections
-            throw new Error(
-                `Task "${task.content}": When assigning tasks to sections, please also specify projectId`,
-            )
+            // For section tasks, resolve the project from the section
+            try {
+                const section = await client.getSection(sectionId)
+                targetProjectId = section.projectId
+            } catch (_error) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This catch now rewrites every `getSection()` failure as `Section "<id>" not found`. If Todoist returns a 403/503 or a transient network error while resolving the section, callers will get a bogus not-found message instead of the real failure. Please only map actual missing-section cases (for example, 404/undefined) and rethrow other errors.

src/tools/add-tasks.ts
```diff
@@ -252,11 +252,13 @@ async function processTask(task: z.infer<typeof TaskSchema>, client: TodoistApi)
...
             }
         } else if (!targetProjectId && sectionId) {
-            // For section tasks, we need to find the project - this is a limitation
-            // For now, we'll require explicit projectId when using assignments with sections
-            throw new Error(
-                `Task "${task.content}": When assigning tasks to sections, please also specify projectId`,
-            )
+            // For section tasks, resolve the project from the section
+            try {
+                const section = await client.getSection(sectionId)
```
doistbot, line 266: <big><kbd>🟡 P2</kbd></big>  This adds one `getSection(sectionId)` call per task. In `add-tasks`, tasks are already grouped by destination and processed sequentially within each group, so a batch of sibling tasks in the same section now repeats the same section fetch N times before assignment validation. Cache `sectionId -> projectId` for the batch/group, or resolve it once before looping, to avoid extra network round trips on multi-task creates.

doistbot (commented):
Thanks scottlovegrove for your contribution 😇. It is great to see this limitation removed so callers can assign tasks using only a section ID.

Few things worth tightening:
- Only map actual missing-section errors (like 404s) to the "not found" message so real API or network failures aren't swallowed.
- Cache the `sectionId` to `projectId` lookup to prevent duplicate network requests when batched tasks target the same section.

<details>

- **[P3] src/tools/__tests__/add-tasks.test.ts:704**: This spy modifies the implementation of `assignmentValidator.validateTaskCreationAssignment` but is never restored. Since `vi.clearAllMocks()` in `beforeEach` only clears call history, this mock implementation will leak to all subsequent tests in the file. While it doesn't currently break other tests, it introduces test pollution that can cause confusing behavior if new assignment tests are added. Consider calling `validateSpy.mockRestore()` at the end of the test.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F491&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A092b37a0-5cd6-11f1-9736-1b9a9e497f3a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A092b37a0-5cd6-11f1-9736-1b9a9e497f3a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/add-tasks.ts
scottlovegrove: Good catch. Switched to mirror the not-found handling in `fetch-object`: `getSection` returns a falsy value for a genuinely missing section, so only that case maps to "Section not found". Real API/network failures (403/503/timeouts) now propagate untouched. Added a test asserting a non-not-found error is rethrown verbatim.

src/tools/add-tasks.ts
scottlovegrove: Done. Added a per-batch `sectionId -> projectId` cache so each section is resolved at most once, even across groups. Sibling tasks in the same section now reuse the first lookup. Added a test asserting `getSection` is called once for two tasks targeting the same section.

PR #493, opened 2026-06-02
fix: allow hidden collaborator emails

roryist:
References https://github.com/Doist/Issues/issues/20418.


roryist:
(This reply is from Rory's agent)

We're closing this PR because we decided not to do this work ad hoc. It needs coordinated planning across clients and the API, including backward compatibility. See https://github.com/Doist/Issues/issues/20418 for the wider discussion.

PR #494, opened 2026-06-05
fix: Always serve the most recent app resource

henningmu:
## Short description

We inject all client-side dependencies (e.g., React) in a single HTML file that we serve as an app resource. That file gets it content-hash appended. Any dependency update produces a new hash, invalidating existing files. This is good as long as MCP clients can correctly request the newest version. However, if their caching/snapshotting policy is too aggressive, it can lead to unreachable resources being referenced.

This is what happened during our submission to ChatGPT. We got rejected because their submission process kept requesting stale resources. Now there are two ways to solve this:

1. Keep old app resources around "indefinitely" (e.g., a year) so they can get served correctly
2. Introduce a catch-all resource that serves the most recent app resource independent of the requested hash

While (1) is clearer, it requires some infrastructure changes to how we deploy our MCP server that I wasn't willing to embark on right now. (2) should work just as fine, and arguably, it's better if clients always get the most recent version.

## Test plan

- `resources/list` advertises the current exact app resource
- `resources/templates/list` advertises the fallback
- `resources/read` for the current URI returns a valid html / mcp app
- `resources/read` for stale URI (e.g., `ui://todoist/task-list@deadbeefcafe` also returns a valid html / mcp pp

## PR Checklist

### DIFF
--- a/src/mcp-apps/resources.test.ts
+++ b/src/mcp-apps/resources.test.ts
@@ -16,7 +16,7 @@ describe('registerTaskListApp', () => {
         registerTaskListApp(server)
 
         expect(consoleErrorSpy).not.toHaveBeenCalled()
-        expect(registerResourceSpy).toHaveBeenCalledTimes(1)
+        expect(registerResourceSpy).toHaveBeenCalledTimes(2)
         expect(registerResourceSpy.mock.calls[0]?.[2]).toMatchObject({
             description: 'Interactive task list widget',
             _meta: {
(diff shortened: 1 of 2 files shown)


src/mcp-apps/resources.test.ts
```diff
@@ -77,5 +77,47 @@ describe('registerTaskListApp', () => {
...
+        if (!fallbackReadCallback) {
+            throw new Error('fallback registerResource callback was not captured')
+        }
+
+        const staleUri = 'ui://todoist/task-list@stale1234567'
+        const fallbackResult = await fallbackReadCallback(new URL(staleUri))
+
+        expect(fallbackResult.contents).toHaveLength(1)
+        expect(fallbackResult.contents[0]?.uri).toBe(staleUri)
+        expect(fallbackResult.contents[0]?.text).toContain('<div id="root"></div>')
```
doistbot, line 121: <big><kbd>🟡 P2</kbd></big>  The stale-URI assertion only checks for `<div id="root"></div>`, so it won't catch regressions where the fallback serves a placeholder or older app shell instead of the current resource. Compare `fallbackResult.contents[0].text` with the current `result.contents[0].text` (or at least assert the module script tag too) so this test proves stale hashes get the latest app HTML.

doistbot (commented):
Thanks Henning for smoothly resolving the stale hash caching issue by introducing a fallback resource template 😄 🙌.

Few things worth tightening:
- The stale-URI test currently only checks for `<div id="root"></div>`, which wouldn't catch regressions if an older app shell is served; consider comparing the fallback result directly with the current result to prove the latest HTML is returned.

<details>

- **[P3] src/mcp-apps/resources.test.ts:81**: This is brittle because it assumes the fallback registration is always `mock.calls[1]`. If another resource gets registered before it, the test will fail even though the behavior is still correct. Find the call by its resource name (`todoist-task-list-hash-fallback`) instead of hard-coding the index.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F494&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aca499700-60ba-11f1-8e32-468e8ade3423%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aca499700-60ba-11f1-8e32-468e8ade3423&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/mcp-apps/resources.test.ts
henningmu: I'm fine with that possibility 🤨

PR #502, opened 2026-06-11
fix(update-tasks): Report per-task failures instead of failing the whole batch

engfragui:
# Overview

`update-tasks` previously used `Promise.all`, so one rejected item discarded every successful result and surfaced an opaque tool failure. This refresh settles items independently and returns structured per-task `failures`, including when every item fails.

The batch is capped at 25 items and transient 502/503/504 responses are retried per item. API failure details retain the specific Todoist error rather than only the SDK’s generic HTTP status.

A move followed by a failed field update is reported as a partial outcome: the moved task remains in `tasks`/`updatedTaskIds`, while `failures` says that the move was applied and the field update failed. This prevents callers from treating an already-applied move as an untouched item.

## Reference

- Helps with https://github.com/Doist/Issues/issues/20441
- Aligns with https://github.com/Doist/todoist-mcp/issues/505

## Verification

- `npm test`
- `npm run type-check`
- `npm run format:check`
- `npm run build`

### DIFF
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -1206,10 +1206,10 @@ describe(`${UPDATE_TASKS} tool`, () => {
                 failureCount: 1,
             })
 
-            // The text content tells the model not to retry the whole batch.
+            // The text content surfaces the per-task failure alongside the success.
             expect(result.textContent).toContain('Updated 1 task')
             expect(result.textContent).toContain('Failed (1)')
-            expect(result.textContent).toContain('not retried automatically')
+            expect(result.textContent).toContain('address or drop these items')
         })
 
         it('throws when every task in the batch fails', async () => {
@@ -1346,6 +1346,48 @@ describe(`${UPDATE_TASKS} tool`, () => {
                 skippedCount: 0,
                 failureCount: 1,
             })
+        })
+
+        it('truncates the failure list to 3 and shows "+N more"', async () => {
+            // A non-throwing batch (one success keeps it from being a total failure) with
+            // more than MAX_FAILURES_SHOWN (3) failures must cap the displayed list and
+            // append "+N more" so the truncation isn't silently dropped by a refactor.
+            mockTodoistApi.updateTask.mockImplementation((id: string) => {
+                if (id === 'ok-task') {
+                    return Promise.resolve(createMockTask({ id: 'ok-task', content: 'ok' }))
+                }
+                return Promise.reject(new Error('API Error: boom'))
+            })
+
+            const result = await updateTasks.execute(
+                {
+                    tasks: [
+                        { id: 'ok-task', content: 'ok' },
+                        { id: 'bad-1', content: 'x' },
+                        { id: 'bad-2', content: 'x' },
+                        { id: 'bad-3', content: 'x' },
+                        { id: 'bad-4', content: 'x' },
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            const { structuredContent, textContent } = result
+            // All 4 failures are retained in the structured output...
+            expect(structuredContent.failures).toHaveLength(4)
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 1,
+                skippedCount: 0,
+                failureCount: 4,
+            })
+
+            // ...but the text summary shows only the first 3 and notes the remainder.
+            expect(textContent).toContain('Failed (4)')
+            expect(textContent).toContain('bad-1')
+            expect(textContent).toContain('bad-2')
+            expect(textContent).toContain('bad-3')
+            expect(textContent).not.toContain('bad-4')
+            expect(textContent).toContain('+1 more')
         })
 
         it('does not throw when the only outcome is a partial success', async () => {
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -1161,6 +1161,105 @@ describe(`${UPDATE_TASKS} tool`, () => {
         })
     })
 
+    describe('partial batch failures', () => {
+        it('keeps successful updates when one task in the batch fails', async () => {
+            const okTask = createMockTask({ id: 'ok-task', content: 'Updated ok' })
+            mockTodoistApi.updateTask.mockResolvedValue(okTask)
+            // The forbidden cross-workspace move shape the API rejects with 403.
+            mockTodoistApi.moveTask.mockRejectedValue(
+                Object.assign(new Error('Request failed with status code 403'), {
+                    httpStatusCode: 403,
+                    responseData: {
+                        error: 'Not allowed to move objects out of a workspace',
+                        error_tag: 'FORBIDDEN',
+                        http_code: 403,
+                    },
+                }),
+            )
+
+            const result = await updateTasks.execute(
+                {
+                    tasks: [
+                        { id: 'ok-task', content: 'Updated ok' },
+                        { id: 'bad-task', projectId: 'personal-project' },
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            const { structuredContent } = result
+            // The valid update is preserved instead of being discarded by the failure.
+            expect(structuredContent.tasks).toHaveLength(1)
+            expect(structuredContent.totalCount).toBe(1)
+            expect(structuredContent.updatedTaskIds).toEqual(['ok-task'])
+
+            // The failure is reported per-task, with the specific API objection surfaced.
+            expect(structuredContent.failures).toHaveLength(1)
+            expect(structuredContent.failures[0]?.item).toBe('bad-task')
+            expect(structuredContent.failures[0]?.error).toContain(
+                'Not allowed to move objects out of a workspace',
+            )
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 1,
+                skippedCount: 0,
+                failureCount: 1,
+            })
+
+            // The text content tells the model not to retry the whole batch.
+            expect(result.textContent).toContain('Updated 1 task')
+            expect(result.textContent).toContain('Failed (1)')
+            expect(result.textContent).toContain('not retried automatically')
+        })
+
+        it('throws when every task in the batch fails', async () => {
+            mockTodoistApi.moveTask.mockRejectedValue(
+                Object.assign(new Error('Request failed with status code 403'), {
+                    httpStatusCode: 403,
+                    responseData: {
+                        error: 'Not allowed to move objects out of a workspace',
+                        http_code: 403,
+                    },
+                }),
+            )
+
+            await expect(
+                updateTasks.execute(
+                    {
+                        tasks: [
+                            { id: 'bad-1', projectId: 'personal-project' },
+                            { id: 'bad-2', projectId: 'personal-project' },
+                        ],
+                    },
+                    mockTodoistApi,
+                ),
+            ).rejects.toThrow('All 2 task update(s) failed')
+        })
+
+        it('counts skipped (no-change) tasks separately from failures', async () => {
+            const okTask = createMockTask({ id: 'ok-task', content: 'Updated ok' })
+            mockTodoistApi.updateTask.mockResolvedValue(okTask)
+
+            const result = await updateTasks.execute(
+                {
+                    tasks: [
+                        { id: 'ok-task', content: 'Updated ok' },
+                        { id: 'noop-task' }, // only id -> skipped, not a failure
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            const { structuredContent } = result
+            expect(structuredContent.tasks).toHaveLength(1)
+            expect(structuredContent.failures).toHaveLength(0)
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 1,
+                skippedCount: 1,
+                failureCount: 0,
+            })
+        })
+    })
+
     describe('isUncompletable parameter', () => {
         it('should pass isUncompletable parameter to SDK', async () => {
             // Mock API response - minimal mock just to prevent errors
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -1266,17 +1266,13 @@ describe(`${UPDATE_TASKS} tool`, () => {
             })
         })
 
-        it('reports a partial success when the move applies but the field update fails', async () => {
-            // Combined move + field update: the move succeeds, then updateTask rejects.
-            // The move is already applied server-side, so the task must NOT be dropped as a
-            // pure failure — it appears in tasks (moved) and the failure names what still
-            // needs retrying (the field change only, not the move).
-            const movedTask = createMockTask({
-                id: 'move-update-task',
-                content: 'Original content',
-                projectId: 'new-project-id',
-            })
-            mockTodoistApi.moveTask.mockResolvedValue(movedTask)
+        it('reports the whole task as a failure when the move succeeds but the field update fails', async () => {
+            // Combined move + field update where the move succeeds but updateTask rejects.
+            // The task is reported as a single failure — we do not surface which part
+            // applied.
+            mockTodoistApi.moveTask.mockResolvedValue(
+                createMockTask({ id: 'move-update-task', projectId: 'new-project-id' }),
+            )
             mockTodoistApi.updateTask.mockRejectedValue(new Error('API Error: Invalid priority'))
 
             const result = await updateTasks.execute(
@@ -1292,27 +1288,16 @@ describe(`${UPDATE_TASKS} tool`, () => {
                 mockTodoistApi,
             )
 
-            expect(mockTodoistApi.moveTask).toHaveBeenCalledWith('move-update-task', {
-                projectId: 'new-project-id',
-            })
-
             const { structuredContent } = result
-            // The completed move is surfaced, not discarded.
-            expect(structuredContent.tasks).toHaveLength(1)
-            expect(structuredContent.tasks[0]?.id).toBe('move-update-task')
-            expect(structuredContent.updatedTaskIds).toEqual(['move-update-task'])
-
-            // The field-update failure is reported as a partial: the move is surfaced as
-            // already applied, without advising a retry.
+            // The task is not reported as updated...
+            expect(structuredContent.tasks).toHaveLength(0)
+            expect(structuredContent.updatedTaskIds).toEqual([])
+            // ...it is a single failure carrying the field-update error.
             expect(structuredContent.failures).toHaveLength(1)
             expect(structuredContent.failures[0]?.item).toBe('move-update-task')
-            expect(structuredContent.failures[0]?.error).toContain('Move applied')
-            expect(structuredContent.failures[0]?.error).toContain('The move is already done')
             expect(structuredContent.failures[0]?.error).toContain('API Error: Invalid priority')
-
-            // Counts reflect both facts: the move applied AND a field update failed.
             expect(structuredContent.appliedOperations).toEqual({
-                updateCount: 1,
+                updateCount: 0,
                 skippedCount: 0,
                 failureCount: 1,
             })
@@ -1358,23 +1343,6 @@ describe(`${UPDATE_TASKS} tool`, () => {
             expect(textContent).toContain('bad-3')
             expect(textContent).not.toContain('bad-4')
             expect(textContent).toContain('+1 more')
-        })
-
-        it('does not throw when the only outcome is a partial success', async () => {
-            // A partial means the move applied, so it is not a total failure even though a
-            // field update failed — the batch returns normally.
-            const movedTask = createMockTask({ id: 'only-partial', projectId: 'new-project-id' })
-            mockTodoistApi.moveTask.mockResolvedValue(movedTask)
-            mockTodoistApi.updateTask.mockRejectedValue(new Error('boom'))
-
-            await expect(
-                updateTasks.execute(
-                    {
-                        tasks: [{ id: 'only-partial', projectId: 'new-project-id', content: 'x' }],
-                    },
-                    mockTodoistApi,
-                ),
-            ).resolves.toBeDefined()
         })
     })
 
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -1258,6 +1258,42 @@ describe(`${UPDATE_TASKS} tool`, () => {
                 failureCount: 0,
             })
         })
+
+        it('does not throw when the batch is only skipped and failed tasks', async () => {
+            // No task is actually updated: one is a no-op skip, the other fails. A skip is
+            // a successful no-op, so this is NOT a total failure and must return normally
+            // with the failure listed — rather than throwing a batch-wide error.
+            mockTodoistApi.moveTask.mockRejectedValue(
+                Object.assign(new Error('Request failed with status code 403'), {
+                    httpStatusCode: 403,
+                    responseData: {
+                        error: 'Not allowed to move objects out of a workspace',
+                        http_code: 403,
+                    },
+                }),
+            )
+
+            const result = await updateTasks.execute(
+                {
+                    tasks: [
+                        { id: 'noop-task' }, // only id -> skipped
+                        { id: 'bad-task', projectId: 'personal-project' }, // move -> fails
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            const { structuredContent } = result
+            expect(structuredContent.tasks).toHaveLength(0)
+            expect(structuredContent.totalCount).toBe(0)
+            expect(structuredContent.failures).toHaveLength(1)
+            expect(structuredContent.failures[0]?.item).toBe('bad-task')
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 0,
+                skippedCount: 1,
+                failureCount: 1,
+            })
+        })
     })
 
     describe('isUncompletable parameter', () => {
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -1295,6 +1295,75 @@ describe(`${UPDATE_TASKS} tool`, () => {
                 failureCount: 1,
             })
         })
+
+        it('reports a partial success when the move applies but the field update fails', async () => {
+            // Combined move + field update: the move succeeds, then updateTask rejects.
+            // The move is already applied server-side, so the task must NOT be dropped as a
+            // pure failure — it appears in tasks (moved) and the failure names what still
+            // needs retrying (the field change only, not the move).
+            const movedTask = createMockTask({
+                id: 'move-update-task',
+                content: 'Original content',
+                projectId: 'new-project-id',
+            })
+            mockTodoistApi.moveTask.mockResolvedValue(movedTask)
+            mockTodoistApi.updateTask.mockRejectedValue(new Error('API Error: Invalid priority'))
+
+            const result = await updateTasks.execute(
+                {
+                    tasks: [
+                        {
+                            id: 'move-update-task',
+                            projectId: 'new-project-id',
+                            content: 'New content',
+                        },
+                    ],
+                },
+                mockTodoistApi,
+            )
+
+            expect(mockTodoistApi.moveTask).toHaveBeenCalledWith('move-update-task', {
+                projectId: 'new-project-id',
+            })
+
+            const { structuredContent } = result
+            // The completed move is surfaced, not discarded.
+            expect(structuredContent.tasks).toHaveLength(1)
+            expect(structuredContent.tasks[0]?.id).toBe('move-update-task')
+            expect(structuredContent.updatedTaskIds).toEqual(['move-update-task'])
+
+            // The field-update failure is reported as a partial: the move is surfaced as
+            // already applied, without advising a retry.
+            expect(structuredContent.failures).toHaveLength(1)
+            expect(structuredContent.failures[0]?.item).toBe('move-update-task')
+            expect(structuredContent.failures[0]?.error).toContain('Move applied')
+            expect(structuredContent.failures[0]?.error).toContain('The move is already done')
+            expect(structuredContent.failures[0]?.error).toContain('API Error: Invalid priority')
+
+            // Counts reflect both facts: the move applied AND a field update failed.
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 1,
+                skippedCount: 0,
+                failureCount: 1,
+            })
+        })
+
+        it('does not throw when the only outcome is a partial success', async () => {
+            // A partial means the move applied, so it is not a total failure even though a
+            // field update failed — the batch returns normally.
+            const movedTask = createMockTask({ id: 'only-partial', projectId: 'new-project-id' })
+            mockTodoistApi.moveTask.mockResolvedValue(movedTask)
+            mockTodoistApi.updateTask.mockRejectedValue(new Error('boom'))
+
+            await expect(
+                updateTasks.execute(
+                    {
+                        tasks: [{ id: 'only-partial', projectId: 'new-project-id', content: 'x' }],
+                    },
+                    mockTodoistApi,
+                ),
+            ).resolves.toBeDefined()
+        })
     })
 
     describe('isUncompletable parameter', () => {
--- a/src/tools/update-tasks.test.ts
+++ b/src/tools/update-tasks.test.ts
@@ -742,75 +742,51 @@ describe(`${UPDATE_TASKS} tool`, () => {
     })
 
     describe('error handling', () => {
-        it('should throw error for invalid duration format', async () => {
-            await expect(
-                updateTasks.execute(
-                    {
-                        tasks: [
-                            {
-                                id: '8485093756',
-                                duration: 'invalid',
-                            },
-                        ],
-                    },
-                    mockTodoistApi,
-                ),
-            ).rejects.toThrow('Task 8485093756: Invalid duration format "invalid"')
+        // The tool never throws for per-item problems — even a single failing task is
+        // reported in the structured `failures` rather than rejecting the whole call.
+        async function expectSingleFailure(
+            params: Parameters<typeof updateTasks.execute>[0]['tasks'][number],
+            expectedError: string,
+        ) {
+            const result = await updateTasks.execute({ tasks: [params] }, mockTodoistApi)
+            const { structuredContent } = result
+            expect(structuredContent.tasks).toHaveLength(0)
+            expect(structuredContent.failures).toHaveLength(1)
+            expect(structuredContent.failures[0]?.item).toBe(params.id)
+            expect(structuredContent.failures[0]?.error).toContain(expectedError)
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 0,
+                skippedCount: 0,
+                failureCount: 1,
+            })
+            return result
+        }
+
+        it('reports invalid duration format as a failure', async () => {
+            await expectSingleFailure(
+                { id: '8485093756', duration: 'invalid' },
+                'Task 8485093756: Invalid duration format "invalid"',
+            )
         })
 
-        it('should throw error for duration exceeding 24 hours', async () => {
-            await expect(
-                updateTasks.execute(
-                    {
-                        tasks: [
-                            {
-                                id: '8485093757',
-                                duration: '25h',
-                            },
-                        ],
-                    },
-                    mockTodoistApi,
-                ),
-            ).rejects.toThrow(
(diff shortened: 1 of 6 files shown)


src/tools/update-tasks.ts
```diff
@@ -113,127 +121,50 @@ const updateTasks = {
...
+                error: formatToolExecutionError(result.reason),
+            })
         })
-        const updatedTasks = (await Promise.all(updateTasksPromises)).filter(
-            (task): task is Task => task !== undefined,
-        )
+
+        // If every attempted task failed (nothing updated, nothing skipped), surface a
+        // hard error so the caller sees a failure rather than a misleading success.
+        if (updatedTasks.length === 0 && failures.length > 0) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This throws even when the batch had skipped no-op items. For example, `[{ id: 'noop' }, { id: 'bad', projectId: 'x' }]` ends up with `updatedTasks.length === 0` and `failures.length === 1`, so the tool raises `All 1 task update(s) failed` instead of returning the partial result with `skippedCount: 1` and the failure entry. Include `skippedCount === 0` (or otherwise exclude skipped items) in this condition so only true total failures throw.

src/tools/update-tasks.test.ts
```diff
@@ -1161,6 +1161,105 @@ describe(`${UPDATE_TASKS} tool`, () => {
...
+            const { structuredContent } = result
+            expect(structuredContent.tasks).toHaveLength(1)
+            expect(structuredContent.failures).toHaveLength(0)
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 1,
+                skippedCount: 1,
+                failureCount: 0,
+            })
+        })
+    })
```
doistbot, line 1398: <big><kbd>🟠 P1</kbd></big>  Missing coverage for a batch containing only skipped and failed tasks. Adding this test reveals a bug in the production code: because skipped tasks aren't added to `updatedTasks`, the check `if (updatedTasks.length === 0 && failures.length > 0)` currently throws a batch-wide error even if some tasks were successfully skipped. Asserting that this scenario does not throw will ensure the implementation accounts for `skippedCount`, properly matching the intent of the `(nothing updated, nothing skipped)` comment.

doistbot (commented):
This transitions `update-tasks` to process modifications independently, ensuring successful updates are preserved even if partial failures occur.

Few things worth tightening:
- Update the total failure condition to account for skipped items (e.g., `skippedCount === 0`) so batches composed entirely of skipped and failed tasks don't mistakenly throw a whole-batch error.
- Add a test verifying a batch containing only skipped and failed tasks to ensure this edge case is properly caught and handled.

<details>

- **[P3] src/tools/update-tasks.ts:151**: `formatToolExecutionError()` is the outer MCP error renderer, not a batch-result serializer. It returns multi-line prose with guidance (`Todoist API request failed...`, `Try next:`), so `failures[].error` becomes presentation text instead of a stable per-item message, and `generateTextContent()` will splice those extra lines into the failure list without indentation. Prefer storing a one-line error (and optional code) here, and keep `formatToolExecutionError()` for the top-level throw path.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9291f310-657a-11f1-841d-9589a5c67205%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9291f310-657a-11f1-841d-9589a5c67205&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
@doistbot /review

src/tools/update-tasks.ts
```diff
@@ -113,127 +121,52 @@ const updateTasks = {
...
-            const movedTask = await client.moveTask(id, moveArgs)
-
-            if (Object.keys(updateArgs).length > 0) {
-                return await client.updateTask(id, updateArgs)
-            }
-
-            return movedTask
+            failures.push({
+                item: tasks[index]?.id ?? `Task ${index + 1}`,
+                error: formatToolExecutionError(result.reason),
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `formatToolExecutionError()` returns multi-line presentation prose intended for top-level errors. Using it for per-item failures breaks formatting in two places: when joined with `'; '` on line 160 during a total failure, and when formatted in `generateTextContent` on line 331 (the embedded newlines will break the list indentation). Prefer capturing a single-line string here, such as `result.reason?.message ?? String(result.reason)`.

doistbot (commented):
This successfully refactors `update-tasks` to process modifications independently, properly handling skipped tasks and preserving partial successes.

Few things worth tightening:
- Using the top-level `formatToolExecutionError()` for per-item failures introduces multi-line strings that break downstream formatting; consider grabbing a single-line string like `result.reason?.message ?? String(result.reason)` instead.

<details>

- **[P3] src/tools/update-tasks.ts:160**: This total-failure path concatenates every per-item error into one big string, but `registerTool()` will pass the thrown error through `formatToolExecutionError()` again, which sanitizes/truncates it before returning it to the caller. On large batches, that means a lot of avoidable allocation and string work for text nobody sees. Limit `details` to a small prefix (for example `DisplayLimits.MAX_FAILURES_SHOWN`) or throw a compact summary/counts-only error here.
- **[P3] src/tools/__tests__/update-tasks.test.ts:1164**: These new cases don't exercise the new failure-list truncation path. `update-tasks` now slices failures to `DisplayLimits.MAX_FAILURES_SHOWN` and appends `+N more`, but there's no test here for `>3` failures with a non-throwing result (for example, 1 skipped/successful item plus 4+ failures). Add one so a refactor can't silently remove the cap and start emitting oversized MCP responses.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A46a9bdd0-657e-11f1-9ea2-f5ac8b92fe16%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A46a9bdd0-657e-11f1-9ea2-f5ac8b92fe16&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
@doistbot /review

src/tools/update-tasks.ts
```diff
@@ -242,35 +175,163 @@ const updateTasks = {
...
+    // If no move parameters are provided, use updateTask without moveTask
+    if (!resolvedProjectId && !sectionId && !parentId) {
+        return await client.updateTask(id, updateArgs)
+    }
+
+    const moveArgs = createMoveTaskArgs(id, resolvedProjectId, sectionId, parentId)
+    const movedTask = await client.moveTask(id, moveArgs)
+
+    if (Object.keys(updateArgs).length > 0) {
+        return await client.updateTask(id, updateArgs)
```
doistbot: <big><kbd>🟡 P2</kbd></big>  A task that combines a move with other field updates can still be partially applied and then reported as a full failure. If `moveTask()` on line 293 succeeds but this `updateTask()` call rejects (for example, API validation on the non-move fields or a transient 5xx), the task has already been moved, yet the batch result drops it from `updatedTasks` and tells callers to retry the failed item. That reintroduces “successful updates are discarded” for mixed move+update items. Catch the second-call failure and surface the moved task as partially applied, or otherwise avoid treating the whole item as a pure failure.

src/tools/update-tasks.ts
```diff
@@ -113,127 +120,53 @@ const updateTasks = {
...
-            const movedTask = await client.moveTask(id, moveArgs)
-
-            if (Object.keys(updateArgs).length > 0) {
-                return await client.updateTask(id, updateArgs)
-            }
-
-            return movedTask
+            failures.push({
+                item: tasks[index]?.id ?? `Task ${index + 1}`,
+                error:
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This new per-item failure API only serializes `reason.message`. That leaves callers with free-form transport text like `"Request failed with status code 403"` when they need to decide which items are retryable. `FailureSchema` already has an optional `code`, and these SDK errors carry stable fields such as `httpStatusCode` / `responseData.error_tag`; populate `code` here (and preferably normalize `error` too) so consumers don't have to parse message strings.

src/tools/update-tasks.test.ts
```diff
@@ -1161,6 +1161,142 @@ describe(`${UPDATE_TASKS} tool`, () => {
...
+            expect(structuredContent.totalCount).toBe(0)
+            expect(structuredContent.failures).toHaveLength(1)
+            expect(structuredContent.failures[0]?.item).toBe('bad-task')
+            expect(structuredContent.appliedOperations).toEqual({
+                updateCount: 0,
+                skippedCount: 1,
+                failureCount: 1,
+            })
+        })
+    })
```
doistbot, line 1398: <big><kbd>🟡 P2</kbd></big>  There is no test exercising the failure-list truncation path (`failures.slice(0, DisplayLimits.MAX_FAILURES_SHOWN)`). Add a test for a non-throwing batch with more than 3 failures (e.g. 1 success/skipped plus 4 failures) to ensure the `+N more` formatting works correctly and a future refactor doesn't silently remove the cap.

doistbot (commented):
This is a great refactor that isolates task updates via `Promise.allSettled`, preventing a single rejection from failing an entire batch.

Few things worth tightening:
- **Move + update edge cases**: A task that successfully moves but fails its subsequent field updates is still reported as a full failure and dropped from `updatedTasks`, risking unsafe retries.
- **Failure codes**: Populate the failure `code` field with the underlying HTTP status or API error tag so callers aren't forced to parse raw transport error messages.
- **Truncation testing**: Add test coverage for batches with more than three failures to ensure the display limit truncation behaves as expected.

<details>

- **[P3] src/tools/update-tasks.ts:329**: This reimplements batch failure formatting inside the tool instead of going through `response-builders.ts`, which the repo uses to keep tool output consistent. We now have two copies of the `MAX_FAILURES_SHOWN` / `+N more` logic (`summarizeBatch` and this block), so the next wording or limit change has to be made in multiple places. Consider extracting a shared helper or extending the existing response builder rather than formatting failures here.
- **[P3] src/tools/update-tasks.ts:160**: This concatenates every failure into one big `details` string, but `registerTool()` catches the thrown `Error` and runs it through `formatToolExecutionError()`, which trims generic error messages to 220 chars before returning them. On large all-failure batches, that turns into O(batch) string work and allocation for text the caller never sees. Limit `details` to a small prefix (for example `DisplayLimits.MAX_FAILURES_SHOWN`) or throw a compact counts-only summary here.
- **[P3] src/tools/__tests__/update-tasks.test.ts:1202**: This assertion hard-codes the SDK/axios wording (`"Request failed with status code 403"`), which makes the test fail if we later improve per-item errors without changing the behavior under review. Prefer asserting that the failure is attached to `bad-task` and that the message includes a stable signal like `403`, rather than matching the transport layer’s exact string.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab874aac0-6587-11f1-9753-9f35f10c1b07%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab874aac0-6587-11f1-9753-9f35f10c1b07&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
```diff
@@ -113,127 +120,53 @@ const updateTasks = {
...
+                    result.reason instanceof Error ? result.reason.message : String(result.reason),
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This still reduces API failures to `error.message`, so the cross-workspace move case becomes a generic `Request failed with status code 403` even when `responseData.error` has the actionable reason (`Not allowed to move objects out of a workspace`). That leaves callers without the per-item objection this PR is trying to surface, and makes it hard to decide whether to retry or drop the item. Pull a one-line message from the Todoist error payload/status here instead of using `Error.message` verbatim.

src/tools/update-tasks.ts
```diff
@@ -242,35 +175,163 @@ const updateTasks = {
...
+
+    // If no move parameters are provided, use updateTask without moveTask
+    if (!resolvedProjectId && !sectionId && !parentId) {
+        return await client.updateTask(id, updateArgs)
+    }
+
+    const moveArgs = createMoveTaskArgs(id, resolvedProjectId, sectionId, parentId)
+    const movedTask = await client.moveTask(id, moveArgs)
+
+    if (Object.keys(updateArgs).length > 0) {
```
doistbot, line 309: <big><kbd>🟡 P2</kbd></big>  A task that needs both a move and a metadata update can now be reported as a plain failure even after part of it succeeded. If `moveTask()` on line 293 succeeds and this follow-up `updateTask()` rejects, the task lands in `failures[]` even though the move already happened. With the new `retry only failed items` guidance, that makes combined move+update requests unsafe to replay. Catch the second call separately and surface that the move was applied (or otherwise model a partial-success outcome) so callers don't assume the task stayed unchanged.

doistbot (commented):
This PR successfully shifts `update-tasks` to independent per-task settling to gracefully capture and preserve successful updates within a batch.

Few things worth tightening:
- Extracting the actionable reason from the API error payload rather than surfacing a generic HTTP status code so callers know why a specific task failed.
- Handling the case where a combined move and metadata update partially succeeds (e.g., the move applies but the update rejects) to prevent callers from unsafely retrying the entire unchanged task.

<details>

- **[P3] src/tools/update-tasks.ts:160**: Concatenating every per-item error can create an excessively long string on large rejected batches. `formatToolExecutionError()` mitigates this downstream, but consider slicing `failures` here (e.g. `failures.slice(0, DisplayLimits.MAX_FAILURES_SHOWN)`) avoiding allocation overhead for text that will mostly be dropped anyway.
- **[P3] src/tools/__tests__/update-tasks.test.ts:1298**: Missing test coverage for the `DisplayLimits.MAX_FAILURES_SHOWN` truncation scenario (i.e. >3 failures and at least one success to prevent throwing). Adding a test ensures future refactors don't inadvertently remove the limit and send exceedingly large text blocks in the MCP response.
- **[P3] src/tools/update-tasks.ts:329**: This reimplements batch-failure rendering inside the tool instead of going through the shared response-builder path. `add-tasks`/`complete-tasks` already rely on `summarizeBatch()` to keep failure wording and truncation rules consistent, but `update-tasks` now has its own `DisplayLimits.MAX_FAILURES_SHOWN` logic and copy of the failure list format to maintain. Prefer composing the shared helper here, or extract a small shared failure-section builder in `response-builders.ts` and reuse it.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A98a29c20-6587-11f1-93b7-381bef77cc26%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A98a29c20-6587-11f1-93b7-381bef77cc26&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
engfragui: 💭 This was a pre-existing concern and it's also true (we don't surface `error.message`, but just relay the http error code). We could address this in a different PR if we wanted (but probably for all operations, not just `update-tasks`). I will not to anything related to this in this PR.

engfragui:
@scottlovegrove @gnapse Tagging you both because it's a chunky/meaty PR and I'd rather have "too much" input than not enough 😅 Take your time with this PR, it's not super priority, it's ok if it gets reviewed/approved next week.

src/tools/update-tasks.ts
```diff
@@ -113,127 +120,56 @@ const updateTasks = {
...
-            if (priority) {
-                updateArgs.priority = convertPriorityToNumber(priority)
-            }
+        // Each task is updated independently. A failure on one task (for example, the
+        // API rejecting a move with "Not allowed to move objects out of a workspace")
+        // must not discard the successful updates in the same batch, nor surface as a
+        // single opaque batch error that nudges the caller into retrying everything —
+        // a retry loop that can trip server-side abuse penalties. So we settle every
+        // task and report per-task outcomes.
+        const settled = await Promise.allSettled(
```
doistbot, line 151: <big><kbd>🟡 P2</kbd></big>  This now bypasses the existing transient-retry path. `registerTool()` only applies `executeWithRetry()` when `execute()` throws, but `Promise.allSettled()` turns a retryable 502/503/504 from `moveTask()` / `updateTask()` into a permanent `failures[]` entry on the first attempt. A single-task batch that gets one temporary 503 used to recover on retry; now it reports failure immediately. Please keep the per-item settling, but add retry around the individual SDK calls inside `processTaskUpdate()` so transient transport errors still use the repo’s retry policy without replaying already successful items.

doistbot (commented):
The shift to independent, per-task settling for `update-tasks` is a solid improvement that gracefully limits the blast radius of partial move and update failures.

Few things worth tightening:
- **Transient retries:** Because `Promise.allSettled` catches errors internally, let's push the existing retry wrapper down into the individual SDK calls so we maintain our resilience against temporary API errors.
- **Machine-readable states:** Adding a stable status field or code for partial updates will save programmatic callers from having to parse human-readable English error strings to figure out what happened.
- **Unbounded execution:** Since the tool now waits for and retains the outcome of every submitted item rather than failing fast, we should introduce a batch cap or bounded concurrency to protect against unbounded fan-out on large inputs.

<details>

- **[P3] src/tools/update-tasks.ts:358**: This logic duplicates the failure list formatting, truncation limit (`DisplayLimits.MAX_FAILURES_SHOWN`), and `+N more` wording from `summarizeBatch()` in `src/utils/response-builders.ts`. Consider extracting the failure section logic into a shared helper there so all tools supporting batch failures stay consistent without syncing code changes across files.
- **[P3] src/tools/__tests__/update-tasks.test.ts:1158**: This assertion hard-codes the SDK transport wording (`"Request failed with status code 403"`). The behavior introduced here is that `bad-task` lands in `failures[]`, not the exact axios message text. As written, any future cleanup of error rendering will break this test even if the feature still works. Prefer asserting a stable signal instead, such as the item id plus `error` containing `403` or being non-empty.
- **[P3] src/tools/__tests__/update-tasks.test.ts:1363**: This is redundant with the single-task partial-success test just above. That earlier case already awaits `updateTasks.execute(...)` for a batch whose only outcome is partial success, so it would fail immediately if the call threw. Keeping a second `resolves.toBeDefined()` case adds maintenance without covering a new regression path; consider folding the non-throwing expectation into the earlier test and dropping this one.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Acceb3300-6593-11f1-9451-5584e3e773ec%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Acceb3300-6593-11f1-9451-5584e3e773ec&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
```diff
@@ -242,35 +178,189 @@ const updateTasks = {
...
+    // partial success rather than a pure failure — otherwise the completed move would be
+    // discarded and the caller told to replay the whole item (re-doing the move).
+    try {
+        return { kind: 'applied', task: await client.updateTask(id, updateArgs) }
+    } catch (error) {
+        const message = error instanceof Error ? error.message : String(error)
+        return {
+            kind: 'partial',
+            task: movedTask,
+            error: `Move applied, but updating the other fields failed: ${message}. The move is already done; only the field changes did not apply.`,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This new `partial` outcome is only serialized as prose inside `failures[].error` (`"Move applied..."`). Callers following the new retry guidance now have to parse English to tell a partial from a true failure and risk re-sending the move anyway. `FailureSchema` already has `code`; give partials a stable discriminator (or add a dedicated status field) and keep `error` as the human-readable message.

src/tools/update-tasks.ts
doistbot, line 151: <big><kbd>🟡 P2</kbd></big>  `Promise.allSettled()` now guarantees we wait for and keep every per-task result until the slowest request finishes, and this tool still accepts an unbounded `tasks` array. That makes a large bad batch an unbounded fan-out + response-size path: we issue all requests, retain the full `settled` array, then serialize one `failures` entry per rejection. The old fail-fast path didn't need to buffer and return every failure. Please add a batch cap or process with bounded concurrency before shipping per-item failures.

src/tools/update-tasks.ts
engfragui: Decided to revert the change that breaks down the failures, so this is not applicable anymore.

src/tools/update-tasks.ts
```diff
@@ -113,127 +120,56 @@ const updateTasks = {
...
-            const deadlineDateUpdate = normalizeAliasValue(
-                deadlineDate,
-                DEADLINE_REMOVAL_ALIASES,
-                null,
-            )
-            if (deadlineDateUpdate !== undefined) {
-                updateArgs = { ...updateArgs, deadlineDate: deadlineDateUpdate }
-            }
+        settled.forEach((result, index) => {
+            const item = tasks[index]?.id ?? `Task ${index + 1}`
```
gnapse: nit: call this variable `task` if possible, not `item`

src/tools/update-tasks.ts
scottlovegrove: This feels valid, but simultaneously feels like a massive change would be required to have our cake and eat it.

src/tools/update-tasks.ts
```diff
@@ -113,127 +120,56 @@ const updateTasks = {
...
-            // Handle deadline changes if provided
-            const deadlineDateUpdate = normalizeAliasValue(
-                deadlineDate,
-                DEADLINE_REMOVAL_ALIASES,
-                null,
-            )
-            if (deadlineDateUpdate !== undefined) {
-                updateArgs = { ...updateArgs, deadlineDate: deadlineDateUpdate }
-            }
+        settled.forEach((result, index) => {
```
gnapse: nit: I'd suggest using a `for` loop instead. Not for aesthetics reasons. The main reason for me is that with this approach, a conceptual loop `continue` below has to be now a `return`. IMO it is more idiomatic to read `continue` which makes it clear you're in a loop and the instruction is to skip the rest of this loop turn and go to the next one. But reading `return` does not convey that information as clearly.

This is especially relevant in loops with long bodies of code inside them, like this one.

src/tools/update-tasks.ts
```diff
@@ -242,35 +178,189 @@ const updateTasks = {
...
+ * - `skipped`: the task carried no changes to make.
+ * - `applied`: every requested change succeeded; `task` is the final state.
+ * - `partial`: the move succeeded but the follow-up field update failed. `task` is the
+ *   moved task (so the completed move isn't discarded) and `error` explains what still
+ *   needs retrying. Anything else throws, so a pure failure is recorded by the caller.
+ */
+type TaskOutcome =
+    | { kind: 'skipped' }
+    | { kind: 'applied'; task: Task }
+    | { kind: 'partial'; task: Task; error: string }
```
scottlovegrove: I suspect this will eventually become a shared type when we move the other tools to this kind of process. 
```
type Outcome<T> =
    | { kind: 'skipped' }
    | { kind: 'applied'; item: T }
    | { kind: 'partial'; item: T; error: string }
```

src/tools/update-tasks.ts
```diff
@@ -242,35 +178,189 @@ const updateTasks = {
...
+            if (!validation.isValid) {
+                const errorMsg = validation.error?.message || 'Assignment validation failed'
+                const suggestions = validation.error?.suggestions?.join('. ') || ''
+                throw new Error(`Task ${id}: ${errorMsg}${suggestions ? `. ${suggestions}` : ''}`)
+            }
+
+            // Use the validated assignee ID
+            updateArgs = { ...updateArgs, assigneeId: validation.resolvedUser?.userId }
+        }
+    }
```
gnapse, line 291: Hard to read with too many nested things going on (3 levels of nested `if`s). I suspect that it could be better to extract the logic into a helper function.

gnapse (approved):

src/tools/update-tasks.ts
engfragui: I have decided not to introduce this in this PR, as I'm already changing enough things. If we do want to inform the user/agent what part of the update failed (the move or the update) we can do that in a follow-up PR (where this type becomes a shared one).

src/tools/update-tasks.ts
engfragui: @scottlovegrove Sorry, I had missed this comment, as doistbot posted it after I had already requested a review. Addressed, following doistbot suggestion (wrapping the single client operations inside `executeWithRetry()`).

src/tools/update-tasks.ts
```diff
@@ -89,18 +91,32 @@ const DUE_DATE_REMOVAL_ALIASES = ['remove', 'no date'] as const
 const DEADLINE_REMOVAL_ALIASES = ['remove', 'no date', 'no deadline'] as const
 const DUE_DATE_REMOVAL_VALUE = 'no date' as const
 
+// Cap the batch size (matching add-tasks) so a single call can't fan out an unbounded
+// number of concurrent SDK requests or buffer an unbounded failures response.
+const MAX_TASKS_PER_OPERATION = 25
```
engfragui: 💭 Same as the limit for `add-tasks` for example

gnapse:
@doistbot /review

src/tools/update-tasks.ts
```diff
@@ -242,35 +191,195 @@ const updateTasks = {
...
+        duration: durationStr,
+        responsibleUser,
+        priority,
+        labels,
+        deadlineDate,
+        ...otherUpdateArgs
+    } = task
+
+    // Resolve "inbox" to actual inbox project ID if needed
+    const resolvedProjectId = await resolveInboxProjectId({
```
doistbot: <a href="#">P2</a>  `resolveInboxProjectId()` calls `client.getUser()` for `projectId: "inbox"`, but this call is outside `executeWithRetry()`. Since its rejection is now absorbed by `Promise.allSettled()`, a transient 502/503/504 becomes a permanent per-item failure instead of receiving the retry this tool previously got from the outer wrapper. Wrap this resolution in `executeWithRetry()` (and add a `getUser()` 503-then-success test) so inbox moves retain the per-item retry guarantee.

doistbot (commented):
`update-tasks` now settles each item independently via `Promise.allSettled` and returns structured per-task `failures`, with a 25-item batch cap and per-item retries for transient 502/503/504 responses.

Few things worth tightening:

- `resolveInboxProjectId()` calls `client.getUser()` outside `executeWithRetry()`, so a transient 502/503/504 during inbox resolution becomes a permanent per-item failure instead of being retried—wrap it in `executeWithRetry()` to restore the retry guarantee.
- The move-applied-then-update-failed partial outcome is only identifiable by parsing the English `error` text; a stable discriminator like `code: "PARTIAL_MOVE_APPLIED"` would let callers safely retry only the field update.
- The `.max(25)` cap only runs at schema-parse time, so direct `execute()` calls (e.g., from `run-tool.ts`) with 26+ tasks still hit unbounded fan-out—enforce the count inside `execute()` as well.

<details>

- **<a href="#">P3</a> src/tools/update-tasks.ts:373**: The failure-list truncation (`failures.slice(0, DisplayLimits.MAX_FAILURES_SHOWN)` and `+N more`) duplicates `summarizeBatch()` from `response-builders.ts`. `add-tasks`, `complete-tasks`, and `uncomplete-tasks` all delegate to `summarizeBatch` for consistent failure formatting — `update-tasks` should do the same. The function already accepts per-item `{ item, error }` failures matching the `summarizeBatch` parameter shape.
- **<a href="#">P3</a> src/tools/update-tasks.ts:102**: This creates a second `MAX_TASKS_PER_OPERATION = 25`; `add-tasks.ts` already defines and exports the same batch cap. Move the shared limit to `utils/constants.ts` (or reuse the existing export) so the two batch tools cannot silently diverge when the cap changes.
- **<a href="#">P3</a> src/tools/update-tasks.test.ts:748**: `expectSingleFailure` (line 748) and `expectSingleMoveFailure` (line 1057) have identical bodies. The duplication means any change to the failure-assertion shape requires updating both. Since both `describe` blocks live in the same file, hoisting one shared helper to the outer `describe` scope (or module scope) would eliminate the copy without losing clarity — the caller already provides descriptive test names.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F502&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3d21a1d0-8cf0-11f1-9652-fd3ec6ebbf5e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3d21a1d0-8cf0-11f1-9652-fd3ec6ebbf5e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
```diff
@@ -242,35 +191,195 @@ const updateTasks = {
...
+
+    if (Object.keys(updateArgs).length > 0) {
+        try {
+            return {
+                kind: 'updated',
+                task: await executeWithRetry(() => client.updateTask(id, updateArgs)),
+            }
+        } catch (error) {
+            return {
+                kind: 'partial',
```
doistbot, line 317: <a href="#">P2</a>  This partial outcome is distinguishable only by parsing the English `error` text. A caller following the new retry guidance cannot reliably tell it must retry only the field update rather than replaying the already-applied move. Add a stable discriminator such as `code: "PARTIAL_MOVE_APPLIED"` to this failure (and include `code` in the local failure type) while keeping the message for display.

src/tools/update-tasks.ts
```diff
@@ -113,127 +135,54 @@ const updateTasks = {
...
-            if (dueStringUpdate !== undefined) {
-                updateArgs = { ...updateArgs, dueString: dueStringUpdate }
-            }
+        // Each task is updated independently. A failure on one task (for example, the
+        // API rejecting a move with "Not allowed to move objects out of a workspace")
+        // must not discard the successful updates in the same batch, nor surface as a
+        // single opaque batch error that nudges the caller into retrying everything —
+        // a retry loop that can trip server-side abuse penalties. So we settle every
+        // task and report per-task outcomes.
+        const settled = await Promise.allSettled(
```
doistbot, line 151: <a href="#">P2</a>  The new `.max(25)` cap only runs when the caller parses `ArgsSchema`. `scripts/run-tool.ts` calls `tool.execute(parsedArgs, client)` directly, and `execute()` does not validate its arguments, so a direct call with 26+ tasks still reaches this unbounded `Promise.allSettled` fan-out. Parse `args` at the start of `execute()` (as `add-tasks` does) or enforce the count here.

PR #510, opened 2026-06-18
fix: bind standalone HTTP server to localhost by default

gnapse:
## Summary

Bind the standalone `todoist-mcp-http` server to `127.0.0.1` by default instead of relying on Node's implicit all-interface listen behavior. Operators can still explicitly expose it with `HOST=0.0.0.0`, but the server now logs a warning and the docs clarify that this mode runs tool calls with the operator-provided `TODOIST_API_KEY` and should only be exposed behind trusted network/auth controls.

This keeps the local HTTP transport useful for development and local MCP clients while reducing the chance that a self-hosted instance accidentally exposes the operator's Todoist account to the network.

PR #516, opened 2026-06-24
fix: validate Host/Origin on HTTP transport to block DNS rebinding

scottlovegrove:
## Summary

The standalone `todoist-mcp-http` server dispatched `POST /mcp` to the MCP transport **without validating the `Host` or `Origin` header**. Binding to `127.0.0.1` (PR #510) did not protect it: a malicious website could use **DNS rebinding** to reach the loopback server same-origin (bypassing CORS) and invoke Todoist tools with the operator's `TODOIST_API_KEY` — with no caller credential.

Resolves the critical advisory **GHSA-833v-v58w-294h** (also raised as Doist/Issues#20524).

## What changed

- **New `requireTrustedHost` middleware** (`src/middleware/require-trusted-host.ts`) validates the `Host` and `Origin` headers against a trusted-hostname allowlist:
  - Defaults: `localhost`, `127.0.0.1`, `[::1]`.
  - Operators extend it with a new `ALLOWED_HOSTS` env var (comma-separated); a concrete non-loopback `HOST` (e.g. `192.168.1.50`) is auto-included. Bind wildcards (`0.0.0.0`/`::`) are never trusted.
  - Untrusted `Host` or `Origin` → `403` before the request reaches the transport. Absent `Origin` is allowed so non-browser clients (mcp-remote, curl) keep working.
  - Allowed hosts and incoming headers are canonicalised through the same parser, so equivalent IPv6 literals match.
- **Guard scoped to the `/mcp` routes**, running ahead of body parsing and token auth. `/health` is intentionally left unguarded so deployment health probes (which use the target's private IP in the `Host` header) keep working.
- **Refactor for testability:** the Express app is extracted into a pure `createHttpApp()` (`src/http-app.ts`), leaving `main-http.ts` as a thin bootstrap; shared host helpers moved to `src/utils/host.ts`.
- Docs (`docs/mcp-server.md`) and `CODEBASE.md` updated.

A per-request bearer token was considered but deferred — it would break existing no-credential local clients, and the advisory lists it as optional extra hardening. `Host`/`Origin` validation closes the DNS-rebinding vector.

## Scope

Only the optional HTTP transport (`todoist-mcp-http`) is affected. The stdio entrypoint is not browser-reachable and is unchanged.

## Testing

- New unit tests for the middleware and the allowlist builder (`require-trusted-host.test.ts`), the host helpers (`utils/host.test.ts`), and the wired middleware chain via an ephemeral server (`http-app.test.ts`).
- Full suite green (968 tests), type-check and format clean.
- Manually verified against the built server: `/mcp` with a spoofed `Host`/`Origin` → `403`; trusted `Host` → passes the guard (`401` for a bad token); `/health` with a spoofed `Host` → `200`; LAN mode (`HOST=0.0.0.0 ALLOWED_HOSTS=mcp.lan`) accepts the configured host and rejects others.

Fixes https://github.com/Doist/Issues/issues/20524

### DIFF
--- a/src/middleware/require-trusted-host.ts
+++ b/src/middleware/require-trusted-host.ts
@@ -1,12 +1,12 @@
 import { isIP } from 'node:net'
 import type { NextFunction, Request, RequestHandler, Response } from 'express'
-import { isLoopbackHost, normalizeListenHost } from '../utils/host.js'
+import { normalizeListenHost } from '../utils/host.js'
 
 type RequireTrustedHostOptions = {
     /**
      * Port-agnostic hostnames permitted in the `Host` and `Origin` headers.
-     * IPv6 entries must be bracketed, e.g. `[::1]`, to match how the URL parser
-     * normalises hostnames.
+     * IPv6 entries must be bracketed, e.g. `[::1]`. Invalid entries cause
+     * {@link requireTrustedHost} to throw at construction time (fail fast).
      */
     allowedHosts: string[]
 }
@@ -17,25 +17,53 @@ type RequireTrustedHostOptions = {
  */
 const DEFAULT_ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]']
 
+// Characters that may appear in a URL but never in a bare `host[:port]`
+// authority. Rejecting them keeps `Host` parsing from being fooled by URL-only
+// syntax such as `foo@127.0.0.1` (which `new URL()` would normalise to a
+// trusted host) or embedded paths/queries/fragments.
+const NON_AUTHORITY_CHARS = /[@/\\?#\s]/
+
 /**
- * Extract the hostname from a `Host` (scheme-less) or `Origin` (with scheme)
- * header value, port-agnostic and lowercased. Returns `undefined` when the
- * value cannot be parsed (e.g. the literal `null` origin).
+ * Canonicalise a bare `host[:port]` authority (a `Host` header value or an
+ * allowlist entry) to its port-agnostic, lowercased hostname. Returns
+ * `undefined` for anything that isn't a valid authority. IPv6 is canonicalised
+ * too, so `[2001:db8:0:0:0:0:0:1]` and `[2001:db8::1]` compare equal.
  */
-function parseHostname(value: string): string | undefined {
+function canonicalHostFromAuthority(value: string): string | undefined {
+    if (NON_AUTHORITY_CHARS.test(value)) {
+        return undefined
+    }
     try {
-        const url = value.includes('://') ? new URL(value) : new URL(`http://${value}`)
-        return url.hostname.toLowerCase()
+        return new URL(`http://${value}`).hostname.toLowerCase()
     } catch {
         return undefined
     }
 }
 
 /**
+ * Canonicalise an `Origin` header (`scheme://host[:port]`) to its lowercased
+ * hostname. Returns `undefined` for the literal `null` origin, malformed values,
+ * or anything carrying userinfo / path / query / fragment (a real Origin has
+ * none of those).
+ */
+function canonicalHostFromOrigin(value: string): string | undefined {
+    let url: URL
+    try {
+        url = new URL(value)
+    } catch {
+        return undefined
+    }
+    if (url.username || url.password || url.pathname !== '/' || url.search || url.hash) {
+        return undefined
+    }
+    return url.hostname.toLowerCase()
+}
+
+/**
  * Build the list of hostnames trusted in the Host and Origin headers. Always
  * includes the loopback defaults, plus any names from `allowedHostsEnv`
- * (comma-separated), plus the configured `host` when it is a concrete
- * non-loopback interface (e.g. a direct bind to 192.168.1.50). Bind wildcards
+ * (comma-separated), plus the configured `host` (so an explicit bind — loopback
+ * alias like 127.0.1.1, a LAN IP, etc. — is always trusted). Bind wildcards
  * (0.0.0.0/::) are never added because they are never valid Host header values —
  * operators exposing on a wildcard must enumerate client hostnames via
  * ALLOWED_HOSTS.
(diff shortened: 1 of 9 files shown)


src/middleware/require-trusted-host.ts
```diff
@@ -0,0 +1,134 @@
...
+            hosts.add(trimmed)
+        }
+    }
+
+    const normalizedHost = normalizeListenHost(host)
+    if (
+        normalizedHost &&
+        normalizedHost !== '0.0.0.0' &&
+        normalizedHost !== '::' &&
+        !isLoopbackHost(normalizedHost)
```
doistbot: <a href="#">P1</a>  If an operator explicitly binds to a loopback IP other than `127.0.0.1` (e.g., `HOST=127.0.1.1`, which is the default mapped localhost on some Linux distributions), `isLoopbackHost` returns `true`, causing the explicit host to be skipped here. Because `DEFAULT_ALLOWED_HOSTS` only covers `127.0.0.1` and `[::1]`, requests to this valid bind will be incorrectly rejected with a 403.

You can safely remove the `!isLoopbackHost(normalizedHost)` condition entirely. Explicit binds should always be allowed, and the `Set` will naturally deduplicate `127.0.0.1` and `[::1]` if they are passed.

src/middleware/require-trusted-host.ts
```diff
@@ -0,0 +1,134 @@
...
+const DEFAULT_ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]']
+
+/**
+ * Extract the hostname from a `Host` (scheme-less) or `Origin` (with scheme)
+ * header value, port-agnostic and lowercased. Returns `undefined` when the
+ * value cannot be parsed (e.g. the literal `null` origin).
+ */
+function parseHostname(value: string): string | undefined {
+    try {
+        const url = value.includes('://') ? new URL(value) : new URL(`http://${value}`)
```
doistbot: <a href="#">P2</a>  `parseHostname()` is now the trust boundary for `Host`, but `new URL(`http://${value}`)` accepts URL-only syntax that a `Host` header should reject. A value like `foo@127.0.0.1` normalizes to `127.0.0.1` here and would pass the allowlist even though the header itself is malformed. That makes the guard depend on WHATWG URL quirks instead of the authority grammar. Parse `Host` as `[host][:port]` (or reject `@`, `/`, `?`, `#`, etc. before URL parsing) so only valid authority values can be trusted.

doistbot (commented):
This PR adds a `requireTrustedHost` middleware that validates `Host` and `Origin` headers against a trusted-hostname allowlist to block DNS-rebinding attacks on the HTTP transport, along with a clean extraction of the Express app for testability.

Few things worth tightening:

- The `!isLoopbackHost(normalizedHost)` check in the allowlist builder incorrectly skips non-default loopback aliases like `127.0.1.1` (a common default on some Linux distros), causing valid explicit binds to be rejected with 403 at runtime. This condition can be safely removed — explicit binds should always be allowed and the `Set` handles deduplication.
- `parseHostname()` leans on `new URL()` which accepts URL-only syntax that a `Host` header should reject (e.g. `foo@127.0.0.1` normalizes to `127.0.0.1` and passes the allowlist). Parsing `Host` as `[host][:port]` or rejecting `@`, `/`, `?`, `#` before URL parsing would close this gap.
- Silently dropping unparseable `allowedHosts` entries in the middleware turns bad startup config into opaque 403s later. Validating/canonicalizing these entries at app build time (or inside `buildAllowedHosts`) and throwing on invalid values would surface misconfigurations early.

<details>

- **<a href="#">P3</a> src/middleware/require-trusted-host.ts:93**: Dropping unparseable `allowedHosts` entries here makes bad startup config fail later as opaque 403s instead of failing fast when the app is built. Since `createHttpApp()` now takes a raw `string[]` contract, a future caller can pass an unbracketed IPv6 literal or other malformed value and silently lose it from the allowlist. It would be safer to validate/canonicalize this input when building the app (or inside `buildAllowedHosts`) and throw on invalid entries.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F516&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad4dda5b0-6fb7-11f1-949b-8fd2587e5e9c%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad4dda5b0-6fb7-11f1-949b-8fd2587e5e9c&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/middleware/require-trusted-host.ts
scottlovegrove: Done in 42ee14c — removed the `!isLoopbackHost(normalizedHost)` condition in `buildAllowedHosts`, so an explicit bind to any loopback alias (e.g. `127.0.1.1`) is now added to the allowlist. The `Set` deduplicates the defaults. Added a regression test covering `127.0.1.1`.

src/middleware/require-trusted-host.ts
scottlovegrove: Done in 42ee14c — `Host` is now parsed as a strict `host[:port]` authority: values containing `@`, `/`, `\`, `?`, `#`, or whitespace are rejected before `new URL()`, so `foo@127.0.0.1` no longer normalises to a trusted host. `Origin` parsing likewise rejects userinfo and any non-empty path/query/fragment. Added tests for `foo@127.0.0.1` and `127.0.0.1/../evil`.

henningmu (approved):
Nice, no more complaints after Doistbot's comments got addressed. Let's merge this 🙌

rfgamaral (commented):
✅

PR #517, opened 2026-06-24
feat(projects): support archived projects in find-projects

scottlovegrove:
## Context

[Doist/automations#3207](https://github.com/Doist/automations/issues/3207): a user built an automation to delete their archived Todoist projects. It ran but found 0 archived projects, because the MCP integration had no way to **discover** them:

- `find-projects` only called `getProjects()`, which returns active projects only.
- The mapped project object never exposed `isArchived`, so a generated filter like `project?.isArchived === true` always evaluated false.

Deletion itself already worked — `delete-object` (`type: 'project'`) is archive-aware (workspace projects must be archived first; personal projects delete regardless). So the only real gap was discovery.

## Changes

- **`find-projects`**: new `archivedStatus` param — `'active'` (default, unchanged), `'archived'` (SDK `getArchivedProjects`), or `'all'` (both, fetched concurrently and merged). Archived projects have no server-side search, so `searchText` filters them client-side.
- **`isArchived`** surfaced on the shared `ProjectSchema` / `mapProject`, plus an `Archived` preview marker.
- New helpers in `tool-helpers`: `compileWildcardQuery` (client-side wildcard matching, compiled once per query), `fetchAllActiveProjects`, `fetchAllArchivedProjects` — mirroring the existing `searchAllProjects` family.
- **Docs**: server guidance + `delete-object` description now explain how to find and delete archived projects.

## Behaviour notes

- Pagination is preserved for single-source scopes (`active`, `archived`). The combined `all` scope and any `searchText` query return all matches without a cursor (same precedent as existing search behaviour).
- Backward compatible: omitting `archivedStatus` behaves exactly as before.

## Acceptance criteria

- [x] Automations can identify archived projects via the MCP
- [x] Automations can delete an archived project (existing `delete-object`)
- [x] Behaviour/limitations documented in tool schemas and server guidance

## Testing

- `tsc --noEmit`, build, and full suite (990 tests) pass; format clean.
- New tests: `find-projects` archived/all/search-combined paths and empty states; `compileWildcardQuery`/`matchesWildcardQuery` unit tests (incl. literal-backslash regression).

Closes https://github.com/Doist/automations/issues/3207

### DIFF
--- a/src/tool-helpers.ts
+++ b/src/tool-helpers.ts
@@ -156,6 +156,79 @@ export async function searchAllProjects(client: TodoistApi, query: string): Prom
 }
 
 /**
+ * Fetches all active (non-archived) projects across every page.
+ *
+ * @param client - The Todoist API client
+ * @returns Promise resolving to array of active projects
+ */
+export async function fetchAllActiveProjects(client: TodoistApi): Promise<Project[]> {
+    return fetchAllPages({
+        apiMethod: client.getProjects.bind(client),
+        args: {},
+        limit: ApiLimits.PROJECTS_MAX,
+    })
+}
+
+/**
+ * Fetches all archived projects across every page.
+ *
+ * The API has no server-side search for archived projects, so callers that need
+ * to filter by name should fetch all and filter client-side with
+ * {@link matchesWildcardQuery}.
+ *
+ * @param client - The Todoist API client
+ * @returns Promise resolving to array of archived projects
+ */
+export async function fetchAllArchivedProjects(client: TodoistApi): Promise<Project[]> {
+    return fetchAllPages({
+        apiMethod: client.getArchivedProjects.bind(client),
+        args: {},
+        limit: ApiLimits.PROJECTS_MAX,
+    })
+}
+
+/**
+ * Compiles a search query into a case-insensitive RegExp using the same wildcard
+ * semantics as server-side search (see {@link toWildcardQuery}): `*` matches any
+ * sequence, `\*` is a literal asterisk, and every other character (backslashes
+ * included) is matched literally. A query without an unescaped `*` matches as a
+ * substring; otherwise the whole name must match.
+ *
+ * Compile once and reuse the result when filtering many names (e.g. archived
+ * projects), rather than calling {@link matchesWildcardQuery} per item.
+ */
+export function compileWildcardQuery(query: string): RegExp {
+    const escapeRegex = (char: string) => char.replaceAll(/[.*+?^${}()|[\]\\]/g, '\\$&')
+    // Unescaped wildcard = `*` preceded by an even number of backslashes
+    const hasWildcard = /(?<!\\)(?:\\\\)*\*/.test(query)
+    const chars = [...query]
+    let pattern = ''
+    for (let i = 0; i < chars.length; i++) {
+        const char = chars[i] as string
+        if (char === '\\' && chars[i + 1] === '*') {
+            pattern += '\\*' // literal asterisk
+            i++
+        } else if (char === '*') {
+            pattern += '.*' // wildcard
+        } else {
+            pattern += escapeRegex(char)
+        }
+    }
+    // No wildcard → substring (unanchored) match; wildcard → match the whole name.
+    return new RegExp(hasWildcard ? `^${pattern}$` : pattern, 'iu')
+}
+
+/**
+ * Tests whether a name matches a search query using the same wildcard semantics
+ * as server-side search. Used for client-side filtering where the API exposes no
+ * search endpoint (e.g. archived projects). For bulk filtering, prefer
+ * {@link compileWildcardQuery} so the pattern is compiled only once.
+ */
+export function matchesWildcardQuery(name: string, query: string): boolean {
+    return compileWildcardQuery(query).test(name)
+}
+
+/**
  * Searches labels by name and fetches all matching pages.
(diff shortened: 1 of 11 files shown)


src/tool-helpers.ts
```diff
@@ -155,6 +155,79 @@ export async function searchAllProjects(client: TodoistApi, query: string): Prom
...
+ */
+export function compileWildcardQuery(query: string): RegExp {
+    const escapeRegex = (char: string) => char.replaceAll(/[.*+?^${}()|[\]\\]/g, '\\$&')
+    // Unescaped wildcard = `*` preceded by an even number of backslashes
+    const hasWildcard = /(?<!\\)(?:\\\\)*\*/.test(query)
+    const chars = [...query]
+    let pattern = ''
+    for (let i = 0; i < chars.length; i++) {
+        const char = chars[i] as string
+        if (char === '\\' && chars[i + 1] === '*') {
```
doistbot: <a href="#">P1</a>  The loop logic contradicts the `hasWildcard` regex above it because it doesn't consume `\\` as an escaped backslash. If the query is `\\*` (an escaped backslash followed by a wildcard), `hasWildcard` correctly returns true. However, the loop treats the first `\` as a literal, and the second `\` + `*` as a literal asterisk `\*`, resulting in a pattern that matches exactly `\\*` literally instead of `\` followed by a wildcard.

To fix this and align with server-side wildcard semantics, consume escaped backslashes:
```ts
        if (char === '\\' && chars[i + 1] === '\\') {
            pattern += '\\\\' // literal backslash
            i++
        } else if (char === '\\' && chars[i + 1] === '*') {
```

doistbot (commented):
This PR adds archived-project discovery to `find-projects` via a new `archivedStatus` param, surfaces `isArchived` on the project schema, and introduces client-side wildcard matching for archived projects that lack server-side search.

<details>

- **<a href="#">P3</a> src/tool-helpers.test.ts:413**: Missing test: `\*` query (escaped backslash followed by an unescaped wildcard). `toWildcardQuery`'s own test suite has `should detect unescaped wildcard after escaped backslash` (`toWildcardQuery('a\\*b')` → `'a\\*b'`, returned unchanged because the `*` is unescaped). `compileWildcardQuery` claims to use "the same wildcard semantics as server-side search" but has no corresponding test. The `hasWildcard` regex (shared with `toWildcardQuery`) correctly reports `true` for `a\*b` and anchors the result. But the pattern-building loop treats every `\*` pair as a literal asterisk regardless of whether the `\` is itself escaped, so it emits `a\*b` (literal backslash + literal asterisk) instead of `a` + literal `\` + `.*` + `b`. This means `compileWildcardQuery` diverges from server-side search for `\*` patterns — a query like `a\*b` would match different archived projects than active ones. A one-liner would expose it: `expect(matchesWildcardQuery('a\Xb', 'a\\*b')).toBe(true)` (server matches `a\` + anything + `b`; `compileWildcardQuery` returns `false`).
- **<a href="#">P3</a> CODEBASE.md:46**: The updated `tool-helpers.ts` summary lists `toWildcardQuery` (pre-existing) and `matchesWildcardQuery` (new), but omits `compileWildcardQuery` — the primary new helper for client-side wildcard matching and the one recommended for bulk filtering. `fetchAllActiveProjects` and `fetchAllArchivedProjects` are also new but follow the same fetch-all pattern as the already-unlisted `searchAllProjects`, so their omission is consistent. The missing `compileWildcardQuery` is the gap.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F517&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A450ad970-6fc7-11f1-8c56-f3dcdbad3ab7%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A450ad970-6fc7-11f1-8c56-f3dcdbad3ab7&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tool-helpers.ts
scottlovegrove: Fixed in 7ce4121 — `compileWildcardQuery` now consumes `\\` as a literal backslash before the `\*` branch, so `\\*` builds `\\.*` (literal backslash + wildcard) and matches server-side semantics. Added a regression test (`treats an escaped backslash followed by a wildcard as literal-backslash + wildcard`).

PR #531, opened 2026-07-08
fix: stop blaming the API token for 403s caused by plan limits

PotHix:
## Short description

Related to https://github.com/Doist/Issues/issues/20596

**What:** Stop appending "Verify your API token and access permissions, then retry." to 403 errors that are not auth problems — most notably `MAX_ITEMS_LIMIT_REACHED` (project at its task limit).

**Why:** The backend already returns a structured 403 body (`error_tag`, `error_code`, message), but `formatToolExecutionError` gave every 401/403 the same token-focused "Try next" hint. Users hitting the 300-item project cap were sent down an auth-debugging rabbit hole — ruling out tokens, scopes, and sharing — while their credentials were fine (two support tickets on Doist/Issues#20301).

**How:** `getNextStepHint` is now tag-aware:

- `*LIMIT_REACHED` tags → point at the limit: complete/archive/delete items or upgrade the plan
- other specific 403 tags (e.g. `PROJECT_FROZEN`) → defer to the error message and state explicitly it is **not** a token problem
- plain 403s (`FORBIDDEN` tag or no payload) and all 401s keep the existing token hint

**Trade-offs:**

- Matching `*LIMIT_REACHED` by suffix covers all backend limit tags (items, projects, sections, labels, reminders) without enumerating them, at the cost of assuming the backend keeps that naming convention.
- Unknown 403 tags get a neutral "not a token problem" hint rather than a tag-specific one — safer than guessing, and the backend message above carries the specifics.

## PR Checklist

### DIFF
--- a/src/tool-execution-error.test.ts
+++ b/src/tool-execution-error.test.ts
@@ -66,6 +66,77 @@ describe('formatToolExecutionError', () => {
         )
     })
 
+    it('points 403 limit errors at the limit, not the API token (Doist/Issues#20301)', () => {
+        const output = formatToolExecutionError({
+            responseData: {
+                error: 'Maximum number of items per user project limit reached',
+                error_code: 49,
+                error_tag: 'MAX_ITEMS_LIMIT_REACHED',
+                http_code: 403,
+                error_extra: {},
+            },
+        })
+
+        expect(output).toContain(
+            'Todoist API request failed (HTTP 403, code 49, tag MAX_ITEMS_LIMIT_REACHED).',
+        )
+        expect(output).toContain('Message: Maximum number of items per user project limit reached')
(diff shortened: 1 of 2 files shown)


scottlovegrove (approved):

henningmu:
@PotHix @scottlovegrove I simultaneously cooked up https://github.com/Doist/todoist-mcp/pull/532 to address the same problem.

I think this PR misses the batch case: if you use add-tasks to add multiple tasks, the error tag will be swallowed, #532 addresses this. Let me know what you think 🙌

scottlovegrove:
Let's go with #532 👍🏻

PotHix:
https://github.com/Doist/todoist-mcp/pull/532 it is! :)

PR #537, opened 2026-07-16
feat!: require node >=24, test on 24 & 26, support npm >=11

scottlovegrove:
Node 20 has reached end-of-life, so this raises the minimum supported node runtime to 24 and drops support for node 20 and 22.

## Changes

- **`.nvmrc`**: `v22` → `24`
- **`package.json`**: added an `engines` block requiring node `>=24` and npm `>=11`
- **`.github/workflows/ci.yml`** (test suite): now runs against a node matrix of `[24, 26]`, with the setup-node step using `node-version: ${{ matrix.node-version }}` instead of `node-version-file: .nvmrc`

Other workflows (release) continue to use a single baseline node via `.nvmrc` (now 24).

## CI

The test workflow now exercises both node 24 and node 26.

## Breaking change

Minimum supported node is now 24. Node 20 and 22 are no longer supported.


package.json
```diff
@@ -18,6 +18,10 @@
         "README.md"
     ],
     "license": "MIT",
+    "engines": {
+        "node": ">=24",
```
doistbot, line 22: <a href="#">P2</a>  Update `CODEBASE.md`, which still describes this project as supporting Node 18+. Contributors following that setup map will use a runtime this change now declares unsupported.

package.json
```diff
@@ -18,6 +18,10 @@
...
+        "npm": ">=11"
```
doistbot, line 23: <a href="#">P2</a>  The engines field requires npm ">=11", but Node 24 (the new minimum Node version) ships with npm 10.x by default in many releases. This constraint will trigger engine warnings (or installation failures if engine-strict is enabled) for users using the standard npm version bundled with Node 24. Unless there is a specific npm 11 feature this project relies on, consider loosening this constraint to ">=10" or removing it entirely to prevent setup friction.

doistbot (commented):
This PR raises the minimum supported Node version to 24, drops support for Node 20 and 22, updates CI to test against a Node 24/26 matrix, and adds an engines block requiring node >=24 and npm >=11.

Few things worth tightening:
- **CODEBASE.md still references Node 18+** as the supported minimum — update it to reflect the new Node 24 baseline so contributors don't follow an outdated setup path.
- **The npm >=11 engine constraint may be too strict** — Node 24 ships with npm 10.x in many releases, which would trigger engine warnings or installation failures. Unless a specific npm 11 feature is needed, consider loosening to >=10 or removing the constraint.
- **Consider `fail-fast: false` on the CI matrix** — since the goal is to verify both Node 24 and 26 independently, disabling fail-fast ensures you always get per-version results instead of cancelling one job when the other fails.

<details>

- **<a href="#">P3</a> .github/workflows/ci.yml:19**: The matrix uses the default `fail-fast: true`, so a failure on one node version cancels the other before it finishes. Since the point of the matrix is to verify both runtimes independently, consider `fail-fast: false` so you always get per-version results instead of losing signal on the cancelled job.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F537&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A06dd0380-8115-11f1-9b53-824d769ff903%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A06dd0380-8115-11f1-9b53-824d769ff903&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

nvignola (approved):
LGTM ✅

package.json
nvignola: Minor but IMO worth it, so it does't keep drifting.

scottlovegrove: Done in ec44850 — CODEBASE.md now states Node >=24 (and npm >=11).

package.json
scottlovegrove: Keeping npm >=11 intentionally — Node 24 ships with npm 11 (not 10.x), and we want to support npm 11 and 12.

package.json
scottlovegrove: Agreed — CODEBASE.md is updated to Node >=24 · npm >=11 in ec44850, so it stays in sync.

PR #543, opened 2026-07-30
feat(fetch-object): add includeChildren for subtasks and sub-projects

### LINKED ISSUE #542: Feature request: surface subtask/children info when fetching a task
### What's missing

fetch-object (type: task) only exposes the child→parent link (parent_id). There's no way to tell from a task whether it has children. The only way to find out is a second call to find-tasks with parentId: <id> — but you only think to do that if you already suspect the task has subtasks.

### Why

A dated "due today" parent can hide an undated subtask that never shows up in an AI agent's normal filter scan (overdue / today / no-date), since the subtask only surfaces if it matches its own filter criteria independently. The workaround — querying parentId for every task you look at, just in case — will burn a lot of tokens. I've had to stop using subtasks because it confuses my agent, which is unfortunate.

### Proposal

Add an optional parameter to fetch-object (e.g. includeChildren) that, when set, has the server make the extra parent_id-filtered call itself and return a child count (or list) alongside the task. Or just always return a childCount when fetching a task.

scottlovegrove:
Closes #542.

`fetch-object` only ever exposed the child→parent link (`parentId`), so nothing on a fetched task told you it *had* children. The only way to find out was a speculative `find-tasks` call with `parentId` — which an agent only makes if it already suspects subtasks exist. The result is the failure mode in the issue: a parent due today quietly hides an undated subtask that never appears in a normal overdue/today/no-date scan.

Setting `includeChildren` now returns `childCount` plus a compact list of the object's **direct** children — subtasks for a task, sub-projects for a project — and flags each child that has children of its own, so deeper nesting stays discoverable without returning the whole tree.

```
Found task: Create a utility to share redux state • id=5w8QJV23M8QQFPg6 • priority=p3 • project=6Pqhp2QXr7vrq7xc • subtasks=3
```
```json
"childCount": 3,
"children": [
  { "id": "5x3gpvWhMcMRQXV6", "content": "Write light spec and share with Frontend team", "checked": false, "hasChildren": false },
  { "id": "5wgmV5mHVGQF6R3c", "content": "Implement utility", "checked": false, "hasChildren": false },
  { "id": "5whFQWWWh66F86c6", "content": "Once merged, announce on affected issues:", "checked": false, "hasChildren": false }
]
```

## How children are read

**Tasks** — one `getTasks({ parentId })` call, then one cheap `limit: 1` probe per child to set `hasChildren` (the API carries no child count on a task). The cost tracks the size of the *answer*, not the size of the project: a leaf task costs exactly one extra call, which is the common case. I deliberately did not scan every task in the project and group by `parentId` in memory (what `get-overview` does) — that is unbounded in project size and sequentially paginated, which just moves the token problem server-side.

**Projects** — projects cannot be filtered by parent server-side, so the hierarchy comes from a single `fetchAllActiveProjects()` grouped by `parentId`, which gives exact grandchild presence for free. Workspace projects nest under folders rather than under a parent project, so they short-circuit to `childCount: 0` with no lookup at all.

The children fields sit at the top level of the structured content rather than inside `object`, which is a union of the shared `TaskSchema`/`ProjectSchema` used by 13+ other tools — nesting would have meant either polluting those or adding union variants to what is already the largest output schema in the repo.

## Behaviour worth knowing

- Only **active** subtasks are counted (`parent_id` listing excludes completed ones) and only **non-archived** sub-projects.
- `childCount` is the length of the returned list, capped at 25; when `hasMoreChildren` is set it is a floor, and `find-tasks` with `parentId` pages the rest.
- A failed children lookup reports `childrenError` instead of failing the fetch **or** silently omitting the count — an absent `childCount` would otherwise read as "no children", which is exactly the bug this is meant to prevent.
- `comment` and `section` ignore the parameter entirely: no fields, no API calls.

## Testing

12 new cases in `fetch-object.test.ts` covering leaf/populated/truncated children, grandchild flags, both failure paths, workspace short-circuit, and the ignored types. Full suite passes (972 tests), plus `type-check`, `format:check` and `lint:schemas`. Also smoke-tested against a real account: parent task with three undated subtasks, a leaf task, a project with four sub-projects, and a section.

### DIFF
--- a/CODEBASE.md
+++ b/CODEBASE.md
@@ -126,6 +126,7 @@ New tool? Full checklist in `AGENTS.md`. Short version: copy `add-tasks.ts`; imp
 - `tool-names.ts` — `ToolNames` enum of every registered tool name
 - `output-schemas.ts` — Reusable Zod schemas: TaskSchema, ProjectSchema, SectionSchema, CommentSchema, etc.
 - `schema-helpers.ts` — Zod builders used across tools
+- `children.ts` — `ChildrenOutputSchema` plus `getTaskChildren`/`getProjectChildren` for returning an object's direct subtasks or sub-projects
 - `priorities.ts` — `"p1"`–`"p4"` ↔ SDK integer conversion (**strings only in tool I/O**)
 - `duration-parser.ts` — `"2h30m"` ↔ ms, plus `formatDuration`
 - `date.ts` — date parsing/formatting (ISO, Todoist strings)
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -119,7 +119,7 @@ You have access to comprehensive Todoist management tools for personal productiv
 
 **General Operations:**
 - **delete-object**: Remove projects, sections, tasks, comments, labels, filters, reminders, or location reminders by type and ID. Deletes both active and archived projects (workspace projects must be archived first; use find-projects with archivedStatus to locate archived projects)
-- **fetch-object**: Fetch a single task, project, comment, or section by its ID
+- **fetch-object**: Fetch a single task, project, comment, or section by its ID. Pass includeChildren=true to also get its direct children (subtasks for a task, sub-projects for a project) with a childCount - use this to check whether a task hides subtasks rather than a speculative find-tasks call
 - **reorder-objects**: Reorder sibling projects or sections, and optionally move projects to a new parent. For projects: set order to reorder siblings, and/or set parentId to move under a new parent (use "root" for top level). For sections: set order to reorder within a project
 - **user-info**: Get user details including timezone, goals, and plan information
 
--- a/src/tools/fetch-object.test.ts
+++ b/src/tools/fetch-object.test.ts
@@ -400,10 +400,19 @@ describe(`${FETCH_OBJECT} tool`, () => {
                 expect(result.textContent).toContain('subtasks=unavailable')
             })
 
-            it('should report a failed grandchild probe without failing the fetch', async () => {
+            it('should keep the subtask list when a grandchild probe fails', async () => {
                 mockTodoistApi.getTasks.mockImplementation(async (args) => {
                     if (args?.parentId === 'task123') {
-                        return { results: [createMockTask({ id: 'sub1' })], nextCursor: null }
+                        return {
+                            results: [
+                                createMockTask({ id: 'sub1' }),
+                                createMockTask({ id: 'sub2' }),
+                            ],
+                            nextCursor: null,
+                        }
+                    }
+                    if (args?.parentId === 'sub2') {
+                        return { results: [], nextCursor: null }
                     }
                     throw new Error('Probe failed')
                 })
@@ -413,8 +422,36 @@ describe(`${FETCH_OBJECT} tool`, () => {
                     mockTodoistApi,
                 )
 
-                expect(childrenOf(result).childrenError).toBe('Probe failed')
-                expect(result.structuredContent).not.toHaveProperty('childCount')
+                expect(childrenOf(result)).toMatchObject({
+                    childCount: 2,
+                    children: [
+                        { id: 'sub1', hasChildren: false },
+                        { id: 'sub2', hasChildren: false },
+                    ],
+                    childrenError:
+                        'Could not check 1 of 2 subtasks for subtasks of their own; those are reported as hasChildren=false.',
+                })
+                expect(result.textContent).toContain('subtasks=2 (partial)')
+            })
+
+            it('should fetch the task and its children concurrently', async () => {
+                const calls: string[] = []
+                mockTodoistApi.getTask.mockImplementation(async () => {
+                    calls.push('getTask')
+                    return createMockTask({ id: 'task123' })
+                })
+                mockTodoistApi.getTasks.mockImplementation(async () => {
+                    calls.push('getTasks')
+                    return { results: [], nextCursor: null }
+                })
+
+                await fetchObject.execute(
+                    { type: 'task', id: 'task123', includeChildren: true },
+                    mockTodoistApi,
+                )
+
+                // Both requests are issued before either has resolved.
+                expect(calls).toEqual(['getTask', 'getTasks'])
             })
         })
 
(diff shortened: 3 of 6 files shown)


doistbot (commented):
This PR adds an `includeChildren` parameter to `fetch-object` that returns direct child subtasks or sub-projects along with a `childCount` and per-child `hasChildren` flag, making nested items discoverable without speculative follow-up calls. No inline issues were flagged. One resilience improvement worth considering: the per-child grandchild probes currently use `Promise.all`, so a single transient probe failure discards the already-fetched child listing — switching to `Promise.allSettled` would keep partial results intact when one call fails.

<details>
<summary>Optional follow-up notes (7)</summary>

- **<a href="#">P3</a> src/tools/fetch-object.ts:111**: `Promise.all` on the grandchild probes means one failing probe discards the entire subtask listing (already fetched on L106). If probe #3 of 5 fails on a transient error, the agent sees only `childrenError` with no subtask list at all — even though 4 subtasks were resolved fine. Consider `Promise.allSettled` so the valid subtasks survive. Failed probes could default `hasChildren` to `false` (treating uncertain as leaf), which is a reasonable tradeoff for preserving the rest of the listing. The `resolveChildren` wrapper would still catch a failure of the *initial* child fetch, which is the more critical path.
- **<a href="#">P3</a> src/tools/fetch-object.ts:24**: These child schemas duplicate fields already defined by the exported `TaskSchema` and `ProjectSchema` below. Derive them with `TaskSchema.pick(...).extend({ hasChildren: ... })` and the equivalent project schema so field types and descriptions stay tied to the shared output contract rather than drifting.
- **<a href="#">P3</a> src/tools/fetch-object.ts:145**: This rebuilds the personal-project parent index and sibling-order logic already implemented by `buildProjectTree` in `get-overview.ts`. Extract the hierarchy/index construction into a shared helper and use it from both tools, so parent handling and ordering have one source of truth.
- **<a href="#">P3</a> src/tools/fetch-object.ts:63**: `childCount` is always equal to `children.length` — it duplicates the array length at every return site. Removing it would simplify the schema, the `ChildrenResult` type, and the output shape without losing any information. Consumers can trivially derive the count from the array.
- **<a href="#">P3</a> src/tools/fetch-object.ts:79**: The `hasMoreChildren` description only mentions `find-tasks` for pagination, but this field also appears on project responses where `find-tasks` is the wrong tool. The hint should be generic (e.g. "Use find-tasks with parentId to page through remaining subtasks, or find-projects to locate sub-projects") or split into type-specific descriptions.
- **<a href="#">P3</a> src/tools/fetch-object.ts:201**: Start the task fetch and children lookup concurrently. `getTask(id)`'s result is not needed by `getTaskChildren`, so successful `includeChildren` task requests currently wait an extra network round-trip before the child listing and probes begin. Create both promises before awaiting them together, while retaining `resolveChildren` for child lookup failures.
- **<a href="#">P3</a> src/tools/fetch-object.test.ts:459**: The task path has a test for truncation (`should flag a truncated child list` at line 370) that verifies `hasMoreChildren: true` when the subtask list exceeds `CHILDREN_MAX`. The project path has no equivalent — `getProjectChildren` uses the same `direct.slice(0, ApiLimits.CHILDREN_MAX)` pattern and sets `hasMoreChildren`, but no test exercises it. A regression in the project truncation logic (e.g., changing the constant or removing the slice) would go unnoticed. Add a test with >25 sub-projects that asserts `hasMoreChildren: true` and `subProjects=25+` in the textContent.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F543&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1c08f100-8c01-11f1-86b2-cba39b633538%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1c08f100-8c01-11f1-86b2-cba39b633538&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove:
Addressed in 4bdf5ab.

Took the main call-out: the grandchild probes now use `Promise.allSettled`, so a transient probe failure costs its own nesting flag rather than the subtask listing that was already fetched. Unprobed children come back as `hasChildren=false`, but the gap is named in `childrenError` and the text summary reads `subtasks=2 (partial)` — defaulting to false silently would have been the same "reads as no children" mistake the field exists to prevent.

Also taken: task and children now fetch concurrently (the child lookup keys off the parent id alone); child schemas derive from `TaskSchema`/`ProjectSchema` via `.pick().extend()`; the `hasMoreChildren` hint covers sub-projects as well as subtasks; and sub-project truncation has a test. Separately, @scottlovegrove asked for the lookups to move out of the tool file, so they now live in `src/utils/children.ts` alongside the shared `ChildrenOutputSchema`.

Two I'm not taking. **Sharing the parent index with `buildProjectTree`**: that builds a fully nested, recursively sorted tree, whereas this needs a flat parent→children index — unifying them means one helper serving two different shapes for little gain. **Dropping `childCount`**: it is redundant against `children.length` for a programmatic consumer, but the consumer here is a model, and an explicit `0` is read far more reliably than an absent array. It is also what makes `childrenError` legible — count present means the list is trustworthy, count absent means the lookup failed outright.

pedroalves0 (commented):
✅

PR #546, opened 2026-08-03
fix(retry): add full jitter to backoff delays

scottlovegrove:
## Context

[Janusz flagged on Comms](https://comms.todoist.com/69/ch/AK6drPS98yMaYmyWmztS2/t/CdffD1JCLJMYmWYSf3Fi2) that `todoist-mcp` accounts for almost all failing task moves the API sees, and that backend locking is changing so conflicting concurrent moves in the same task tree will fail with **503 + Retry-After**. That makes our retry behaviour load-bearing in a way it wasn't before.

Retries used a fixed exponential delay — 500ms, then 1000ms, identical for every caller. A batch of requests that fails together therefore retries in lockstep and arrives as another simultaneous burst, which is the same contention that caused the first failure.

This is the first of three PRs from that thread. It is independent of the other two and safe to land on its own.

## What changed

The delay is now a random point in `[0, cap]` (full jitter) rather than the cap itself, with a 50ms floor so a low draw can't turn a retry into an effectively immediate re-request. The floor is clamped by the cap, so an explicit zero-delay config — which the tests use — still waits nothing.

`RetryConfig` gains an injectable `random` (defaulting to `Math.random`) and `getRetryDelay` is exported, so delays are assertable rather than timing-dependent.

Deliberately unchanged: `maxRetries` stays at 2, and 429 stays non-retryable. More attempts under lock contention would mean more load, not fewer failures.

## Testing

The two existing exact-delay tests now inject `random: () => 1`, which pins the ceiling exactly. New coverage: the floor at the lowest draw, an explicit zero-delay config beating the floor, delays staying inside `[floor, ceiling]` across 50 draws, and a regression guard that two retriers with different draws get different delays.

`npm test` (1062 tests), `npm run type-check`, `npm run check` all pass.

## Follow-up, not in this PR

Honouring the `Retry-After` header itself. `TodoistRequestError` carries no response headers, but `createTrackedFetch` in `src/usage-tracking.ts` does see the raw `Response`. Doing it there would stack a third retry layer on top of `executeWithRetry` and the SDK's own `fetchWithRetry`, multiplying attempts, so it wants its own change that moves 5xx retry authority down to the transport.

### DIFF
--- a/src/utils/retry.test.ts
+++ b/src/utils/retry.test.ts
@@ -1,4 +1,10 @@
-import { executeWithRetry, extractHttpStatusCode, isTransientError } from './retry.js'
+import {
+    executeWithRetry,
+    extractHttpStatusCode,
+    getRetryDelay,
+    isTransientError,
+    MIN_RETRY_DELAY_MS,
+} from './retry.js'
 
 const NO_DELAY: { baseDelayMs: number; maxDelayMs: number } = {
     baseDelayMs: 0,
@@ -222,7 +228,7 @@ describe('executeWithRetry', () => {
         expect(fn).toHaveBeenCalledTimes(2)
     })
 
-    it('should use exponential backoff delays', async () => {
+    it('should use exponential backoff delays, drawing up to the ceiling', async () => {
         vi.useFakeTimers()
 
         const error503 = Object.assign(new Error('HTTP 503: Service Unavailable'), {
@@ -231,7 +237,12 @@ describe('executeWithRetry', () => {
         const fn = vi.fn().mockRejectedValueOnce(error503).mockResolvedValue('recovered')
         const sleepSpy = vi.spyOn(globalThis, 'setTimeout')
 
-        const promise = executeWithRetry(fn, { baseDelayMs: 500, maxDelayMs: 2000 })
+        // random: () => 1 draws the top of the jitter range, making the ceiling exact
+        const promise = executeWithRetry(fn, {
+            baseDelayMs: 500,
+            maxDelayMs: 2000,
+            random: () => 1,
+        })
         await vi.advanceTimersByTimeAsync(500)
         const result = await promise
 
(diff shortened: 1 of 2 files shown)


doistbot (commented):
This PR replaces fixed exponential backoff delays with full jitter (a random value in [0, cap] with a 50ms floor clamped by the cap), adds an injectable random source for deterministic testing, and exports getRetryDelay for assertability. No inline issues were flagged. The implementation is sound — the floor is correctly clamped so zero-delay configs still wait nothing, the injectable random makes delays testable rather than timing-dependent, and the scope is well-contained within the retry module without touching any security-sensitive surface.

<details>

- **<a href="#">P3</a> src/utils/retry.test.ts:327**: Closing `executeWithRetry` here places the existing “should use default config when none provided” execution test inside `describe('getRetryDelay')` below. Move that test above this closing brace (or into its own execution suite) so the test hierarchy matches what it exercises.
- **<a href="#">P3</a> src/utils/retry.test.ts:338**: The test name says "should never return the ceiling when the draw is lower" but the implementation uses `Math.round`, so a draw close enough to 1.0 (e.g. 0.9995 with cap=1000 → `Math.round(999.5) = 1000`) *can* reach the ceiling. The test only checks `random: () => 0.5`, which doesn't exercise that edge. Rename to something like "should return less than the ceiling for a moderate draw" to match what it actually verifies.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F546&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac5faf200-8f4c-11f1-95cb-c4b3947614c3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac5faf200-8f4c-11f1-95cb-c4b3947614c3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

gnapse (commented):
✅

PR #547, opened 2026-08-03
fix(update-tasks): bound concurrent write requests

scottlovegrove:
## Context

[Janusz flagged on Comms](https://comms.todoist.com/69/ch/AK6drPS98yMaYmyWmztS2/t/CdffD1JCLJMYmWYSf3Fi2) that `todoist-mcp` submits large numbers of concurrent `POST /api/v1/tasks/<id>/move` requests — one case moved 88 tasks at once — and that by volume almost all failing moves platform-wide come from us. Backend locking is being fixed such that concurrent moves within the same task tree will fail with 503 + Retry-After, so unbounded fan-out stops being merely rude and starts being the failure.

`update-tasks` fired every task in a batch through `Promise.allSettled` with no limiter, and the 25-item cap is enforced *per call*, so an agent making four parallel calls reached 100 in-flight writes with nothing to stop it. There was no concurrency limiter anywhere in the repo.

Second of three PRs from that thread. Stacked on #546. No semantic change to what a call does — only to how many requests it has open at once.

## What changed

New `src/utils/concurrency.ts`: a FIFO semaphore with two lanes, applied in `update-tasks` around the move calls and the `updateTask` fan-out.

**Limiters are keyed by account, not by client instance.** `src/http-app.ts` calls `getMcpServer()` per request, so parallel `POST /mcp` calls get distinct `TodoistApi` instances — an instance-keyed `WeakMap` would have bounded nothing in the deployment that matters. Registration happens in `createTodoistClient`, keyed on a SHA-256 of the token so no credential is retained. Clients built outside that factory (tests, direct SDK use) fall back to a shared pair, so no call site is ever unbounded.

`ConcurrencyLimits.TASK_MOVES` is **1**. The lock covers the whole task tree, and a tree spans a task's source as well as its destination — two sibling subtasks moved to different projects still contend, and the loser fails. Grouping by destination (PR 3) doesn't help with that, and since that PR collapses the common case to a single request anyway, serialising costs little. Accepted worst case: 4 concurrent calls × 25 distinct destinations ≈ 100 serialised requests ≈ 15s, which is pathological rather than realistic. `WRITES` is 4.

**Why the tool layer and not `customFetch`:** gating inside the transport would catch every call site automatically, but the SDK installs its 30s timeout `AbortSignal` *before* invoking `customFetch`, so queue time would be charged against the request timeout and long-queued requests would fail without ever running. The reasoning is recorded in the module JSDoc so nobody moves it later.

## Testing

New `src/utils/concurrency.test.ts` covers the semaphore directly: the cap holds under load, FIFO order is preserved, a rejecting task releases its slot, and — the subtle one — a task arriving in the window where a slot is being handed off to a queued task cannot slip past the limit. Plus per-account sharing, account isolation, separate move/write lanes, and the unregistered-client fallback.

`update-tasks.test.ts` gains in-flight assertions driven by deferred promises: 4 tasks to 4 destinations peak at 1 concurrent move, 10 content-only updates peak at `ConcurrencyLimits.WRITES`.

`npm test` (1062 tests), `npm run type-check`, `npm run check` all pass.

## Follow-up, not in this PR

The write lane should also go on `manage-assignments`, `add-tasks`, `update-projects`, `update-sections`, `add-comments` and `reschedule-tasks`, all of which fan out unbounded today.

### DIFF
--- a/src/utils/concurrency.ts
+++ b/src/utils/concurrency.ts
@@ -1,0 +1,136 @@
+import { createHash } from 'node:crypto'
+import { ConcurrencyLimits } from './constants.js'
+
+/**
+ * Wraps a task so it only runs once a slot is free.
+ */
+type Limiter = <T>(fn: () => Promise<T>) => Promise<T>
+
+type LimiterPair = {
+    /** Serialises task-move requests, which contend on server-side tree locks. */
+    moves: Limiter
+    /** Bounds all other write requests. */
+    writes: Limiter
+}
+
+/**
+ * A FIFO semaphore. Tasks beyond `maxConcurrent` queue and start as slots free up.
+ *
+ * Deliberately applied at the tool layer rather than inside the SDK's
+ * `customFetch`: the SDK installs its request-timeout `AbortSignal` *before*
+ * calling `customFetch`, so gating there would charge queue time against the
+ * request timeout and fail long-queued requests that never actually ran.
+ */
+function createLimiter(maxConcurrent: number): Limiter {
+    if (maxConcurrent < 1) {
+        throw new Error(`maxConcurrent must be at least 1, received ${maxConcurrent}`)
+    }
+
+    const queue: Array<() => void> = []
+    let active = 0
+
+    async function acquire(): Promise<void> {
+        if (active < maxConcurrent) {
+            active++
+            return
+        }
+        // A queued task inherits the releasing task's slot, so `active` stays put.
+        await new Promise<void>((resolve) => queue.push(resolve))
+    }
+
+    function release() {
+        const next = queue.shift()
+        if (next) {
+            next()
+            return
+        }
+        active--
+    }
+
+    return async function limit<T>(fn: () => Promise<T>): Promise<T> {
+        await acquire()
+        try {
+            return await fn()
+        } finally {
+            // Release in `finally` so a rejected task never leaks its slot.
+            release()
+        }
+    }
+}
+
+function createLimiterPair(): LimiterPair {
+    return {
+        moves: createLimiter(ConcurrencyLimits.TASK_MOVES),
+        writes: createLimiter(ConcurrencyLimits.WRITES),
+    }
+}
+
+/**
+ * Limiters are keyed by account rather than by client instance because the HTTP
+ * transport builds a fresh `TodoistApi` for every request — instance-keyed
+ * limiters would let concurrent requests for one account fan out without bound.
+ */
+const limitersByAccount = new Map<string, LimiterPair>()
+
+/** Lets a tool find its account's limiters from the client it was handed. */
+const limitersByClient = new WeakMap<object, LimiterPair>()
+
+/**
+ * Catches clients built outside `createTodoistClient` (tests, direct SDK use) so
+ * every call site is bounded even when nothing registered it.
+ */
+let fallbackLimiters: LimiterPair | undefined
+
+function getFallbackLimiters(): LimiterPair {
+    fallbackLimiters ??= createLimiterPair()
+    return fallbackLimiters
+}
+
+/**
+ * Derives a stable per-account key. The raw token is never stored — only its
+ * digest — so limiter bookkeeping can't leak credentials.
+ */
+function accountKeyFromApiKey(apiKey: string): string {
+    return createHash('sha256').update(apiKey).digest('hex')
+}
+
+/**
+ * Associates a client with its account's limiters, sharing one pair across every
+ * client built for the same account.
+ */
+function registerClientLimiters(client: object, apiKey: string): void {
+    const key = accountKeyFromApiKey(apiKey)
+    let limiters = limitersByAccount.get(key)
+    if (!limiters) {
+        limiters = createLimiterPair()
+        limitersByAccount.set(key, limiters)
+    }
+    limitersByClient.set(client, limiters)
+}
+
+function getLimiters(client: object): LimiterPair {
+    return limitersByClient.get(client) ?? getFallbackLimiters()
+}
+
+function getMoveLimiter(client: object): Limiter {
+    return getLimiters(client).moves
+}
+
+function getWriteLimiter(client: object): Limiter {
+    return getLimiters(client).writes
+}
+
+/** Test-only: drops all registered limiters so cases start from a clean slate. */
+function resetLimitersForTesting(): void {
+    limitersByAccount.clear()
+    fallbackLimiters = undefined
+}
+
+export {
+    createLimiter,
+    getMoveLimiter,
+    getWriteLimiter,
+    type Limiter,
+    registerClientLimiters,
+    resetLimitersForTesting,
+}
--- a/src/utils/concurrency.ts
+++ b/src/utils/concurrency.ts
@@ -13,6 +13,12 @@ type LimiterPair = {
     writes: Limiter
 }
 
+type QueuedTask = {
+    admit: () => void
+    abandon: (reason: Error) => void
+    timer?: NodeJS.Timeout
+}
+
 /**
  * A FIFO semaphore. Tasks beyond `maxConcurrent` queue and start as slots free up.
  *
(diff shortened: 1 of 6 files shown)


src/utils/concurrency.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * A FIFO semaphore. Tasks beyond `maxConcurrent` queue and start as slots free up.
+ *
+ * Deliberately applied at the tool layer rather than inside the SDK's
+ * `customFetch`: the SDK installs its request-timeout `AbortSignal` *before*
+ * calling `customFetch`, so gating there would charge queue time against the
+ * request timeout and fail long-queued requests that never actually ran.
+ */
+function createLimiter(maxConcurrent: number): Limiter {
+    if (maxConcurrent < 1) {
```
doistbot: <a href="#">P2</a>  `maxConcurrent` accepts `NaN`, `Infinity`, and fractional values: `NaN` leaves all tasks queued forever, `Infinity` disables the bound, and `1.5` admits two tasks. Require a positive integer (for example, `Number.isInteger(maxConcurrent) && maxConcurrent >= 1`) so this exported limiter always enforces its advertised cap.

src/utils/concurrency.ts
```diff
@@ -0,0 +1,136 @@
...
+        throw new Error(`maxConcurrent must be at least 1, received ${maxConcurrent}`)
+    }
+
+    const queue: Array<() => void> = []
```
doistbot: <a href="#">P1</a>  `queue` is unbounded. Each concurrent `update-tasks` call immediately schedules up to 25 operations, while the HTTP endpoint accepts concurrent POSTs for the same account; once arrivals exceed the one-move/four-write service rate, this array retains an unlimited number of closures and request promises. Add a maximum pending count and fail/return an overload error when it is reached (or apply request-level backpressure before scheduling the batch).

doistbot (commented):
This PR adds an account-keyed FIFO semaphore to bound concurrent move and write requests in `update-tasks`, preventing unbounded fan-out that was causing platform-wide failures.

Few things worth tightening:

- **Validate `maxConcurrent`** — currently `NaN` silently queues everything forever, `Infinity` disables the bound, and fractional values like `1.5` admit extra tasks. Require a positive integer so the limiter always enforces its cap.
- **Bound the pending queue** — multiple concurrent `update-tasks` calls each schedule up to 25 operations against a single move slot, so under load the queue array can retain an unbounded number of closures and request promises. Consider a maximum pending count that returns an overload error when reached, or apply backpressure before scheduling the batch.

<details>
<summary>Optional follow-up notes (5)</summary>

- **<a href="#">P3</a> src/utils/concurrency.ts:1**: Add this reusable concurrency utility to CODEBASE.md's `src/utils` catalog. It is intended for additional write tools, so documenting it there will help future callers reuse the shared limiter rather than introducing another implementation.
- **<a href="#">P3</a> src/utils/concurrency.test.ts:14**: The `deferred()` helper reimplements `Promise.withResolvers()`, which has been available since Node 22. The project requires Node ≥24, so the native API could replace this 7-line helper. Each call site would become `const { promise, resolve } = Promise.withResolvers<T>()`. (The `reject` return value is never used in any test case, so it can be omitted.)
- **<a href="#">P3</a> src/tools/update-tasks.test.ts:1464**: `trackInFlight` calls `mockImplementation` without restoring the mock afterward. Vitest's `vi.clearAllMocks()` (called in `beforeEach`) does not reset implementations — it only clears call history. After the concurrency tests, both `mockTodoistApi.moveTask` and `mockTodoistApi.updateTask` retain their deferred implementations. If someone adds a test after this describe block that calls either mock without first setting its own implementation via `mockResolvedValue`, the test will hang indefinitely because the mock pushes resolvers into a dead `release` array. Fix: either restore the original mock in an `afterEach` within the describe block, or have `trackInFlight` return a cleanup function that each test calls.
- **<a href="#">P3</a> src/utils/concurrency.test.ts:153**: `should return the task result` verifies that `createLimiter` passes through the wrapped function's return value — i.e., that `return await fn()` works. This is already covered by `should give moves and writes separate lanes` (line 221), which asserts `getWriteLimiter` resolves to `'written'` through the same `createLimiter` code path, and by the `update-tasks` integration tests that check returned task data. Consider removing this test to reduce trivia coverage.
- **<a href="#">P3</a> src/utils/concurrency.ts:73**: The in-memory `Map` and `WeakMap` limiter state (lines 73–82) means per-account concurrency bounds aren't shared across process boundaries. If the HTTP server runs with multiple instances behind a load balancer, the effective limit per account becomes `instances × limit`. The [Service Production Readiness standard](https://handbook.doist.com/doc/standard-service-production-readiness-qgNRng8TJx) (Concurrency and Scaling) says: "Assume services will run with multiple parallel instances. … Shared state must be externalized." Consider either documenting that the HTTP transport is intentionally single-instance, or using an external store (e.g., Redis) for the limiter slots when `TODOIST_BASE_URL` or another env flag indicates a multi-instance deployment. The stdio transport is unaffected since it's always single-process.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F547&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad56f8ca0-8f4c-11f1-9f81-82978054332a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad56f8ca0-8f4c-11f1-9f81-82978054332a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/utils/concurrency.ts
scottlovegrove: Done in c919c4f — `createLimiter` now requires a positive integer (`Number.isInteger(maxConcurrent) && maxConcurrent >= 1`). Confirmed the NaN path you describe: `NaN < 1` is false, so it passed the old guard and then failed every `active < maxConcurrent` comparison, queueing work forever. Added table-driven cases for zero, negative, NaN, Infinity and 1.5.

src/utils/concurrency.ts
scottlovegrove: Addressed in c919c4f, though with a deadline rather than a pending cap.

Queued closures each correspond to an MCP request that is already in flight and holding its own memory, so the queue is not new unbounded allocation — it tracks work the process had already accepted. What the queue genuinely lacked was a time bound: because the limiter sits outside the SDK, a request's own 30s timeout only starts once it reaches the front, so a task could wait indefinitely, long past the point its caller gave up, and then still go out.

Queued tasks now get a 30s deadline of their own, matching the SDK's request timeout and comfortably above the worst realistic queue (serialised moves to 100 distinct destinations is roughly 15s). The error states the request was never sent, so it is distinguishable from an API failure, and each abandoned task is spliced out of the queue before rejecting — a slot handed to a task that has given up would never be released and would shrink the pool permanently. That case is covered by a test.

This bounds the queue in the dimension that was actually unbounded while keeping `update-tasks` semantics unchanged; an overload error would add a failure mode with no corresponding API error for callers to reason about. Request-level backpressure still belongs at the HTTP layer if we want it.

rfgamaral (commented):
✅

PR #548, opened 2026-08-03
fix(update-tasks): skip redundant moves and batch them by destination

scottlovegrove:
## Context

[Janusz asked on Comms](https://comms.todoist.com/69/ch/AK6drPS98yMaYmyWmztS2/t/CdffD1JCLJMYmWYSf3Fi2) why `todoist-mcp` is so keen on moves, and whether they are real moves or a by-product of moves being bundled with updates. My hypothesis on the thread was that agents pass a `projectId` identical to the one the task already sits in, and we fall into the move path anyway. That is confirmed: the move gate only checked whether a container field was *present*, and the tool never read the task's current state, so it could not tell a relocation from an echo.

**Stacked on #550**, which holds the move-decision logic (`move-planner.ts`) as pure functions with its own tests. Merge #550 first — this PR's diff then drops from ~1400 lines to ~1060, since the planner files are shared. GitHub won't let me retarget the base onto #550's branch (the PR is flagged as part of the now-merged #547 stack), so until #550 lands the diff shown here includes those 354 planner lines. This PR is the wiring and the restructure it forces.

Follows #546 (retry jitter) and #547 (concurrency limiter) from the same thread.

## What changed

**Redundant moves are dropped.** `update-tasks` reads the containers of every task that might move — one `GET /tasks?ids=…` for the whole batch — and drops destinations the task already satisfies, using `planMove` from #550.

**Real moves are batched by destination.** Same-destination moves go through `moveTasks`, which the SDK has always exposed and which nothing in `src/` used: 25 tasks bound for one project become one sync request instead of 25 concurrent ones. A lone move keeps the single-task REST endpoint, whose errors already name the task they belong to.

**Per-task failure reporting survives batching.** `performSyncRequest` throws on the first bad `syncStatus` entry and discards the rest of the map, and the per-command uuids never escape the SDK, so there is no per-command status to map back. Since sync commands apply independently server-side, reporting the whole group as failed would misreport tasks that did move. Instead unresolved groups are collected and reconciled from a single read, letting current state say which commands landed. Zero extra requests on the happy path, one extra read on the error path, no extra writes ever. The trade-off is that failures within a group share the batch's error message — acceptable, since the dominant real error (cross-workspace 403) is uniform across a same-destination group.

**Unknown state always falls through to performing the move,** so a failed or unfiltered read degrades to today's behaviour rather than silently dropping a relocation.

Also: `"inbox"` is resolved once per batch rather than once per task (was up to 25 `getUser` calls), and a failed inbox lookup now fails only the tasks that needed it.

## Why `update-tasks.ts` is a restructure rather than an addition

Most of the diff in that file is churn, not new surface. The old code was a single `processTaskUpdate` doing everything for one task at a time. Prefetching state for a whole batch and grouping moves by destination cannot be expressed per task, so the tool became phased: prepare → prefetch → plan → move → update → collect. Splitting that into smaller PRs would mean landing dead code or a half-wired path, so it is one change; the three commits are the intended review order.

## Response shape

`appliedOperations` gains `redundantMovesSkipped`, and the text response reads `Updated 1 task (1 already in requested destination)` — the feedback channel that should teach callers to stop echoing. The schema descriptions for `projectId`/`sectionId`/`parentId` now say to omit unchanged containers and spell out the detach behaviour, and `src/mcp-server.ts` says the same. Token footprint 37,450 → 37,580 against the 40k soft cap.

The third commit also corrects the `ConcurrencyLimits` documentation from #547: those limits are per process, so a deployment running N instances allows up to N× per account, and a rolling release temporarily doubles the instance count. The service (`todoist-ai-integrations`) runs 2 ECS tasks, autoscaling to 8, with a rolling deploy — so the real per-account ceiling is 2–8 concurrent moves, not 1. A true global ceiling would need shared state, which this server does not have.

## Testing

Verified end-to-end against a real account on two throwaway projects, logging every HTTP request (both projects deleted afterwards, 0 leftovers):

| Scenario | Requests issued |
|---|---|
| Echo current `projectId` + rename | `GET /tasks?ids=…`, `POST /tasks/{id}` — zero moves, `redundantMovesSkipped: 1` |
| Genuine bulk move, 3 tasks → one project | `GET /tasks?ids=…`, then **one** `POST /sync` (previously 3 concurrent `POST /tasks/{id}/move`) |
| Same move re-issued | `GET /tasks?ids=…` only — zero writes, `redundantMovesSkipped: 3` |

`update-tasks.test.ts` gains `describe('redundant moves')` and `describe('batched moves')`: echo skips the move, echo-in-a-section still moves, same-destination tasks produce one `moveTasks` call, a rejected batch reports only the tasks that didn't land, a failed reconcile fails the group, a response omitting a task is checked rather than assumed, a task whose move failed does not get its field update applied, the inbox resolves once, three failed groups still cost one reconcile read, and a repeated id keeps its outcomes separate.

`npm test` (1072), `npm run type-check`, `npm run check` all pass.

## Open question for the API side

Whether `item_move` with `project_id` really does detach section and parent. The predicate is safe either way — it can only under-skip, never wrongly skip — but a definitive answer lets us relax the extra conditions and skip more. Separately, `moveTasks` sends no `syncToken` and hardcodes `resourceTypes: ['items']`, so response size is server-determined and worth measuring on a large account.

## Out of scope

`reorder-objects` already collapses everything into one `client.sync()` call and makes zero concurrent task moves; `project-move` handles one project on a different resource, and no-op detection there would cost a read to save a harmless write.


src/tools/update-tasks.ts
```diff
@@ -204,23 +326,302 @@ const updateTasks = {
...
+ * they target different places.
+ */
+async function applyMoves({
+    client,
+    moving,
+}: {
+    client: TodoistApi
+    moving: PlannedMove[]
+}): Promise<Map<string, MoveResult>> {
+    const results = new Map<string, MoveResult>()
```
doistbot: <a href="#">P2</a>  Preserve results per input entry or reject duplicate task IDs. The schema allows the same ID more than once, but this map overwrites an earlier move's result when that ID appears in another destination group. For example, if moving `x` to project A succeeds and its later move to project B fails, both entries read B's error here, so the successful A move is reported as failed and its field update is skipped. Key results by the prepared item/index, or add uniqueness validation for `tasks[].id`.

src/tools/update-tasks.ts
```diff
@@ -204,23 +326,302 @@ const updateTasks = {
...
+            if (soleId !== undefined) {
+                results.set(soleId, { error: formatBatchItemError(error) })
+                continue
+            }
+
+            // A batched move is one request carrying many commands, and the API applies
+            // them independently while reporting only the first problem. Treating the
+            // whole group as failed would misreport tasks that did move, so read back
+            // the group and let the server say which commands landed.
+            await reconcileMoves({ client, ids, destination, results, error })
```
doistbot: <a href="#">P2</a>  This performs and awaits a separate `getTasks` reconciliation for each failed multi-task destination group. A valid 25-task request split into pairs can therefore make 12 sequential read requests when the moves fail, even though all affected IDs fit in the existing single 25-item query. Accumulate failed/omitted IDs and their destinations, then reconcile disjoint groups from one read after the move loop (preserving per-group destination checks); this avoids an error-path N+1.

doistbot (commented):
This PR makes `update-tasks` smarter about moves: it prefetches task state to skip redundant moves, batches real moves by destination, and reconciles failures with a single read. Two correctness/efficiency issues around duplicate IDs and error-path N+1 queries are worth tightening.

<details>

- **<a href="#">P3</a> src/utils/move-planner.ts:145**: This repeats the exactly-one-destination validation and error already implemented by `createMoveTaskArgs` in `tool-helpers.ts`; this PR also removes its only production caller. Consolidate the validation in one shared primitive (for example, have the existing helper delegate to a planner export, or use it after redundant destinations are filtered) so the two paths cannot drift.
- **<a href="#">P3</a> src/tools/update-tasks.ts:401**: `Math.max(ids.length, MAX_TASKS_PER_OPERATION)` always evaluates to `MAX_TASKS_PER_OPERATION` (25) because `ids` can never exceed the batch cap enforced by the schema. The expression reads as a dynamic guard that never fires; `limit: MAX_TASKS_PER_OPERATION` would be clearer.
- **<a href="#">P3</a> src/utils/move-planner.ts:55**: `matchesCurrentContainer` is a private function used exactly once — inside `planMove` at line 136. Its logic is a simpler variant of `isMoveRedundant` that drops the side-effect checks. Inlining it into `planMove` would reduce a layer of indirection without losing clarity, since the asymmetry between the two predicates is already explained in `planMove`'s JSDoc.
- **<a href="#">P3</a> src/tools/update-tasks.test.ts:1568**: This test only asserts `moveTask` was called, while its sibling ('still moves a task whose project matches but which sits in a section') also checks `structuredContent.appliedOperations.redundantMovesSkipped` is 0. A regression that incorrectly set `redundantMovesSkipped` to 1 would pass here. Consider adding the same `redundantMovesSkipped` assertion for consistency.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F548&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aec93c630-8f4c-11f1-9aff-834b9cd949b1%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aec93c630-8f4c-11f1-9aff-834b9cd949b1&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
scottlovegrove: Done in 4db945a — results are now keyed by the prepared item's index rather than the task id.

Reproduced your exact scenario first. With `x` to project A succeeding and `x` to project B failing, both entries reported B's error and `updatedTaskIds` was empty, so the successful A move was counted as a failure. That is also a regression against the previous code, which processed each entry independently.

Keying by index rather than rejecting duplicates keeps that independence: both moves still execute and each reports its own outcome. Locked in with a test asserting one success and one failure for the repeated id. The prefetch id list is deduplicated too, since a repeated id would otherwise be sent twice in the same query.

src/tools/update-tasks.ts
scottlovegrove: Done in 4db945a — unresolved groups are collected into an `UnresolvedMoveGroup[]` during the move loop and reconciled from one read afterwards, each group still checked against its own destination.

This covers both paths into reconciliation: a group whose request failed, and a group the response omitted. New test moves six tasks across three destinations with every batch failing, and asserts `getTasks` is called exactly twice — one prefetch plus one reconcile — instead of once per failed group.

scottlovegrove:
@doistbot /review

src/tools/update-tasks.ts
```diff
@@ -204,23 +341,313 @@ const updateTasks = {
...
+        } else {
+            groups.set(key, [planned])
+        }
+    }
+
+    const moveLimiter = getMoveLimiter(client)
+    const unresolved: UnresolvedMoveGroup[] = []
+
+    for (const group of groups.values()) {
+        const ids = group.map(({ item }) => item.id)
```
doistbot: <a href="#">P2</a>  Do not send repeated task IDs through `moveTasks`. The schema permits duplicates, but a group such as `[{ id: "x", projectId: "A" }, { id: "x", projectId: "A" }]` produces two sync commands for `x`. If one lands and the other fails, the batch throws and reconciliation sees only that `x` is now in A, so lines 579–583 mark both input entries successful. This regresses the per-entry independence preserved by index-keyed results. Detect IDs repeated in the input and execute those entries with the single-task endpoint (or otherwise avoid batching them).

src/tools/update-tasks.ts
```diff
@@ -149,38 +206,117 @@ const updateTasks = {
...
+        }
 
-                updatedTasks.push(result.value.task)
-                if (result.value.kind === 'partial') {
-                    failures.push({
-                        item: tasks[index]?.id ?? `Task ${index + 1}`,
-                        error: result.value.error,
-                        code: 'PARTIAL_MOVE_APPLIED',
-                    })
+        const movedTasks = await applyMoves({ client, moving })
```
doistbot: <a href="#">P2</a>  This await makes all field updates wait for every move group (and any reconciliation) to finish. Since moves are intentionally serialized, a mixed 25-task batch with many distinct destinations delays even a field-only update until all move requests complete; previously updates could use the separate write limiter while later moves ran. Start field-only updates once planning completes, and enqueue a moved item's field update as soon as that item's move outcome is known, while preserving the per-item move-before-update dependency.

doistbot (commented):
`update-tasks` now prefetches task state to skip redundant moves and batches same-destination moves into single sync requests, reducing redundant API calls and improving efficiency.

<details>

- **<a href="#">P3</a> src/tools/update-tasks.ts:416**: `Math.max(ids.length, MAX_TASKS_PER_OPERATION)` always evaluates to `MAX_TASKS_PER_OPERATION` (25). The schema caps `tasks` at 25 and `ids` is a deduplicated subset of those, so `ids.length` can never exceed the cap. The expression reads as a dynamic guard that never fires; `limit: MAX_TASKS_PER_OPERATION` would be clearer.
- **<a href="#">P3</a> src/tools/update-tasks.test.ts:1575**: This test only asserts `moveTask` was called, while its sibling ('still moves a task whose project matches but which sits in a section') also asserts `redundantMovesSkipped` is 0. A regression that incorrectly set `redundantMovesSkipped` to 1 would pass here. Adding the same assertion would close the gap.
- **<a href="#">P3</a> src/tools/update-tasks.ts:643**: `prepareTaskUpdate` no longer reads `client`: inbox resolution now happens before this phase. Remove this parameter and the corresponding argument at the call site so the function's dependencies reflect its actual work.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F548&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A99059cb0-8f55-11f1-8a81-382c1dbf640e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A99059cb0-8f55-11f1-8a81-382c1dbf640e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/update-tasks.ts
scottlovegrove: Done in defa3ae, by deduplicating the ids sent to `moveTasks` rather than routing repeated entries through the single-task endpoint.

Confirmed the input first: `[{id:'x',projectId:'A'},{id:'x',projectId:'A'}]` produced `moveTasks(['x','x'], …)` — two `item_move` commands for one task in a single sync request, which is the tree contention this batching exists to avoid, self-inflicted in the code meant to prevent it. Now one command per task, with the returned task mapped back to every entry that asked for it.

On the reporting: for a repeated id sharing a destination I think marking both entries successful is correct rather than a regression. Both asked for x to be in A, and x is in A; if the batch fails for a real reason, reconciliation finds x not in A and fails both. The independence that matters is across *different* destinations, which is what index-keyed results preserve. Deduplicating avoids the wasted command and the self-contention without needing two serialised single-task requests. Test added for the dedupe and for both entries being satisfied.

src/tools/update-tasks.ts
scottlovegrove: Done in defa3ae — field updates are now queued as their dependency clears rather than after the whole move phase.

Measured it before changing anything. Three moves to distinct destinations plus one field-only task gave:

    move-start:m1 | move-end:m1 | move-start:m2 | move-end:m2 | move-start:m3 | move-end:m3 | update:field-only

and now:

    update:field-only | move-start:m1 | move-end:m1 | move-start:m2 | move-end:m2 | move-start:m3 | move-end:m3

Took the full pipelining rather than just unblocking field-only tasks, since it was little extra: `applyMoves` calls back per group as soon as that group's outcome is known, so a moved task's update overlaps the next group's move. Three tasks each with a move and a content change now interleave `move-end:m1 | update:m1 | move-start:m2`. Per-task move-before-update is unchanged, and a failed move still suppresses its update.

Entries that need reconciliation are queued after the single reconcile read — that is the error path, where waiting is fine and a shared read is worth more than earlier starts. Two tests added, asserting the field-only ordering and the per-item overlap.

rmartins90:
✅

PR #550, opened 2026-08-03
feat(move-planner): decide when a requested container change is a real move

scottlovegrove:
## Context

Split out of #548 to keep that PR reviewable. This is the decision logic on its own — pure functions, no API involved, no callers yet. The wiring into `update-tasks` is #548, which is stacked on this branch.

[Janusz reported on Comms](https://comms.todoist.com/69/ch/AK6drPS98yMaYmyWmztS2/t/CdffD1JCLJMYmWYSf3Fi2) that by volume almost all failing task moves the API sees come from `todoist-mcp`. The cause is that `update-tasks` only checks whether a container field is *present*, so a caller echoing back a task's existing `projectId` alongside a field edit gets a real move write. Telling those apart needs the task's current state plus a rule about what a move actually does to a task's position — and that rule doesn't belong tangled up in the tool's request handling.

## What this adds

- **`isMoveRedundant(current, destination)`** — whether moving the task would leave it exactly where it already is. Matching the destination field alone isn't sufficient: moving a task to a project also lifts it out of its section and out from under its parent, and moving it to a section lifts it out from under its parent. So a task whose `projectId` matches but which sits in a section is *not* already in the requested end state.
- **`planMove({ taskId, request, current })`** — reduces the containers a caller sent to the single move actually needed, or none. Throws the existing "only one of projectId, sectionId, or parentId" error if more than one *genuine* destination remains.
- **`destinationKey(move)`** — identifies a destination so tasks heading to the same place can be collapsed into one request later.

## The rule worth reviewing

`planMove` is deliberately asymmetric:

- **One container** → judged on the whole position it describes. Asking for a task's current project still moves it, because that would lift it out of its section, which is a real change the caller may well want.
- **Several containers** → judged field by field. A task described as already being in its project *and* its section is left alone rather than pulled out of that section.

My first version applied the single-container rule uniformly, which would have silently detached tasks from their sections whenever a caller echoed a whole task object. The asymmetry is what prevents that.

It also turns a pure-loss case into a working one: `{ projectId: <current>, sectionId: <current> }` previously failed outright as "only one of…", so a full echo was a guaranteed failed call. Some of the error volume Janusz is seeing is likely exactly this.

## Residual uncertainty

The detaching behaviour these rules assume is not verifiable from this repo — the SDK just forwards `project_id`. The predicate is therefore conservative: it only reports "redundant" when the task already looks the way the move would leave it under the detaching reading. If the API turned out *not* to detach, we merely under-skip; we never wrongly skip. That reasoning is in the JSDoc so it doesn't get "simplified" later. A definitive answer from the API side would let us relax the extra conditions and skip more.

## Testing

`move-planner.test.ts` covers the redundancy matrix — three destination kinds × section/parent present or absent — plus the multi-container reduction, the full-echo cases, the ≥2-real-destinations throw, and unknown current state always moving. 26 tests.

`npm test` (1050), `npm run type-check`, `npm run check` all pass.

## Note for review

Nothing imports these functions on this branch, so they are unused until #548 lands. That is the cost of splitting; the alternative was one ~1400-line PR.

### DIFF
--- a/src/utils/move-planner.test.ts
+++ b/src/utils/move-planner.test.ts
@@ -1,0 +1,200 @@
+import { destinationKey, isMoveRedundant, planMove } from './move-planner.js'
+import { createMockTask } from './test-helpers.js'
+
+const PROJECT = 'project-1'
+const SECTION = 'section-1'
+const PARENT = 'parent-1'
+
+describe('isMoveRedundant', () => {
+    describe('a project destination', () => {
+        it('is redundant when the task already sits at the project root', () => {
+            const task = createMockTask({ projectId: PROJECT, sectionId: null, parentId: null })
+            expect(isMoveRedundant(task, { projectId: PROJECT })).toBe(true)
+        })
+
+        it('is not redundant for a different project', () => {
+            const task = createMockTask({ projectId: 'other', sectionId: null, parentId: null })
+            expect(isMoveRedundant(task, { projectId: PROJECT })).toBe(false)
+        })
+
+        it('is not redundant when the task is in a section, since the move lifts it out', () => {
+            const task = createMockTask({ projectId: PROJECT, sectionId: SECTION, parentId: null })
+            expect(isMoveRedundant(task, { projectId: PROJECT })).toBe(false)
+        })
+
+        it('is not redundant when the task has a parent, since the move lifts it out', () => {
+            const task = createMockTask({ projectId: PROJECT, sectionId: null, parentId: PARENT })
+            expect(isMoveRedundant(task, { projectId: PROJECT })).toBe(false)
+        })
+    })
+
+    describe('a section destination', () => {
+        it('is redundant when the task already sits directly in that section', () => {
+            const task = createMockTask({ sectionId: SECTION, parentId: null })
+            expect(isMoveRedundant(task, { sectionId: SECTION })).toBe(true)
+        })
+
+        it('is not redundant for a different section', () => {
+            const task = createMockTask({ sectionId: 'other', parentId: null })
+            expect(isMoveRedundant(task, { sectionId: SECTION })).toBe(false)
+        })
+
+        it('is not redundant when the task has no section', () => {
+            const task = createMockTask({ sectionId: null, parentId: null })
+            expect(isMoveRedundant(task, { sectionId: SECTION })).toBe(false)
+        })
+
+        it('is not redundant when the task has a parent, since the move lifts it out', () => {
+            const task = createMockTask({ sectionId: SECTION, parentId: PARENT })
+            expect(isMoveRedundant(task, { sectionId: SECTION })).toBe(false)
+        })
+    })
+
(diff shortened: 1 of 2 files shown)


doistbot (commented):
This PR adds pure decision functions — `isMoveRedundant`, `planMove`, and `destinationKey` — that determine whether a requested container change is an actual move or a no-op, laying the groundwork for fixing redundant move writes in `update-tasks`. No inline issues were flagged.

<details>

- **<a href="#">P3</a> src/utils/move-planner.ts:145**: This reimplements `createMoveTaskArgs` from `src/tool-helpers.ts`, including the same one-destination invariant and error text. After reducing `effective`, pass its project/section/parent values to that helper (or extract a lower-level shared helper) so move validation and its error contract have one source of truth.
- **<a href="#">P3</a> src/utils/move-planner.test.ts:126**: Missing coverage for the symmetric multi-container case: an echoed section alongside a real project change. The existing test at L126 covers `{ projectId: PROJECT, sectionId: SECTION }` where the project matches current but the section is new (→ section move). The reverse scenario — `{ projectId: 'elsewhere', sectionId: SECTION }` where the section matches current but the project is new (→ project move) — exercises a distinct path through `matchesCurrentContainer` that has no test. This is a realistic echo pattern (caller provides both fields, one happens to match) and would silently break if `matchesCurrentContainer` regressed for project matching.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F550&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A173e9170-8f53-11f1-933c-b448b3f80222%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A173e9170-8f53-11f1-933c-b448b3f80222&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

pedroalves0 (commented):
✅

PR #551, opened 2026-08-03
test(request-profile): assert how many requests a batch makes

scottlovegrove:
## Context

Stacked on #548. Retarget to `main` once that merges.

Reviewing the test suite after this work raised a fair question: were the tests added in #548 things we should have had already? Mostly no — they cover behaviour that didn't exist before. But one dimension was genuinely missing, and it is the one that let the original problem live here unnoticed.

Before #548, `update-tasks` had 48 tests with **8 request-count assertions between them**, and the only `getUser` count assertion was incidental — it checked that a retry made two attempts. So a 25-task inbox batch issuing **25 `getUser` calls**, and 25 concurrent moves, passed the suite happily for as long as that code existed. The tests verified what a call *returned*; nothing verified the shape of the traffic. That is precisely the dimension [Janusz's report](https://comms.todoist.com/69/ch/AK6drPS98yMaYmyWmztS2/t/CdffD1JCLJMYmWYSf3Fi2) lives in, and it was a backend dashboard that noticed rather than CI.

It isn't only that file. `add-tasks` had 4 count assertions across 34 tests, `update-projects` 3 across 18, `reschedule-tasks` 2 across 12.

## The helper

`src/utils/request-profile.ts`, test-only:

```ts
const { client, profile } = createProfilingClient({
    getTasks: () => ({ results: [], nextCursor: null }),
    moveTasks: (ids: string[]) => ids.map((id) => createMockTask({ id })),
})

await updateTasks.execute({ tasks: twentyFiveTasks }, client)

expectRequestProfile(profile, {
    getTasks: { count: 1, peak: 1 },
    moveTasks: { count: 1, peak: 1 },
})
```

- **`createProfilingClient(handlers)`** — a mock client whose requests are recorded. Handlers return a value (or throw); the profile exposes per-method `count` and `peak` concurrency, the `sequence` requests started in, and `argsFor(method)`.
- **`expectRequestProfile(profile, expected)`** — asserts the *whole* profile. A method absent from `expected` must not have been called, so a newly introduced request fails the test rather than slipping in unnoticed. Failures print the entire profile, which is far more use than "expected 1, got 2" on one method.

One implementation detail worth knowing: handlers resolve a macrotask later, deliberately. A handler that resolved immediately would let each caller finish before the next began, so every `peak` would read as 1 no matter how the tool actually behaved.

## It found two read amplifications immediately

**`add-tasks` checked the archived-project status once per task** (`add-tasks.ts:204`) — 25 identical reads for a batch bound for one project. Fixed here by caching the lookup for the duration of the batch, next to the section cache that already existed in the same function. The *promise* is cached rather than the result, so tasks racing for the same project share one request instead of each starting their own.

**`reschedule-tasks` makes two reads per task, all in flight at once**, while its writes are already collapsed into a single sync request — 10 tasks cost 20 concurrent `getTask` calls. Not fixed here: switching it to a batched read is a change in its own right and needs the comma-separated `ids` workaround from #548. The cost is now asserted with a comment saying so, so the test fails deliberately when that lands rather than the improvement going unnoticed.

## Coverage added

`update-tasks` — a field-only batch costs one request per task and nothing else; 25 tasks moving to one destination cost two requests in total; the inbox resolves once however many tasks target it; moves never overlap regardless of destination count; a batch of entirely redundant moves writes nothing at all.

`add-tasks` — a shared destination project is checked once, distinct projects once each, and no project means no reads. (That last test caught my own wrong assumption: same-destination tasks are created sequentially to preserve sibling order, so peak is 1, not N.)

`reschedule-tasks` — the characterisation described above.

Plus 12 tests for the helper itself, including that `expectRequestProfile` fails on an unexpected request, a wrong count, and excess concurrency — the failure modes it exists to produce.

## Notes

- `CODEBASE.md` gains the helper in the utils catalogue and a line under Testing conventions, since the point is for the next batch tool to use it rather than rediscover the gap.
- `manage-assignments.ts` has no test file of its own (only `assignment-integration.test.ts`) and is another unbounded fan-out site — `Promise.all` over up to 50 tasks plus a `getTask` prefetch each. Out of scope here, but it is the next obvious candidate.

`npm test` (1096), `npm run type-check`, `npm run check` all pass.

### DIFF
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -2,7 +2,7 @@ import type { AddTaskArgs, Task, TodoistApi } from '@doist/todoist-sdk'
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { formatBatchItemError } from '../tool-execution-error.js'
-import { isInboxProjectId, mapTask } from '../tool-helpers.js'
+import { isInboxProjectId, mapTask, type Project } from '../tool-helpers.js'
 import { assignmentValidator } from '../utils/assignment-validator.js'
 import { BatchLimits } from '../utils/constants.js'
 import { DurationParseError, parseDuration } from '../utils/duration-parser.js'
@@ -101,6 +101,9 @@ const addTasks = {
 
         // Resolve each section's project at most once across the whole batch
         const sectionProjectCache = new Map<string, string>()
+        // Same for the archived check: a batch bound for one project asked the same
+        // question once per task, which is 25 identical reads at the batch cap.
+        const projectCache = new Map<string, Promise<Project>>()
 
         // Process groups in parallel; within each group, process sequentially
         type IndexedResult = { index: number; result: PromiseSettledResult<Task> }
@@ -109,7 +112,10 @@ const addTasks = {
                 const results: IndexedResult[] = []
                 for (const { task, index } of group) {
                     try {
-                        const created = await processTask(task, client, sectionProjectCache)
+                        const created = await processTask(task, client, {
+                            sectionProjectCache,
+                            projectCache,
+                        })
                         results.push({ index, result: { status: 'fulfilled', value: created } })
                     } catch (error) {
                         results.push({
@@ -181,8 +187,12 @@ function destinationKey(task: z.infer<typeof TaskSchema>): string {
 async function processTask(
     task: z.infer<typeof TaskSchema>,
     client: TodoistApi,
-    sectionProjectCache: Map<string, string>,
+    caches: {
+        sectionProjectCache: Map<string, string>
+        projectCache: Map<string, Promise<Project>>
+    },
 ): Promise<Task> {
+    const { sectionProjectCache, projectCache } = caches
     const {
         duration: durationStr,
         projectId,
@@ -201,7 +211,14 @@ async function processTask(
 
     // Validate project is not archived
     if (resolvedProjectId) {
-        const project = await client.getProject(resolvedProjectId)
+        // Cached as the promise, not the result, so tasks racing for the same project
+        // share one request rather than each starting their own.
+        let pending = projectCache.get(resolvedProjectId)
+        if (!pending) {
+            pending = client.getProject(resolvedProjectId)
+            projectCache.set(resolvedProjectId, pending)
+        }
+        const project = await pending
         if (project.isArchived) {
             throw new Error(
                 `Task "${task.content}": Cannot create task in archived project "${project.name}"`,
--- a/src/utils/request-profile.ts
+++ b/src/utils/request-profile.ts
@@ -1,0 +1,155 @@
+import type { TodoistApi } from '@doist/todoist-sdk'
+import { vi } from 'vitest'
+
+/**
+ * Test-only. Records the API requests a tool makes, so a batch's request behaviour can
+ * be asserted rather than inferred.
+ *
+ * Existing tool tests check what a call *returns*. That leaves the shape of the traffic
+ * unverified, which is how this server ended up issuing one `getUser` per task in a
+ * batch and firing every move in a batch at once: both looked correct per task, and no
+ * test could tell. A backend engineer's dashboard noticed instead of CI.
+ *
+ * Two properties are worth pinning for any batch tool:
+ *
+ * - **how many requests** a batch of N items makes. Anything that scales with N when it
+ *   needn't is the fan-out this exists to catch.
+ * - **how many run at once**, since concurrent writes to the same account contend
+ *   server-side.
+ */
+
+type Handlers = {
+    [K in keyof TodoistApi]?: TodoistApi[K] extends (...args: infer A) => unknown
+        ? (...args: A) => unknown
+        : never
+}
+
+type MethodProfile = {
+    /** Requests made to this method. */
+    count: number
+    /** The most that were ever in flight at once. */
+    peak: number
+}
+
+type RequestProfile = {
+    /** Per-method counts and peak concurrency, for methods that were called. */
+    byMethod: Record<string, MethodProfile>
+    /** Total requests across every method. */
+    total: number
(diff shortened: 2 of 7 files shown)


src/utils/request-profile.ts
```diff
@@ -0,0 +1,155 @@
...
+    const client = {} as Record<string, unknown>
+    for (const [method, handler] of Object.entries(handlers)) {
+        client[method] = vi.fn(async (...args: unknown[]) => {
+            record(method)
+            argsByMethod.set(method, [...(argsByMethod.get(method) ?? []), args])
+            try {
+                // Yield so concurrent callers all reach the handler before any returns,
+                // which is what makes `peak` meaningful.
+                await new Promise((resolve) => setImmediate(resolve))
+                return (handler as (...handlerArgs: unknown[]) => unknown)(...args)
```
doistbot, line 90: <a href="#">P2</a>  Await async handlers before clearing `inFlight`. Handlers are typed to return `unknown`, so an async handler is valid; `return handler(...)` runs the `finally` as soon as it receives that promise, while the client call remains pending through async return adoption. A staggered second request then undercounts peak concurrency. Use `return await handler(...)` and cover an async handler in the helper tests.

src/tools/add-tasks.ts
```diff
@@ -201,7 +211,14 @@ async function processTask(
 
     // Validate project is not archived
     if (resolvedProjectId) {
-        const project = await client.getProject(resolvedProjectId)
+        // Cached as the promise, not the result, so tasks racing for the same project
+        // share one request rather than each starting their own.
+        let pending = projectCache.get(resolvedProjectId)
+        if (!pending) {
+            pending = client.getProject(resolvedProjectId)
+            projectCache.set(resolvedProjectId, pending)
```
doistbot, line 219: <a href="#">P2</a>  Evict a rejected lookup from `projectCache`. This stores a rejected promise for the rest of the batch, so if the first `getProject` call fails transiently, every later task targeting that project fails immediately without another attempt. This is especially visible in a mixed-destination batch, where successful tasks prevent the outer tool retry. Keep sharing in-flight requests, but delete this entry when its promise rejects.

doistbot (commented):
This PR introduces a test-only request-profiling helper to assert the count and peak concurrency of API requests made by batch tools, along with a fix for a per-task read-amplification bug in `add-tasks` where the archived-project check ran once per task instead of once per project.

Few things worth tightening:

- Await async handlers before clearing `inFlight` — using `return await handler(...)` instead of `return handler(...)` ensures the in-flight count stays accurate through promise settlement, and the helper tests should cover an async handler case.
- Evict rejected lookups from `projectCache` — a rejected promise cached for the rest of the batch means a transient `getProject` failure dooms every later task targeting that project without retry; delete the entry on rejection while still sharing in-flight requests.

<details>

- **<a href="#">P3</a> src/tools/add-tasks.test.ts:1201**: The "checks each distinct project once" test uses a targeted `profile.count` assertion rather than `expectRequestProfile`, and doesn't assert peak concurrency. Since the two projects are in different destination groups processed in parallel via `Promise.all`, `getProject` peak should be 2. The other two tests in this `describe('request profile')` block use `expectRequestProfile` which catches both count and peak — this one is the odd one out.
- **<a href="#">P3</a> src/utils/request-profile.ts:151**: Two unreachable fallbacks in the normalization step. `spec?.count ?? 0` — the `count` property in `expected`'s type is required (`{ count: number; peak?: number }`), so `spec.count` is always a `number`. The optional chaining and `?? 0` can never trigger. `actual[method]?.peak ?? 0` — the key-set comparison on line 146 (`toEqual(expectedMethods.toSorted())`) guarantees that every method in `expected` also exists in `actual`, so `actual[method]` is always defined and its `.peak` is always a number. Simplify to `{ count: spec.count, peak: spec.peak ?? actual[method].peak }`.
- **<a href="#">P3</a> src/utils/request-profile.test.ts:140**: This test only checks that the profiling client satisfies TypeScript's structural `TodoistApi` type. That contract is already exercised — and actually verified — by every tool-level test in this same PR that passes `client` to `execute()`. The test provides no independent regression protection and is pure implementation trivia.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F551&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0ba2e710-8f5a-11f1-9234-e4c3a03dae0e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0ba2e710-8f5a-11f1-9234-e4c3a03dae0e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove:
Closing — this went the wrong way. The ask was to find shared code that would reduce the size of #548, and this added 518 lines instead. Pursuing the extraction inside #548 itself.

Two findings from it worth keeping a note of, each small enough to stand alone later: `add-tasks` runs the archived-project check once per task (25 identical reads for a batch bound for one project), and `reschedule-tasks` makes two reads per task with all of them in flight at once.

PR #552, opened 2026-08-03
fix: add due string recovery guidance

gnapse:
# Pull Request

Closes Doist/Issues#20569

## Short description

Adds narrowly scoped recovery guidance when Todoist rejects a supplied dueString with the current Invalid date format response. The MCP response identifies dueString and, for the malformed recurring prefix, tells the caller to retry with every year on July 1 without silently rewriting the request.

The batch error path now preserves this recovery text so it reaches MCP callers for both partial and all-failed batches. The add-tasks schema and MCP tool guidance document the valid yearly recurrence form.

### Validation

- Reproduced the issue first
    - Ran the malformed request through a local MCP server against the configured German account on an isolated HEAD worktree. Baseline returned only HTTP 400, code 42, tag BAD_REQUEST, and generic parameter-format retry guidance.
- Then we confirmed it is mitigated by the changes
    - Ran the same request through the current local MCP server. It returned the explicit dueString correction and retry instruction.
- Verified the English control every year on July 1 creates a yearly task due 2027-07-01 on the German account.
- Verified the German input jeden 1. Juli also succeeds without dueLang on that account.
- Deleted the temporary Inbox tasks and confirmed a final MCP #20569 search returned zero tasks.
- Ran npm test (1085 passed), npm run type-check, npm run check, and git diff --check.

## PR Checklist

- [x] Updated tool schema and MCP guidance
- [ ] New tools added to getMcpServer AND exported in src/index.ts.


src/tool-execution-error.ts
```diff
@@ -529,6 +557,45 @@ export function formatBatchItemError(error: unknown): string {
...
+    { taskContent, dueString }: { taskContent: string; dueString: string | undefined },
+): string | undefined {
+    if (!dueString) {
+        return undefined
+    }
+
+    const parsedApiError = extractApiErrorInfo(error)
+    const tag = parsedApiError?.tag?.toUpperCase()
+    const isStableDueStringTag = tag === 'INVALID_DUE_STRING'
+    const isCurrentGenericDueFormatError = /^invalid date format\.?$/i.test(
```
doistbot: <a href="#">P2</a>  The generic `Invalid date format` fallback only checks whether `dueString` was supplied, but `add-tasks` also sends the independently user-controlled `deadlineDate`. A request with a valid `dueString` (for example, `tomorrow`) and an invalid `deadlineDate` can therefore be told to change *only* `dueString`. Pass enough context to skip this generic fallback when another date field could be responsible, or reserve the attribution for the stable due-string tag/field hint.

src/tool-execution-error.ts
```diff
@@ -506,6 +506,29 @@ function formatGenericError(error: unknown): string {
...
+    const taskStart = error.message.lastIndexOf('Task "', markerIndex)
+    const retryEnd = error.message.indexOf('retry.', markerIndex)
+    if (taskStart === -1 || retryEnd === -1) {
+        return undefined
+    }
+
+    // A batch that entirely fails wraps its per-task error in a new Error.
+    // Preserve this known recovery message verbatim rather than truncating it
+    // through the generic-error formatter before MCP can return it to the caller.
+    return error.message.slice(taskStart, retryEnd + 'retry.'.length)
```
doistbot: <a href="#">P2</a>  `extractDueStringRecovery` extracts only the first recovery message from the aggregated all-failed error. When multiple tasks fail with dueString errors, `addTasks.execute` throws `All N task(s) failed to create: "item1": <recovery1>; "item2": <recovery2>`. The `lastIndexOf('Task "', markerIndex)` + `indexOf('retry.', markerIndex)` logic finds the first `Task "`…`retry.` span, so `formatToolExecutionError` returns only `recovery1`. The failure count ("All 2 task(s)") and every subsequent task's failure are silently dropped from the MCP response. The test at `add-tasks.test.ts` line 698 uses `.rejects.toThrow(substring)` which passes on the first task's prefix alone, so it doesn't catch this. Consider either skipping extraction when the message contains multiple recovery markers (fall through to the generic formatter which preserves the full aggregated message), or extracting all matches and joining them.

src/tool-execution-error.test.ts
```diff
@@ -247,6 +251,57 @@ describe('formatToolExecutionError', () => {
...
+            expect(formatBatchItemError(new Error(recovery))).toBe(recovery)
+        })
+
+        it('does not attribute unrelated generic API errors to dueString', () => {
+            const error = Object.assign(new Error('HTTP 400: Bad Request'), {
+                httpStatusCode: 400,
+                responseData: { error: 'Invalid project' },
+            })
+
+            expect(
```
doistbot, line 330: <a href="#">P2</a>  The no-recurring-prefix branch of `formatDueStringParseError` (the final `return` at line 596 of tool-execution-error.ts) produces a distinct, shorter recovery message, but no test exercises it with a matching error. This test pairs a non-recurring `dueString` ('tomorrow') with an unrelated error ('Invalid project'), so it returns `undefined` before reaching that branch. Add a case where the error is 'Invalid date format' and the dueString has no `recurring` prefix to cover the alternate user-visible output.

doistbot (commented):
This PR adds due-string recovery guidance so that when Todoist rejects a malformed `dueString`, the MCP response identifies the field, suggests the corrected form, and instructs retry—propagating that text through both partial and all-failed batch paths.

Few things worth tightening:
- The generic `Invalid date format` fallback triggers whenever `dueString` is supplied, even if another date field like `deadlineDate` is the actual culprit. Consider reserving the attribution to cases where `dueString` is the likely cause or skipping it when other date fields are present.
- `extractDueStringRecovery` only surfaces the first task's recovery message in an all-failed batch, silently dropping subsequent failures and their counts. Either extract all matches or skip extraction when multiple recovery markers are present so the full aggregated error is preserved.
- The no-recurring-prefix branch of `formatDueStringParseError` (returning a shorter recovery message) has no test coverage—add a case pairing a non-recurring `dueString` with an actual `Invalid date format` error.
- When `dueString` is exactly the bare word `"recurring"`, the stripped example is empty, producing a confusing message with empty backtick quotes.

<details>

- **<a href="#">P3</a> src/tool-execution-error.ts:593**: When `dueString` is exactly `"recurring"` (or `"recurring "` with trailing whitespace), `withoutRecurringPrefix` becomes an empty string. The condition `withoutRecurringPrefix !== dueString.trim()` is still true, so the function returns a message with an empty example: `for example \`\`; don't prefix it with \`recurring\`.` An LLM that sends `dueString: "recurring"` would get unhelpful guidance. Consider falling through to the generic message (line 596) when `withoutRecurringPrefix` is empty, or providing a default example like `every day`.
- **<a href="#">P3</a> src/tool-execution-error.ts:509**: `formatDueStringParseError` (line 566) constructs a recovery message and wraps it in `new Error(recovery)`, then `extractDueStringRecovery` (line 509) re-parses that same message back out of the error via fragile `indexOf`/`lastIndexOf` string operations. These two functions form an implicit contract: if the message format changes, the extraction silently breaks. A custom `DueStringParseError` class (following the existing `DurationParseError` pattern) would let `formatBatchItemError` and `formatToolExecutionError` use `instanceof` checks and eliminate the re-parsing.
- **<a href="#">P3</a> src/tool-execution-error.test.ts:283**: Low-value test. `extractDueStringRecovery` uses marker-based slicing (`Task "` ... `retry.`), not length-based truncation, so the long-content assertion doesn't guard against any actual code risk. The behavior is already covered by the integration tests that exercise the full recovery path with realistic inputs. Remove this test to reduce maintenance cost without losing signal.
- **<a href="#">P3</a> src/tools/add-tasks.ts:35**: The schema description for `dueString` now documents recurrence syntax ("every year on July 1", don't prefix with "recurring"), but the matching `add-tasks` entry in `src/mcp-server.ts` (line 74) still only says `dueString (natural language like "tomorrow", "next Friday", "2024-12-25")` with no recurrence guidance. CLAUDE.md's Documentation Requirements state: when adding tool features, update both the schema description **and** `src/mcp-server.ts` tool usage guidelines. Add the recurrence-syntax note to the mcp-server.ts instructions so both channels stay aligned.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F552&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A57156bc0-8f85-11f1-8b12-302b0b82d4e7%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A57156bc0-8f85-11f1-8b12-302b0b82d4e7&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

pedroalves0 (commented):
✅

PR #555, opened 2026-08-04
feat: Add template tools for importing and exporting projects

rainsjm:
Added some template tools, as the Todoist Teams squad is working on them this quarter. With these tools, you can apply a Todoist template to a project, and export a project as a template to reuse or share. 

Details:

- export-project-template: returns a project as CSV content, or as a shareable link with format: "url". Prefer the link for large projects — it doesn't pull the whole file into the response.
- import-project-template: adds a template's tasks, sections and comments to an existing project. Source it with a gallery slug (product-launch), a personal
  template ID (UT_...), a full template URL, or CSV from export-project-template via csvFileContent.
- Paste any Todoist template URL and it resolves to the ID, so gallery links and in-app .../my-templates/UT_.../view links both work without the user having to dig
  out an ID.
- "Start a new project from this template" works by composing add-projects then import-project-template — the create-from-template endpoint only accepts an
  uploaded file, so it can't take a gallery template anyway. The tool description tells the model to do that.
- A 403 on someone else's personal template now explains that templates are private to whoever saved them, and suggests the gallery or a shared CSV, instead of
  surfacing a bare HTTP error.

## Testing

- 25 new unit tests across both tools, including a table-driven pass over the URL/ID forms (gallery URL, in-app `/view` URL, local dev URL, query/hash, trailing slash, bare slug, bare `UT_` ID)
- Verified end to end against the real API: imported a gallery template by URL into a scratch project, exported it back, then deleted the project
- Full suite passes (1118 tests); type-check, oxlint, oxfmt and schema lint clean


src/tools/import-template.ts
```diff
@@ -0,0 +1,136 @@
...
+        .string()
+        .min(1)
+        .describe('The ID of the existing project to import the template into.'),
+    templateId: z
+        .string()
+        .optional()
+        .describe(
+            'The template to import — either a gallery slug ("product-launch"), a personal template ID ("UT_28Ex..."), or a full Todoist template URL, which is reduced to the ID automatically. Gallery templates work for anyone; personal templates only for the account that owns them. There is no way to list templates through this server, so only use an ID or URL the user supplied. Provide either this or `file`, not both.',
+        ),
+    file: z
```
doistbot: <a href="#">P2</a>  `file` accepts the full raw CSV, but `registerTool` logs the complete `args` object whenever execution fails. A caller can import CSV containing private task/comment data and force an API error (for example, with an invalid project ID), causing that content to be written to server logs. Redact or omit `file` from error logging before accepting raw template content.

src/tools/import-project-template.ts
```diff
@@ -0,0 +1,136 @@
...
+        .describe(
+            'Raw CSV template content, as produced by export-template. Provide either this or `templateId`, not both.',
+        ),
+    locale: z
+        .string()
+        .optional()
+        .describe('Locale for the imported content when using `templateId`. Defaults to "en".'),
+}
+
+const OutputSchema = {
```
doistbot, line 86: <a href="#">P2</a>  The output schema returns four unbounded arrays of full entity schemas (`ProjectSchema`, `SectionSchema`, `TaskSchema`, `CommentSchema`), costing ~1760 tokens at the fixed-tool level. Other tools with array outputs enforce explicit limits — `add-tasks` caps at 25, `find-tasks` defaults to 30 with a 200 max. A large project template could produce hundreds of tasks/sections/comments in the structured response, which the LLM doesn't need in full detail after an import. Consider either capping the returned arrays or replacing them with lighter summary schemas (e.g., `sections` already uses `toSectionSummary` — the same approach for tasks and comments would help). The `projects` array is almost always empty since import-targets an existing project, so it could be dropped or made optional.

src/tools/import-template.ts
```diff
@@ -0,0 +1,136 @@
...
+        return trimmed
+    }
+
+    const segments = trimmed
+        .replace(/[?#].*$/, '')
+        .split('/')
+        .filter((segment) => segment.length > 0)
+    const last = segments.at(-1)
+
+    return (last === 'view' ? segments.at(-2) : last) ?? trimmed
```
doistbot: <a href="#">P2</a>  Reject non-Todoist URLs instead of extracting their last path segment. For example, `https://example.com/product-launch` becomes `product-launch`, which is a valid gallery slug and will import that template into the target project despite the user supplying an invalid URL. Validate the Todoist host and supported template URL paths before reducing a URL to an ID, and fail otherwise.

src/tools/import-project-template.test.ts
```diff
@@ -0,0 +1,213 @@
...
+                { projectId: TEST_IDS.PROJECT_TEST, templateId: 'product-launch' },
+                mockTodoistApi,
+            )
+
+            expect(mockTodoistApi.importTemplateFromId).toHaveBeenCalledWith({
+                projectId: TEST_IDS.PROJECT_TEST,
+                templateId: 'product-launch',
+                locale: undefined,
+            })
+            expect(result.structuredContent?.totalCount).toBe(3)
```
doistbot, line 120: <a href="#">P2</a>  This test sets up a section (id=SECTION_1) and two tasks (TASK_1, TASK_2) but only asserts `totalCount` and `textContent` — it never checks that `structuredContent.sections` and `structuredContent.tasks` actually contain the mapped objects. A broken or swapped mapping (e.g. passing `result.sections` raw instead of the mapped array, or dropping `mapTask`) would still produce the right count and pass. The codebase's own pattern in `add-tasks.test.ts` verifies array contents with `expect.arrayContaining([expect.objectContaining({ id: ... })])`; applying the same here would close the gap.

doistbot (commented):
This PR adds two new MCP tools — `export-template` and `import-template` — for exporting projects as CSV/shareable links and importing template tasks, sections, and comments into an existing project from a gallery slug, personal template ID, URL, or raw CSV.

Few things worth tightening:

- **Redact `file` from error logs.** `registerTool` logs the full `args` object on failure, so supplying raw CSV with an invalid project ID would write private template content to server logs. Strip or redact `file` before logging.
- **Reject non-Todoist URLs.** `extractTemplateId` falls back to the last path segment for any URL, so `https://example.com/product-launch` would resolve to the `product-launch` gallery slug and silently import a template from an untrusted source. Validate the host and supported path patterns before extracting an ID.
- **Cap or summarize the output arrays.** All four returned arrays (`projects`, `sections`, `tasks`, `comments`) are unbounded and use full entity schemas, unlike `add-tasks` (max 25) or `find-tasks` (default 30). Consider reusing summary schemas or dropping the `projects` array, which is almost always empty for an import into an existing project.
- **Strengthen the structured-content test.** The import test checks `totalCount` and `textContent` but never verifies that `structuredContent.sections` and `structuredContent.tasks` actually contain the mapped objects — a swapped or dropped mapping would still pass. Use `expect.arrayContaining` / `expect.objectContaining` as done in `add-tasks.test.ts`.

<details>

- **<a href="#">P3</a> src/tools/export-template.ts:50**: `scripts/run-tool.ts` passes raw JSON directly to `execute` without parsing the Zod schema. Running `export-template` with only `projectId` therefore leaves `format` undefined: this takes the file branch but returns structured content without the required `format` field. Default it in `execute` (for example, `format = 'file'`) so the documented default also works through the direct runner.
- **<a href="#">P3</a> src/tools/import-template.ts:19**: `extractTemplateId` hand-rolls URL parsing with `.replace(/[?#].*$/, '')` and manual `.split('/')` to extract path segments. JavaScript's built-in `new URL()` natively separates pathname from query/hash and is already used for path extraction elsewhere in the codebase (`view-attachment.ts` line 56). The same logic can be expressed more robustly: ```typescript function extractTemplateId(input: string): string { const trimmed = input.trim() if (!/^https?:\/\//i.test(trimmed)) return trimmed try { const segments = new URL(trimmed).pathname.replace(/\/$/, '').split('/').filter(Boolean) const last = segments.at(-1) return (last === 'view' ? segments.at(-2) : last) ?? trimmed } catch { return trimmed } } ``` This also gains graceful handling of malformed URLs via try/catch.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F555&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Afd61e100-9036-11f1-869b-0b7343bbfdd6%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Afd61e100-9036-11f1-869b-0b7343bbfdd6&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/export-template.ts
```diff
@@ -0,0 +1,89 @@
+import { z } from 'zod'
+import type { TodoistTool } from '../todoist-tool.js'
+import { ToolNames } from '../utils/tool-names.js'
+
+const PREVIEW_LINE_COUNT = 5
+
+const ArgsSchema = {
+    projectId: z.string().min(1).describe('The ID of the project to export as a template.'),
+    format: z
+        .enum(['file', 'url'])
```
scottlovegrove: Let's extract this array as a constant, it can then be re-used in the outputschema.

src/utils/tool-names.ts
```diff
@@ -72,6 +72,10 @@ export const ToolNames = {
     ADD_FILTERS: 'add-filters',
     UPDATE_FILTERS: 'update-filters',
 
+    // Template tools
+    EXPORT_TEMPLATE: 'export-template',
+    IMPORT_TEMPLATE: 'import-template',
```
scottlovegrove: I wonder if these should be `export-project-template` and `import-project-template`, and this would apply to their respective file names too as well as any other places in which the names are used. 

I know _we_ know that templates are about projects, but having project in the name just makes it easier for an LLM to judge this too, without having to confirm in the description.

src/tools/import-template.ts
```diff
@@ -0,0 +1,162 @@
...
+const ArgsSchema = {
+    projectId: z
+        .string()
+        .min(1)
+        .describe('The ID of the existing project to import the template into.'),
+    templateId: z
+        .string()
+        .optional()
+        .describe(
+            'The template to import — either a gallery slug ("product-launch"), a personal template ID ("UT_28Ex..."), or a full Todoist template URL, which is reduced to the ID automatically. Gallery templates work for anyone; personal templates only for the account that owns them. There is no way to list templates through this server, so only use an ID or URL the user supplied. Provide either this or `file`, not both.',
```
scottlovegrove:
```suggestion
            'The template to import — either a gallery slug ("product-launch"), a personal template ID ("UT_28Ex..."), or a full Todoist template URL, which is reduced to the ID automatically. Gallery templates work for anyone; personal templates only for the account that owns them. There is no way to list templates through this server, so only use the ID or URL the user supplied. Provide either this or `file`, not both.',
```

src/tools/import-template.ts
```diff
@@ -0,0 +1,162 @@
...
+        ),
+    file: z
```
scottlovegrove: I wonder if this should be called `csvFileContent` so that LLMs don't try sending an actual file.

src/tools/import-template.ts
```diff
@@ -0,0 +1,162 @@
...
+    sections: z.array(SectionSchema).describe('Sections created by the import.'),
+    tasks: z.array(TaskSchema).describe('Tasks created by the import.'),
+    comments: z.array(CommentSchema).describe('Comments created by the import.'),
+    totalCount: z.number().describe('The total number of objects created by the import.'),
+}
+
+const importTemplate = {
+    name: ToolNames.IMPORT_TEMPLATE,
+    // Additive rather than destructive, but there is no dry run and no undo: a wrong ID
+    // silently fills a real project. Require an explicit projectId; never infer one.
```
scottlovegrove: Why are these comments here? They look like instructions for the consuming LLM, but they LLM will never see these.

scottlovegrove (approved):

PR #557, opened 2026-08-04
fix: load all task list pages in MCP app

### LINKED ISSUE #554: Todoist "Today" query returns a partial, consistently-truncated task list (9 of 25 actual tasks)
## Description

The Todoist ChatGPT app/connector consistently returns only 9 of 25 actual tasks when asked for the user's "Today" list. The truncation is stable across repeated, independent invocations, and the full, correct list is confirmed present in the actual Todoist account.

## Reproduction

1. User has 25 tasks due today in their Todoist account (confirmed via the Todoist web app).
2. User asks the ChatGPT "Todoist: To Do List & Calendar" app "what are my tasks for today?"
3. The app renders a "Tasks for Today" card showing only 9 tasks.
4. User re-invokes the integration two more times in the same conversation (`@Todoist: To Do List & Calendar`) — both times, the same 9 tasks are returned.
5. The card's only interactive control ("Open in Todoist") deep-links out to `app.todoist.com/app`, where the full, correct list of 25 tasks is visible. There is no in-card pagination or "load more" affordance.

## Expected behavior

A "Today" query should return all tasks due today, matching what's shown in the Todoist app.

## Actual behavior

Only 9 of 25 tasks are returned, consistently, across multiple fresh invocations.

## Additional notes

- When asked why tasks were missing, the connector/model claimed it had "explicitly excluded overdue tasks" from the query. This explanation doesn't hold up: one of the 9 tasks returned was itself overdue, and all 16 missing tasks were due today (not overdue). So whatever's truncating the result isn't simply an overdue-date filter, at least not as described.
- Not a pagination/UI misunderstanding, confirmed the card's only button opens the web app, it doesn't load more items.
- No existing issue found matching this specific symptom (checked for "today" list truncation via the ChatGPT connector specifically; distinct from #20330 and #20642, which involve different mechanisms).

## References

Zendesk ticket: https://todoist.zendesk.com/agent/tickets/938977

gnapse:
## 🗺️ Overview

- follow cursor-based task pages inside the MCP task-list widget
- append fetched pages while guarding against stale responses
- show a non-blocking error if a later page cannot load

Fixes #554

## 🎥 Demo

Here you can see the widget running locally in isolation connected to a demo account with more than 20 tasks in the today view. It now shows them all.

https://github.com/user-attachments/assets/b2882738-288d-492e-a1ef-e3a9b5486238

### Prerequisites

1. Configure the local MCP server with a valid Todoist API token:

   ```bash
   cp .env.example .env
   # Set TODOIST_API_KEY in .env
   ```

2. Use an account with more than 10 active tasks that match `startDate: "today"` (due today or overdue). The account must be the one referenced by `TODOIST_API_KEY`.
3. Build the local server:

   ```bash
   npm run build
   ```

### Steps

1. Start the local MCP server:

   ```bash
   PORT=3001 npm run start:http
   ```

2. Start an MCP Apps-compatible host connected to `http://localhost:3001/mcp`. To use the reference `basic-host`:

   ```bash
   cd /private/tmp/ext-apps
   npm install
   cd examples/basic-host
   SERVERS="[\"http://localhost:3001/mcp\"]" npm start
   ```

3. Open `http://localhost:8080`, select `find-tasks-by-date`, and call it with:

   ```json
   { "startDate": "today" }
   ```

**Before this change:** the task-list widget renders only the initial cursor page (the default is 10 tasks).

**After this change:** the widget follows `nextCursor` and appends every remaining page.

### Reference-host note

`basic-host` runs in a browser, so connecting it directly to the local server requires a temporary local CORS bridge or proxy. That is test-harness configuration only and is deliberately not part of this PR.

### DIFF
--- a/src/mcp-apps/task-list/app.tsx
+++ b/src/mcp-apps/task-list/app.tsx
@@ -168,10 +168,11 @@ export function App() {
                 fetchPage: (cursor) => fetchTaskPage(app, paginationRequest.args, cursor),
                 onPage: (page) => {
                     if (cancelled || paginationRequest.requestId !== requestId.current) {
-                        return
+                        return false
                     }
 
                     setTasks((previous) => [...(previous ?? []), ...page.tasks])
+                    return true
                 },
             }).catch((error: unknown) => {
                 if (!cancelled && paginationRequest.requestId === requestId.current) {
@@ -192,7 +193,11 @@ export function App() {
         async (taskId: string) => {
             if (!app || !tasks) return
 
-            const previousTasks = tasks
+            const completedTaskIndex = tasks.findIndex((task) => task.id === taskId)
+            const completedTask = tasks[completedTaskIndex]
+            if (!completedTask) return
+
+            const completionRequestId = requestId.current
 
             setTasks((previous) => previous?.filter((task) => task.id !== taskId) ?? [])
             setToolError(null)
@@ -209,7 +214,22 @@ export function App() {
                     )
                 }
             } catch (error) {
-                setTasks(previousTasks)
+                if (completionRequestId !== requestId.current) {
+                    return
+                }
+
+                setTasks((previous) => {
+                    if (!previous || previous.some((task) => task.id === taskId)) {
+                        return previous
+                    }
+
+                    const insertionIndex = Math.min(completedTaskIndex, previous.length)
+                    return [
+                        ...previous.slice(0, insertionIndex),
+                        completedTask,
+                        ...previous.slice(insertionIndex),
+                    ]
+                })
                 setToolError(error instanceof Error ? error.message : 'Failed to complete task.')
             }
         },
--- a/src/mcp-apps/task-list/app.tsx
+++ b/src/mcp-apps/task-list/app.tsx
@@ -1,10 +1,11 @@
-import { useApp, useHostStyles } from '@modelcontextprotocol/ext-apps/react'
-import { useCallback, useMemo, useState } from 'react'
+import { type App as McpApp, useApp, useHostStyles } from '@modelcontextprotocol/ext-apps/react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import type { z } from 'zod'
 import { ArgsSchema } from '../../tools/find-tasks-by-date.js'
 import { ToolNames } from '../../utils/tool-names.js'
 import { Empty } from './empty'
 import { Loading } from './loading'
+import { loadRemainingPages } from './pagination'
 import { TaskList } from './task-list'
 import type { Task } from './types'
 import styles from './task-list.module.css'
@@ -13,6 +14,7 @@ type FindTasksByDateArgs = z.input<z.ZodObject<typeof ArgsSchema>>
 
 type ToolOutput = {
     tasks: Task[]
+    nextCursor?: string
     appliedFilters?: FindTasksByDateArgs
 }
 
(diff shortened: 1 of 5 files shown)


src/mcp-apps/task-list/app.tsx
```diff
@@ -112,6 +155,39 @@ export function App() {
...
+
+            void loadRemainingPages({
+                initialCursor: paginationRequest.cursor,
+                fetchPage: (cursor) => fetchTaskPage(app, paginationRequest.args, cursor),
+                onPage: (page) => {
+                    if (cancelled || paginationRequest.requestId !== requestId.current) {
+                        return
+                    }
+
+                    setTasks((previous) => [...(previous ?? []), ...page.tasks])
```
doistbot, line 174: <a href="#">P2</a>  Preserve pages that arrive while a task completion is pending. If a later page is appended here and `handleComplete` subsequently fails, its catch restores the earlier `previousTasks` snapshot, silently removing all pages appended during that request; pagination will not restart. Roll back only the completed task into the current state rather than replacing the entire task array.

doistbot (commented):
This PR adds cursor-based pagination to the MCP task-list widget so all task pages are fetched and appended incrementally, with guards against stale responses and non-blocking error handling.

Few things worth tightening:
- When a task completion fails, the catch handler restores the entire `previousTasks` snapshot, which silently drops any pages appended during that request and prevents pagination from restarting. Consider rolling back only the completed task within the current state instead of replacing the whole task array.

<details>

- **<a href="#">P3</a> src/mcp-apps/task-list/pagination.ts:31**: When the React effect is cleaned up (e.g. user switches query mid-pagination), the `cancelled` flag in `app.tsx` prevents stale `onPage` state updates, but `loadRemainingPages` has no visibility into cancellation — it keeps calling `fetchPage` for every remaining cursor. Each call hits the MCP server via `app.callServerTool`, so a user who switches queries while a multi-page result is loading triggers unnecessary network requests until the old loop exhausts its cursors. Consider having `onPage` return a `boolean` (or accept a cancellation signal) so `loadRemainingPages` can break out of the loop early: ```ts onPage: (page: PageResult) => boolean // ... const shouldContinue = onPage(page) if (!shouldContinue) return cursor = page.nextCursor ```

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F557&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad395e690-903c-11f1-917c-a5da627efff0%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad395e690-903c-11f1-917c-a5da627efff0&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/mcp-apps/task-list/app.tsx
gnapse: Fixed in 402df52: rollback now reinserts only the failed task into the current list, preserving pages appended while completion was pending. It also ignores completions from a superseded task-list request.

rfgamaral (approved):
Looks good to me 🚀

scottlovegrove (approved):

Issue #558, opened 2026-08-06
Add a reduced toolset profile and an on-demand usage skill resource for small-model clients


scottlovegrove:
## Background

A user on [Reddit](https://www.reddit.com/r/todoist/comments/1vgzccy/using_todoist_mcp_with_claude_making_agent/) reports the MCP server works well in Claude Code and Pi, but fails on nearly every tool call from a phone chat client running Haiku for voice-driven triage. Their diagnosis: 48 tools each carrying a verbose schema is too much for a small model to pick through first time, so it emits malformed calls and gives up. They found the `instructions` block but describe it as general guidance rather than a usage recipe, and worked around it by having the model write per-tool instructions from worked examples.

The failure is schema volume, not absent prose — more always-on guidance makes the context larger, not smaller. The two changes below cut what a client must hold at once and move detail to on-demand delivery.

## What exists today

- `instructions` at `src/mcp-server.ts:61` — ~150 lines, always in context wherever the host surfaces it.
- 48 tools registered unconditionally as a flat list (`src/mcp-server.ts:213-289`).
- `Features` (`src/mcp-helpers.ts:48`) is plumbed through `getMcpServer` but modifies tool *behaviour* (only `strip_emails`); it does not gate registration.
- Resource support exists via `src/mcp-apps/resources.ts`. No `skills/` directory.

## 1. Reduced toolset profile

Control which tools get registered, defaulting to the current full set so existing deployments are unaffected.

- Add a `core` profile; keep `full` as default. Proposed `core`, to confirm during implementation: `find-tasks`, `find-tasks-by-date`, `add-tasks`, `update-tasks`, `reschedule-tasks`, `complete-tasks`, `get-overview`, `find-projects` — 8 against 48.
- Plumb selection through `getMcpServer` options alongside `features`; expose via env var on `src/main.ts` and per-connection on `src/http-app.ts`. Export the type from `src/index.ts`.
- Scope `instructions` to the active profile, so a `core` server ships no guidance for unregistered tools.

## 2. Usage skill served as an MCP resource

Move detailed how-to material out of `instructions` and serve it as fetchable resources — the pattern Figma's MCP server uses: a short instructions block pointing at resource URIs the client retrieves when relevant.

- Add `skills/` with a top-level usage skill plus focused documents for the areas generating the most malformed calls: task create/update semantics, `update-tasks` vs `reschedule-tasks`, filters and search, assignment and collaborator lookup.
[...]

PR #559, opened 2026-08-06
fix: derive the tool surface from a single registry

scottlovegrove:
First of four PRs from #558. No behaviour change — nothing a client receives is different. This one makes the tool surface derive from one place, so the later token work is measuring something trustworthy.

## The drift

`AGENTS.md` told contributors to register each new tool by hand in eight places. Measured against the 47 tools the server actually registers, three of those lists had silently drifted:

| List | State before |
|---|---|
| `src/utils/tool-names.ts` | complete — and the only one with a test asserting it |
| `src/index.ts` `tools` | missing 6: `project-management`, `project-move`, `add-reminders`, `find-reminders`, `update-reminders`, `fetch-object` |
| `scripts/run-tool.ts` | missing 4: `add-reminders`, `find-reminders`, `update-reminders`, `get-productivity-stats` |
| `src/token-footprint.test.ts` | mirrored the bare `findTasksByDate`, not the widget-wrapped object that is actually registered |

`scripts/validate-schemas.ts` iterates the `src/index.ts` export, so `npm run lint:schemas` — the Gemini-compatibility guardrail — had been checking 41 of 47 tools. It now reports `Checked 47 tools with 118 parameters`. The four missing from `run-tool.ts` could not be invoked at all.

The only list that stayed correct is the only one under test. A longer checklist was never going to fix this, so the duplicate lists are gone rather than corrected.

## What changed

`src/tool-registry.ts` is now the single source of truth. `mcp-server.ts` registers by iterating it, and the public `tools` export, `run-tool.ts`, `validate-schemas.ts` and the token-footprint baseline all derive from it. `src/tool-registry.test.ts` fails the build when the remaining views disagree — verified by removing a tool and watching it report 44 against 47.

The `AGENTS.md` checklist drops from eight registration points to one, and `CODEBASE.md` picks up the registry.

## Measurement fix

`token-footprint.test.ts` serialised input schemas with Zod's default `io: 'output'`, while the SDK emits `io: 'input'` with `target: 'draft-7'`. For tools using `.pipe()` that gap is large — `find-activity` measured 938 tokens against an actual 369, because on the output side the pipe resolves to two full ISO regexes.

Reported baseline moves from 40,508 to 39,437. Nothing on the wire changed; the old number was simply wrong. The budget constant stays at 41,000 in this PR and drops in the next one.

## registerTool is no longer generic

Everything `registerTool` does with a tool is dynamic. The generic existed only to type a callback that was already suppressed with `// @ts-expect-error I give up`, so dropping it removes that suppression on the argument path. The remaining one is narrowed to the real cause — `getToolOutput` builds its result incrementally and so is typed looser than `CallToolResult` — and explained in place.

`AnyTodoistTool` declares `execute`'s argument as `never`, which is what lets tools of every parameter shape sit in one collection without a cast: `never` is assignable to any parameter type, so storing a tool is safe while calling it still requires narrowing. The one widening happens inside `registerTool`, where the SDK has already validated the arguments against that tool's own `inputSchema`.

## Verification

- `npm test` — 72 files, 1132 tests
- `npm run type-check`, `npm run format:check` clean
- `npm run lint:schemas` — 47 tools, up from 41
- `getMcpServer()` still registers 47 tools, and `find-tasks-by-date` keeps its `outputSchema` and both `_meta` spellings of the widget resource URI

Downstream `todoist-ai-integrations` needs no change: it calls `getMcpServer({ todoistApiKey, baseUrl, features })`, whose signature is untouched. Its `Object.values(tools)` path in `src/shared/tool-call.ts` gains the six previously-missing tools, which is additive.

### DIFF
--- a/scripts/validate-schemas.ts
+++ b/scripts/validate-schemas.ts
@@ -158,11 +158,9 @@ function validateToolSchema(tool: {
  */
 async function validateAllSchemas(verbose: boolean = false): Promise<ValidationResult> {
     try {
-        const { registeredTools, tools } = await import(`${process.cwd()}/dist/index.js`)
+        const { registeredTools } = await import(`${process.cwd()}/dist/index.js`)
 
-        // `registeredTools` is the full MCP surface; `tools` is the older keyed
-        // export, kept as a fallback so a stale build still validates something.
-        const allTools = registeredTools ?? Object.values(tools)
+        const allTools = registeredTools
 
         const allIssues: ValidationIssue[] = []
         let totalParameters = 0
--- a/scripts/validate-schemas.ts
+++ b/scripts/validate-schemas.ts
@@ -158,14 +158,17 @@ function validateToolSchema(tool: {
  */
 async function validateAllSchemas(verbose: boolean = false): Promise<ValidationResult> {
     try {
-        const { tools } = await import(`${process.cwd()}/dist/index.js`)
+        const { registeredTools, tools } = await import(`${process.cwd()}/dist/index.js`)
+
+        // `registeredTools` is the full MCP surface; `tools` is the older keyed
+        // export, kept as a fallback so a stale build still validates something.
+        const allTools = registeredTools ?? Object.values(tools)
 
         const allIssues: ValidationIssue[] = []
         let totalParameters = 0
-        const toolNames = Object.keys(tools)
 
-        for (const toolName of toolNames) {
-            const tool = tools[toolName]
+        for (const tool of allTools) {
+            const toolName = tool.name
             const toolIssues = validateToolSchema(tool)
             allIssues.push(...toolIssues)
 
@@ -199,7 +202,7 @@ async function validateAllSchemas(verbose: boolean = false): Promise<ValidationR
         return {
             success: allIssues.length === 0,
             issues: allIssues,
-            toolsChecked: toolNames.length,
+            toolsChecked: allTools.length,
             parametersChecked: totalParameters,
         }
     } catch (error) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ import {
     requireValidTodoistToken,
     type RequireValidTodoistTokenOptions,
 } from './middleware/require-valid-todoist-token.js'
-import { registeredTools } from './tool-registry.js'
+import { registeredTools, toolRegistry } from './tool-registry.js'
 // Comment management tools
 import { addComments } from './tools/add-comments.js'
 // Filter management tools
@@ -64,71 +64,12 @@ import { userInfo } from './tools/user-info.js'
 import { viewAttachment } from './tools/view-attachment.js'
 import { validateTodoistToken } from './utils/validate-todoist-token.js'
 
-const tools = {
-    // Task management tools
-    addTasks,
-    completeTasks,
-    uncompleteTasks,
-    updateTasks,
-    findTasks,
-    findTasksByDate,
-    findCompletedTasks,
-    rescheduleTasks,
-    // Project management tools
-    addProjects,
-    updateProjects,
-    findProjects,
-    projectManagement,
-    projectMove,
-    // Section management tools
-    addSections,
-    updateSections,
-    findSections,
-    // Comment management tools
-    addComments,
-    updateComments,
-    findComments,
-    // Reminder management tools
-    addReminders,
-    findReminders,
-    updateReminders,
-    // Attachment tools
-    viewAttachment,
-    // Label management tools
-    addLabels,
-    updateLabels,
-    findLabels,
-    // Filter management tools
-    findFilters,
-    addFilters,
-    updateFilters,
-
-    // Activity and audit tools
-    findActivity,
-    getProductivityStats,
-    // Health and insights tools
-    getProjectHealth,
-    getProjectActivityStats,
-    analyzeProjectHealth,
-    getWorkspaceInsights,
-    // General tools
-    getOverview,
-    deleteObject,
-    fetchObject,
-    reorderObjects,
-    userInfo,
-    // Assignment and collaboration tools
-    findProjectCollaborators,
-    manageAssignments,
-    // Template tools
-    exportProjectTemplate,
-    importProjectTemplate,
-    // Workspace tools
-    listWorkspaces,
-    // OpenAI MCP tools
-    search,
-    fetch,
-}
+/**
+ * Every tool, keyed by its export name.
+ *
+ * Derived from the registry so it cannot drift from what the server registers.
+ */
+const tools = toolRegistry
 
 export {
     // Task management tools
--- a/src/token-footprint.test.ts
+++ b/src/token-footprint.test.ts
@@ -3,106 +3,15 @@ import { describe, expect, it } from 'vitest'
 import { z } from 'zod'
 
 import { instructions } from './mcp-server.js'
-import type { TodoistTool } from './todoist-tool.js'
-import { addComments } from './tools/add-comments.js'
-import { addFilters } from './tools/add-filters.js'
-import { addLabels } from './tools/add-labels.js'
-import { addProjects } from './tools/add-projects.js'
-import { addReminders } from './tools/add-reminders.js'
-import { addSections } from './tools/add-sections.js'
-import { addTasks } from './tools/add-tasks.js'
-import { analyzeProjectHealth } from './tools/analyze-project-health.js'
-import { completeTasks } from './tools/complete-tasks.js'
-import { deleteObject } from './tools/delete-object.js'
-import { exportProjectTemplate } from './tools/export-project-template.js'
-import { fetchObject } from './tools/fetch-object.js'
-import { fetch } from './tools/fetch.js'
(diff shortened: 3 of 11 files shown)


src/token-footprint.test.ts
```diff
@@ -143,7 +52,7 @@ type Row = {
 }
 
 function measure(): Row[] {
-    return allTools.map((tool) => {
+    return registeredTools.map((tool) => {
```
doistbot, line 79: <a href="#">P2</a>  Include `tool._meta` in `buildToolListEntry`. `registeredTools` now contains the widget-wrapped `find-tasks-by-date`, and registration forwards its `_meta.ui.resourceUri` to the client, but the measured entry serializes no `_meta`. The reported `tools/list` and combined totals therefore still undercount the actual payload. Add the same conditional `_meta` field to the measured entry.

src/index.ts
```diff
@@ -4,6 +4,7 @@ import {
     requireValidTodoistToken,
     type RequireValidTodoistTokenOptions,
 } from './middleware/require-valid-todoist-token.js'
+import { registeredTools } from './tool-registry.js'
```
doistbot: <a href="#">P2</a>  `registeredTools` is only re-exported here; the `tools` object below is still assembled from a separate hand-maintained list. This preserves the exact drift point this change is meant to remove: adding a registry entry still requires a matching edit in `index.ts`, with the new test only detecting the omission after the fact. Derive the keyed `tools` view from `registeredTools` (or include the public export key in the registry) so it is actually wired from the single source of truth.

scripts/validate-schemas.ts
```diff
@@ -158,14 +158,17 @@ function validateToolSchema(tool: {
  */
 async function validateAllSchemas(verbose: boolean = false): Promise<ValidationResult> {
     try {
-        const { tools } = await import(`${process.cwd()}/dist/index.js`)
+        const { registeredTools, tools } = await import(`${process.cwd()}/dist/index.js`)
+
+        // `registeredTools` is the full MCP surface; `tools` is the older keyed
+        // export, kept as a fallback so a stale build still validates something.
+        const allTools = registeredTools ?? Object.values(tools)
```
doistbot: <a href="#">P2</a>  The `?? Object.values(tools)` fallback re-introduces the exact silent-degradation mode this PR is fixing. On a stale build where `registeredTools` is absent, the script quietly validates the older keyed `tools` export (which the PR notes was missing 6 tools) and prints `Checked 41 tools` instead of failing. Since `tool-registry.test.ts` now guarantees the two agree, the fallback only ever fires for a stale dist — a user error that should surface fast. Drop the fallback and use `registeredTools` directly; a missing export will throw a clear import error instead of reporting an incomplete pass.

doistbot (commented):
This PR consolidates the tool surface into a single registry (`tool-registry.ts`) so all downstream consumers — the MCP server, public exports, run-tool script, schema validation, and token-footprint tests — derive from one source of truth instead of eight hand-maintained lists.

Few things worth tightening:
- Include `tool._meta` in `buildToolListEntry` so the token-footprint measurement accounts for the `_meta.ui.resourceUri` that registration actually forwards to clients — right now the reported totals undercount the real payload.
- Derive the keyed `tools` export in `src/index.ts` from `registeredTools` directly instead of re-exporting the registry alongside a separately hand-maintained object; otherwise the exact drift point this PR removes is preserved, just caught one CI run later.
- Drop the `?? Object.values(tools)` fallback in `validate-schemas.ts` and use `registeredTools` directly — the fallback silently validates the stale, incomplete list on a dist mismatch instead of failing loud, which is the same silent-degradation mode this PR is fixing.

<details>
<summary>Optional follow-up notes (5)</summary>

- **<a href="#">P3</a> src/tool-registry.ts:157**: `findTasksByDateWithUi` is exported but never imported outside this file. Every consumer uses `registeredTools` instead, which already contains this wrapped version. The unused export adds noise to the module's public API surface.
- **<a href="#">P3</a> src/index.ts:9**: `index.ts` imports all 47 tool modules from their individual file paths (`./tools/add-comments.js`, etc.), duplicating the exact same imports in `tool-registry.ts`. If `tool-registry.ts` exported each tool individually (in addition to the `registeredTools` array), `index.ts` could import every tool from a single module — `import { addComments, addFilters, ... } from './tool-registry.js'`. The `tools` object and named exports would still need explicit entries in `index.ts` (camelCase keys and per-tool types can't be derived mechanically), but tool file paths would live in only one place. Renames or moves would then require updating `tool-registry.ts` alone.
- **<a href="#">P3</a> src/index.ts:142**: The named exports (`export { addTasks, ... }`) form a third manually-maintained list that `tool-registry.test.ts` does not validate. The test only checks that `Object.values(tools)` matches `registeredTools`, but the named re-exports — which let consumers `import { addTasks } from '@doist/todoist-mcp'` — can silently diverge. Adding a tool to `tool-registry.ts` and the `tools` object without adding the corresponding named export would compile and pass tests, leaving it inaccessible via direct import. Consider adding an assertion (or deriving the named exports from `registeredTools` at the type level) so all three views stay in lockstep.
- **<a href="#">P3</a> src/tool-registry.ts:1**: Importing the registry now also imports `mcp-apps/resources`, whose module initialization reads the entire task-list HTML file and hashes it synchronously. This makes every `scripts/run-tool.ts` invocation—including `--help` and non-widget tools—do unnecessary filesystem work, whereas the script only needs executable definitions. Defer the widget HTML/hash initialization until `registerTaskListApp()`/widget registration, and keep the registry's metadata path free of that import-time work.
- **<a href="#">P3</a> src/tool-registry.test.ts:38**: This test reads `run-tool.ts` source and asserts specific import/code strings. That couples it to implementation details: renaming the registry import, extracting a helper function, or rewriting the `.map()` call would break the test even though `run-tool.ts` still correctly derives its tool list from the registry. The real invariant is "every tool in the registry is invokable through run-tool.ts." Verifying that `run-tool.ts` imports `registeredTools` at all is a reasonable smoke test; pinning the exact `.map()` expression is unnecessarily tight. Consider dropping the second `expect` (the `toMatch` on the `.map()` call) and keeping only the import check, or — better — validating end-to-end by importing and calling into `run-tool.ts`'s tool lookup directly.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F559&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A86f81770-91b5-11f1-91b0-31adac345533%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A86f81770-91b5-11f1-91b0-31adac345533&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/token-footprint.test.ts
scottlovegrove: Done in 2db0f8f — `buildToolListEntry` now includes `_meta`. Confirmed the SDK emits it unconditionally on every entry (`server/mcp.js` builds `toolDefinition` with `_meta: tool._meta`), so it is genuinely part of the fixed cost. I also mirror the normalisation `registerAppTool` applies, which fills in whichever of the `ui.resourceUri` / `ui/resourceUri` spellings the tool did not declare — measuring the raw `_meta` would still have undercounted. Baseline moves 39,437 → 39,478.

src/index.ts
scottlovegrove: Done in 2db0f8f — the registry is now an object keyed by public export name and `tools` is that object, so there is no parallel list left to drift. Adding a tool wires up the keyed export with no second edit.

One consequence worth noting: `tools.findTasksByDate` is now the widget-wrapped object rather than the bare definition. That matches what the server actually registers, so the export is more accurate, and the only difference is the extra `_meta`.

The named re-exports (`export { addTasks, ... }`) are still hand-written since ESM needs them static — covered by a new assertion rather than left unguarded.

scripts/validate-schemas.ts
scottlovegrove: Done in 2db0f8f — fallback dropped, `registeredTools` used directly. Agreed the fallback recreated exactly the failure mode this change removes.

nats12 (commented):
✅

Issue #343, opened 2026-03-02
`find-projects` silently swallows color validation errors, returning empty results instead of an error


ghelleks:
**Tool affected:** `find-projects` (both with and without `search` parameter)

**Severity:** High — the search path fails silently, making projects appear not to exist when they do.

### Description

`find-projects` has two distinct failure modes depending on how it's called:

**Without `search` (full list):** Returns a loud `MCP error -32602: Output validation error` when any project has a color value outside the 20-key `ColorKeySchema` enum. At least the failure is visible.

**With `search` (name search):** Silently swallows the same validation error and returns `{"projects": [], "totalCount": 0}` — indistinguishable from a genuine "no match found" response. The user (and any LLM agent) concludes the project doesn't exist.

This was confirmed by searching for `"AI"`, which is an unambiguous substring of the project `"🤖 AI"`. The search returned zero results with no error.

### Steps to Reproduce

1. Have a Todoist account where any project uses a color outside the 20-key enum
2. Have a project whose name contains a searchable substring (e.g. a project named `"🤖 AI"`)
3. Call `find-projects` with `search: "AI"`
4. Observe: `{"projects": [], "totalCount": 0}` — no error, no match, project appears not to exist
5. Call `find-projects` without `search`
6. Observe: loud `MCP error -32602` output validation error

### Root Cause

The name search path appears to catch the output validation exception and return an empty result set rather than propagating the error. This transforms a detectable failure into a silent one, which is significantly worse: an agent receiving an empty result has no signal that anything went wrong and no basis for retrying or escalating.

### Expected Behavior

Both call paths should behave consistently. Either:
- Both return a validation error (loud failure, at least diagnosable), or
- Both coerce/ignore unrecognised color values and return results (correct fix — see below)

### Correct Fix

The output `ProjectSchema` color field should use `.catch(undefined)` or `z.string()` rather than a strict enum, consistent with how `ColorSchema` handles unrecognised input values in PR #328. This would allow both call paths to return projects regardless of color values Todoist has added since the enum was defined.

### Relationship to PR #328

PR #328 added typed color support to input schemas and noted test fixes in `find-projects`. The output schema color field was not updated. Additionally, the silent failure mode in the search path was not addressed.

### Workaround

None reliable via MCP. Use `get-overview` (no `projectId`) to enumerate all projects — it does not apply the same color validation and returns the full project hierarchy successfully.

### Environment

- todoist-ai version: current (`main` as of 2026-03-01)
[...]

PR #560, opened 2026-08-07
feat: advertise output schemas only for widget-backed tools

scottlovegrove:
Second of four PRs from #558, and the one that does the work. Builds on #559.

## The problem

Output schemas were **21k of the server's 39,478-token fixed cost** — more than half of what every client pays before doing anything. They are also invisible to the model: a client uses `outputSchema` to validate `structuredContent` or to render a widget from it, and tools return `structuredContent` whether or not the schema is advertised.

So `registerTool` now only sends `outputSchema` for tools that carry widget metadata.

```
                     before     after
instructions          3,108     3,108
tools/list           36,370    15,640
combined fixed       39,478    18,748     -52.5%
```

No tool removed. No tool result changed. Nothing a client receives is different except the absence of a schema it was not using.

## Why this is safe for the ChatGPT app

`find-tasks-by-date` is the only widget-backed tool, and it keeps its output schema. I diffed its full `tools/list` entry against `main`: **4945 bytes on both sides, identical**.

The rest, verified in the SDK rather than assumed:

- `validateToolOutput` early-returns when a tool has no schema (`server/mcp.js:186-188`). Nothing in the SDK, the Protocol layer or `ext-apps` strips `structuredContent` when no schema is declared — `grep structuredContent` over `server/index.js` and `shared/protocol.js` returns nothing.
- `registerAppTool` only normalises `_meta` between the `ui.resourceUri` and `ui/resourceUri` spellings; it never touches `outputSchema`.
- Client-side validation only runs when a validator was cached, and one is only cached when `outputSchema` is present (`client/index.js:496-545`).
- The widget reads `params.structuredContent` directly (`task-list/app.tsx:63-67,138-139`), never the schema.
- OpenAI's [Apps SDK reference](https://developers.openai.com/apps-sdk/reference) treats `outputSchema` as prescriptive guidance, not a requirement, and documents `window.openai.toolOutput` as "Your `structuredContent`" with no schema precondition.

## Gated on `hasAppUiMeta`, not a list of names

The predicate already decides which registration path a tool takes, three lines further down, so the schema decision and the routing decision cannot disagree — and a future widget tool is covered without anyone remembering to update a list.

## Output schemas are still declared, and still checked

Tool definitions keep `outputSchema`. It types the return, documents the shape, and satisfies each tool's `satisfies TodoistTool<...>` clause.

The one thing genuinely lost is the SDK's runtime output validation, which only runs for advertised schemas. `registerTool` re-runs that check when `NODE_ENV === 'test'`, with a test proving it fires, and production pays nothing.

Be clear about how far that reaches, though: it only covers calls made through the MCP callback, and the tool suites call `tool.execute()` directly — 636 call sites across 48 files. So it is defence in depth on one path, not a guarantee across the suite. Where a tool's output can diverge from its schema at runtime rather than at compile time, `assertStructuredContentMatchesSchema` is exported for the tool's own test to call; `find-projects` uses it, and AGENTS.md documents when to reach for it.

I tried enforcing it everywhere with a vitest `setupFile` wrapping each tool's `execute`. It does not work — importing the tool modules in setup instantiates the real module graph before each test file's `vi.mock` applies, so mocked collaborators get bypassed and 104 tests fail. Snapshots were the other candidate, but they hold only text summaries, not `structuredContent`.

Two things keep the residual risk small. TypeScript still enforces the contract at each tool's definition through its `satisfies TodoistTool<..., typeof OutputSchema>` clause, so only runtime-vs-type divergence can escape. And with the schema unadvertised, that divergence no longer fails a client's tool call — it just means the declared shape is stale. In the one documented instance, issue #343, the SDK's validation was what *caused* the user-visible -32602 failure rather than what prevented it, so losing it there is an improvement.

Worth noting `validateToolOutput` only ever validated — it discards `parseResult.data` — so it was never reshaping payloads, and dropping it changes no bytes.

## Consumers that do read output schemas

Automations reads them. Its authoring model is shown each tool's output field names so it can write `{{steps.X.field}}` references for follow-up steps, and the schema carries the instruction "You MUST destructure the response according to this schema"; its test runner also shapes mocked tool results from them. It connects to `ai.todoist.net/mcp` with no version pin, so this would have reached it on deploy.

So this adds an `output_schemas` feature, using the existing `Features` mechanism, letting such a consumer opt back in. Default behaviour is unchanged.

Merge order, because the header is a no-op until the server understands it but Automations must not be left without schemas:

1. [Doist/automations#3832](https://github.com/Doist/automations/pull/3832) — starts sending `x-mcp-output-schemas: true`
2. **This PR** — releases the feature
3. [Doist/todoist-ai-integrations#183](https://github.com/Doist/todoist-ai-integrations/pull/183) — bumps the dependency and maps the header onto the feature

Nothing changes for any client until step 3 deploys.

## Guardrails

`TOKEN_BUDGET` drops 41,000 → 19,000 and is now documented as a ratchet rather than headroom. Two new assertions:

- exactly one tool may advertise an output schema, so re-enabling them broadly fails loudly rather than silently doubling the surface
- no single tool may exceed 1,300 tokens, which the combined budget would not notice

## Verification

- `npm test` — 72 files, 1138 tests
- `npm run type-check`, `npm run format:check`, `npm run lint:schemas` (47 tools) all clean
- `find-tasks-by-date` `tools/list` entry byte-identical to `main`

Still to do before this reaches users: render the widget end-to-end against the deployed ChatGPT app once this is released. Everything checkable from the code side is checked, but that last step needs the real host.

Downstream `todoist-ai-integrations` needs no change — `getMcpServer`'s signature is untouched. This is `feat:` rather than `feat!:` deliberately: the package's TypeScript API is unchanged, so a major bump would propagate downstream for no reason.

### DIFF
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -349,6 +349,7 @@ function registerTool({
 }
 
 export {
+    assertStructuredContentMatchesSchema,
     FEATURE_NAMES,
     type Feature,
     type FeatureName,
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -2,7 +2,7 @@ import type { TodoistApi } from '@doist/todoist-sdk'
 import { registerAppTool } from '@modelcontextprotocol/ext-apps/server'
 import type { McpServer, ToolCallback } from '@modelcontextprotocol/sdk/server/mcp.js'
 import type { ContentBlock, ToolAnnotations } from '@modelcontextprotocol/sdk/types.js'
-import type { z } from 'zod'
+import { z } from 'zod'
 import type { AnyTodoistTool, ExecuteResult } from './todoist-tool.js'
 import { formatToolExecutionError } from './tool-execution-error.js'
 import { runWithUsageTrackingContext } from './usage-tracking.js'
@@ -232,6 +232,34 @@ function stripEmailsFromText(text: string): string {
 }
 
 /**
+ * Check a tool's output against its declared schema, in tests only.
+ *
+ * The SDK validates `structuredContent` against a tool's `outputSchema`, but
+ * only for tools that advertise one — and most no longer do, to keep them out
+ * of the `tools/list` payload. Tool definitions still carry their schema, so
+ * this keeps that check where mismatches are worth catching: a schema that has
+ * drifted from what the tool actually returns fails CI rather than reaching a
+ * client. Production pays nothing.
+ *
+ * @throws if the output does not match the tool's declared `outputSchema`.
+ */
+function assertStructuredContentMatchesSchema(
+    tool: AnyTodoistTool,
+    structuredContent: Record<string, unknown> | undefined,
+) {
+    if (process.env.NODE_ENV !== 'test' || !tool.outputSchema || !structuredContent) {
+        return
+    }
+
+    const result = z.object(tool.outputSchema).safeParse(structuredContent)
+    if (!result.success) {
+        throw new Error(
+            `Tool ${tool.name} returned structuredContent that does not match its outputSchema: ${result.error.message}`,
+        )
+    }
+}
+
+/**
  * Register a Todoist tool in an MCP server.
  */
 function registerTool({
@@ -270,6 +298,8 @@ function registerTool({
                     executeWithRetry(() => execute(args, client)),
                 )
 
+            assertStructuredContentMatchesSchema(tool, structuredContent)
+
             // Strip emails from outputs for ChatGPT clients on collaborator-related tools
             if (shouldStripEmails) {
                 if (textContent) {
@@ -287,21 +317,28 @@ function registerTool({
         }
     }
 
+    // Output schemas are more than half the fixed cost of `tools/list`, and a
+    // client only needs one to validate `structuredContent` or to render a
+    // widget from it. Tools still return `structuredContent` either way, so
+    // advertising the schema is only worth its tokens for tools with a UI
+    // contract to honour.
+    const appUiMeta = hasAppUiMeta(tool._meta) ? tool._meta : undefined
+
     const toolConfig = {
         description: tool.description,
         inputSchema: tool.parameters,
-        ...(tool.outputSchema ? { outputSchema: tool.outputSchema } : {}),
+        ...(appUiMeta && tool.outputSchema ? { outputSchema: tool.outputSchema } : {}),
         annotations: getMcpAnnotations(tool),
         ...(tool._meta ? { _meta: tool._meta } : {}),
     }
 
-    if (hasAppUiMeta(tool._meta)) {
+    if (appUiMeta) {
         registerAppTool(
             server,
             tool.name,
             {
                 ...toolConfig,
-                _meta: tool._meta,
+                _meta: appUiMeta,
             },
             cb,
         )
@@ -316,6 +353,7 @@ export {
     type Feature,
     type FeatureName,
     type Features,
+    hasAppUiMeta,
     registerTool,
     stripEmailsFromObject,
     stripEmailsFromText,
--- a/src/mcp-helpers.ts
+++ b/src/mcp-helpers.ts
@@ -17,6 +17,8 @@ import { ToolNames } from './utils/tool-names.js'
  * - `'strip_emails'`: Strips email addresses from collaborator tool outputs
  *   (affects: find-project-collaborators, find-completed-tasks). Useful for
  *   clients like ChatGPT that should not have access to user emails.
+ * - `'output_schemas'`: Advertises every tool's output schema, rather than only
+ *   those needed to render a widget.
  */
 const FEATURE_NAMES = {
     /**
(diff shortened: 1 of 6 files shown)


src/mcp-helpers.ts
```diff
@@ -270,6 +298,8 @@ function registerTool({
                     executeWithRetry(() => execute(args, client)),
                 )
 
+            assertStructuredContentMatchesSchema(tool, structuredContent)
```
doistbot, line 317: <a href="#">P2</a>  This check only runs when tests invoke the MCP callback, but the tool suites invoke `tool.execute()` directly (including the `find-projects` regression tests noted in the PR). As a result, an output/schema drift in those tools will still pass CI after their schemas stop being advertised. Add contract coverage that validates representative outputs through this callback (or invokes the same assertion from the tool-output test path) for every declared output schema.

doistbot (commented):
This PR cuts the server's fixed token cost by ~52% by only advertising `outputSchema` to clients for tools that carry widget UI metadata, while keeping a test-only validation guard so declared schemas still get checked in CI.

Few things worth tightening:

- The new test-only schema validation only runs through the MCP callback path, but the individual tool suites call `tool.execute()` directly — so output/schema drift in non-widget tools will slip past CI now that their schemas are no longer advertised. Consider routing representative outputs for every declared schema through the same assertion (or calling it from each tool's own output test) so the guard actually covers the existing test paths.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F560&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1ed154f0-9245-11f1-8bb8-e0645597498d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1ed154f0-9245-11f1-8bb8-e0645597498d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/mcp-helpers.ts
scottlovegrove: Right on both counts — the check only fires on the callback path, and I overstated its reach. The #343 tests say so in their own comment: "The unit tests below call tool.execute() directly, so they do not go through the MCP SDK output-validation layer." 636 `execute()` call sites across 48 files, none reaching the assertion.

I tried enforcing it globally with a vitest `setupFile` wrapping every tool's `execute`. It does not work — importing the tool modules in setup instantiates the real module graph before each test file's `vi.mock` applies, so mocked collaborators get bypassed. 104 tests fail with things like `client.getTasksByFilter is not a function`; `find-tasks.test.ts` alone goes from 53 passing to 33 failing. Reverted.

Snapshots were the other candidate for a broad check, but they only hold text summaries, not `structuredContent`.

So in 51e6698: `assertStructuredContentMatchesSchema` is exported for tool tests to call, used in `find-projects` where this class of drift actually occurs (a colour the API returns but the enum does not list), and documented in AGENTS.md as the thing to reach for when a tool's output can diverge from its schema at runtime.

Worth saying the severity is lower than it first appears. TypeScript still enforces the contract at each tool's definition via `satisfies TodoistTool<..., typeof OutputSchema>`, so only runtime-vs-type divergence can escape — and with the schema unadvertised, that divergence no longer fails a client's tool call; it just means the declared shape is stale. In the one documented instance, #343, the SDK's validation was what *caused* the user-visible -32602 failure rather than what prevented it.

Happy to adopt the helper more widely if you'd rather have the belt and braces — it is a one-line addition per tool test.

scottlovegrove:
Superseded. Dropping output schemas turned out to be the wrong trade: Automations reads them to author manifests (it shows the authoring model each tool's output field names so it can reference them from follow-up steps), which disproved the premise here that clients only need a schema to render a widget — on the first consumer we checked. Making it work needed a feature flag, a header contract and a three-repo ordered deploy, for a benefit we never validated against the failure that prompted this.

Replacing it with a selective trim of the output-schema descriptions: schemas stay advertised, every field name, type and pagination key is preserved, and prose that merely restates the field name goes. No breakage, no coordination. New PR shortly.

PR #561, opened 2026-08-07
perf: trim redundant output schema descriptions

scottlovegrove:
Replaces #560, which tried to stop advertising output schemas altogether. That turned out to be the wrong trade — see the closing comment there — so this takes the part of the win that carries no risk.

## What

Every tool's `outputSchema` ships on every `tools/list`, and the shared schemas in `src/utils/output-schemas.ts` are inlined once per tool that uses them. A single description on `TaskSchema.id` was therefore paid eight times over, and most of them restated the field name:

```
id           "The unique ID of the task."
projectId    "The ID of the project this task belongs to."
color        "The color key of the entity."
hasMore      "Whether there are more results available."
```

Those are gone. What stays is anything the schema cannot express on its own:

| kept | why |
|---|---|
| `dueDate: 'ISO 8601.'` | format, not derivable from `string` |
| `priority: 'p1 is highest, p4 lowest.'` | the enum does not say which end is which |
| `recurring: 'False when not recurring, otherwise the recurrence string.'` | a `boolean \| string` union |
| `workspaceId: 'Undefined for personal projects.'` | conditional presence |
| `isUncompletable: 'An organizational header, not a real task.'` | non-obvious meaning |
| `duration: 'e.g. "2h30m".'` | format |

Where a kept description was wordy it got shortened — `"The due date of the task (ISO 8601 format)"` → `"ISO 8601"`.

## Numbers

```
output-schema descriptions   7,115 → 4,074
combined fixed cost         39,478 → 35,776
```

Deliberately narrower than stripping every output description, which would have saved 9,259 rather than 3,702. The remaining 4,074 is not padding: batch-failure semantics ("A failure here does not affect the other items"), child-list truncation rules, and unions the JSON Schema type alone does not convey. Cutting those would trade real meaning for tokens.

## What is not touched

**The output side only.** Tool descriptions and input field descriptions are how a model decides whether and how to call a tool; both are untouched. `ColorSchema` is a good illustration — its *input* description still enumerates all twenty valid colours, while its *output* counterpart no longer says "The color key of the entity".

Nothing about the tool surface changes: same 47 tools, same schemas advertised, same `structuredContent`. Consumers that read output schemas — [Automations](https://github.com/Doist/automations) shows the authoring model each tool's output field names so it can reference them from follow-up steps — keep every field name, type and pagination key. That is what #560 would have taken away.

## Guardrails

[...]

PR #562, opened 2026-08-07
chore(scripts): add an eval harness for the tool surface

scottlovegrove:
## Why

Every change in #558 so far has been a judgement call about what a model can still do with less text, checked only by reading the diff. `src/token-footprint.test.ts` says what the surface *costs*; nothing said whether it still *works*.

This sends the tool definitions and instructions a real client sends, gives a model a prompt, and asserts which tool it reaches for and how it fills the arguments.

## What it checks

Ten scenarios, each targeting a rule whose loss would produce a malformed or destructive call:

| scenario | guards |
|---|---|
| `reschedule-not-update` | reschedule-tasks vs update-tasks (recurrence loss) |
| `completed-via-activity` | find-activity vs find-completed-tasks |
| `resolve-person` | resolving a name to a user ID |
| `subtask-check` | fetch-object + includeChildren over a speculative find-tasks |
| `label-by-name` | filter by label name, not ID |
| `archive-before-delete` | workspace projects archive before delete |
| `priority-string` | `"p1"`, never the integer `1` |
| `recurring-due-string` | no `"recurring"` prefix on dueString |
| `today-includes-overdue` | `startDate: 'today'` |
| `no-container-echo` | never echo projectId/sectionId/parentId back |

Only the **first** tool call of a turn is inspected and nothing is executed, so it touches no Todoist data and needs no Todoist token.

## First result — the instructions trim is safe

Run against the current `instructions` (3,108 tokens) and against a cut-down version (707 tokens, −77%), 5 repeats × 2 models × 10 scenarios per side, 200 attempts total:

```
                          before   after
claude-haiku-4-5            92%     92%
claude-sonnet-5             86%     84%
```

Eight of ten scenarios sit at 100% on both models on both sides. The two that move are ±20% — one attempt out of five — and they move in opposite directions, so that is noise rather than signal. Cutting 77% of the instructions costs nothing measurable.

## Second result — a pre-existing bug the trim did not cause

`completed-via-activity` fails on **both** sides: 40% → 20% on Haiku, **0% on Sonnet 5 either way**. Asked "what did I actually get done last week?", the model reaches for `find-completed-tasks` or `get-productivity-stats` rather than `find-activity` with `eventType="completed"` — so recurring-task occurrences get missed.

The instructions have said to use `find-activity` for this since before any of this work, in both the long and short versions. Saying it there is not working; the guidance likely needs to sit in `find-completed-tasks`' own description, where the model is actually looking. Worth its own issue.

## Cost and running it

Not wired into `npm test` — it costs money and is non-deterministic, so it reports a pass rate over N repeats rather than pass/fail.

```bash
npx tsx scripts/eval-instructions.ts --label before
# apply your change
npx tsx scripts/eval-instructions.ts --label after
```

Auth comes from the standard Anthropic credential chain, so `ant auth login` is enough — no `ANTHROPIC_API_KEY` required. A cache breakpoint on the system block means the shared ~33k prefix is read rather than rewritten on each attempt: the full 200-attempt sweep above billed ~4.3M cache-read tokens and ~50k cache writes, which is pennies rather than dollars.

`--repeats`, `--models`, and `--scenario` all narrow a run while iterating.

## Caveat worth knowing

The Messages API tool shape has no output-schema field, so a tool's `outputSchema` never reaches the model regardless of how the server advertises it. This harness therefore cannot measure #561 — that change is about what the MCP *client* receives, not what the model sees. It measures tool descriptions, input schemas, and instructions.

### DIFF
--- a/scripts/eval-instructions.ts
+++ b/scripts/eval-instructions.ts
@@ -46,8 +46,15 @@ function firstItem(input: Record<string, unknown>, key: string): Record<string, 
 type Scenario = {
     id: string
     prompt: string
-    /** Any of these counts as the right tool. */
-    expect: string[]
+    /**
+     * Any of these counts as the right tool. Use for a rule that names the
+     * tool to reach for. Omit when the rule is "don't do X" — an allowlist
+     * then fails legitimate lookup steps it did not anticipate, which is
+     * noise, not signal.
+     */
+    expect?: string[]
+    /** Calling any of these fails the scenario. Use for "don't do X" rules. */
+    forbid?: string[]
     /** Extra assertion on the arguments; return a reason on failure. */
     check?: Check
     /** What this scenario is protecting. */
@@ -61,7 +68,9 @@ const SCENARIOS: Scenario[] = [
         // the rule under test.
         id: 'reschedule-not-update',
         prompt: 'Move task 6XG4Vw2c9J ("Weekly review", repeats every Monday) to next Tuesday.',
-        expect: [ToolNames.RESCHEDULE_TASKS],
+        // The rule is "not update-tasks", so anything else -- including a
+        // preliminary lookup -- is acceptable.
+        forbid: [ToolNames.UPDATE_TASKS],
         guards: 'instructions: reschedule-tasks vs update-tasks (recurrence loss)',
     },
     {
@@ -113,10 +122,10 @@ const SCENARIOS: Scenario[] = [
     {
         id: 'archive-before-delete',
         prompt: 'Delete workspace project 6XQ3Plan99 ("Q3 Planning").',
-        // delete-object is deliberately NOT accepted here: deleting a workspace
-        // project before archiving it is the exact failure under test, so listing
-        // it would leave the scenario unable to fail.
-        expect: [ToolNames.PROJECT_MANAGEMENT, ToolNames.FIND_PROJECTS, ToolNames.GET_OVERVIEW],
+        // The rule is "archive first", so the only wrong first move is the
+        // delete itself. Archiving and any lookup (find-projects, get-overview,
+        // fetch-object) are all legitimate openers.
+        forbid: [ToolNames.DELETE_OBJECT],
         guards: 'instructions: workspace projects archive before delete',
     },
     {
@@ -276,13 +285,26 @@ async function runAttempt(
         if (!call) {
             return { ...base, calledTool: null, pass: false, reason: 'no tool call' }
         }
-        if (!scenario.expect.includes(call.name)) {
+        if (scenario.forbid?.includes(call.name)) {
+            return {
+                ...base,
+                calledTool: call.name,
+                pass: false,
+                reason: `called ${call.name}, which this rule forbids`,
+            }
+        }
+        if (scenario.expect && !scenario.expect.includes(call.name)) {
             return {
                 ...base,
                 calledTool: call.name,
                 pass: false,
                 reason: `expected ${scenario.expect.join(' or ')}`,
             }
+        }
+        // An argument check written for a specific tool must not run against a
+        // different one a forbid-only scenario legitimately allows.
+        if (scenario.forbid && !scenario.expect) {
+            return { ...base, calledTool: call.name, pass: true, reason: null }
         }
         const reason = scenario.check?.(call.input as Record<string, unknown>) ?? null
         return { ...base, calledTool: call.name, pass: reason === null, reason }
--- a/scripts/eval-instructions.ts
+++ b/scripts/eval-instructions.ts
@@ -151,7 +151,7 @@ const SCENARIOS: Scenario[] = [
     },
     {
         id: 'today-includes-overdue',
-        prompt: 'What should I focus on today?',
+        prompt: 'Show me the tasks due today.',
         expect: [ToolNames.FIND_TASKS_BY_DATE, ToolNames.GET_OVERVIEW],
         check: (input) => {
             // get-overview takes no date argument; only assert on the dated tool.
--- a/scripts/eval-instructions.ts
+++ b/scripts/eval-instructions.ts
@@ -1,0 +1,345 @@
+#!/usr/bin/env npx tsx
+/**
+ * Measure whether the server's tool surface leads a model to the right tool.
+ *
+ * The token-footprint test says what the surface costs; it cannot say whether
+ * the surface still works. This does: it sends the same tool definitions and
+ * instructions a real client sends, gives the model a prompt, and checks which
+ * tool it reaches for and how it fills the arguments.
+ *
+ * It reads the surface from the working tree, so the way to compare two
+ * versions is to run it on each and diff the pass rates:
+ *
+ *   npx tsx scripts/eval-instructions.ts --label before
+ *   git stash pop
+ *   npx tsx scripts/eval-instructions.ts --label after
+ *
+ * Tool calls are never executed — only the first call of each turn is
+ * inspected — so this touches no Todoist data and needs no Todoist token. It
+ * does spend money on model calls; see --repeats and --models.
+ *
+ * Auth comes from the standard Anthropic credential chain, so `ant auth login`
+ * is enough (no ANTHROPIC_API_KEY needed).
+ *
+ * Usage:
+ *   npx tsx scripts/eval-instructions.ts [--label NAME] [--repeats N]
+ *                                        [--models a,b] [--scenario ID]
+ */
+import { mkdirSync, writeFileSync } from 'node:fs'
+import Anthropic from '@anthropic-ai/sdk'
+import { z } from 'zod'
+import { instructions } from '../src/mcp-server.js'
+import { registeredTools } from '../src/tool-registry.js'
+import { ToolNames } from '../src/utils/tool-names.js'
+
+type Check = (input: Record<string, unknown>) => string | null
+
+type Scenario = {
+    id: string
+    prompt: string
+    /** Any of these counts as the right tool. */
+    expect: string[]
+    /** Extra assertion on the arguments; return a reason on failure. */
+    check?: Check
+    /** What this scenario is protecting. */
+    guards: string
+}
+
+const SCENARIOS: Scenario[] = [
+    {
+        // The ID is in the prompt on purpose. Without one the model has to
+        // search first, which is correct behaviour and tells us nothing about
+        // the rule under test.
+        id: 'reschedule-not-update',
+        prompt: 'Move task 6XG4Vw2c9J ("Weekly review", repeats every Monday) to next Tuesday.',
+        expect: [ToolNames.RESCHEDULE_TASKS],
+        guards: 'instructions: reschedule-tasks vs update-tasks (recurrence loss)',
+    },
+    {
+        id: 'completed-via-activity',
+        prompt: 'What did I actually get done last week?',
+        expect: [ToolNames.FIND_ACTIVITY],
+        check: (input) => {
+            const json = JSON.stringify(input)
+            return json.includes('completed') ? null : `eventType not "completed": ${json}`
+        },
+        guards: 'instructions: find-activity vs find-completed-tasks',
+    },
+    {
+        id: 'resolve-person',
+        prompt: 'Who is Sarah?',
+        expect: [ToolNames.FIND_PROJECT_COLLABORATORS],
+        guards: 'instructions: resolving a name to a user ID',
+    },
+    {
+        id: 'subtask-check',
+        prompt: 'Does task 6XG4Vw2c9J have any subtasks?',
+        expect: [ToolNames.FETCH_OBJECT],
+        check: (input) =>
+            input.includeChildren === true ? null : 'includeChildren not set to true',
+        guards: 'instructions: fetch-object over a speculative find-tasks',
+    },
+    {
+        id: 'label-by-name',
+        prompt: 'Show me my tasks labelled urgent.',
+        expect: [ToolNames.FIND_TASKS, ToolNames.FIND_LABELS],
+        check: (input) => {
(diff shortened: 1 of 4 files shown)


scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+        guards: 'instructions: filter by label name, not ID',
+    },
+    {
+        id: 'archive-before-delete',
+        prompt: 'Delete workspace project 6XQ3Plan99 ("Q3 Planning").',
+        expect: [
+            ToolNames.PROJECT_MANAGEMENT,
+            ToolNames.FIND_PROJECTS,
+            ToolNames.GET_OVERVIEW,
+            ToolNames.DELETE_OBJECT,
```
doistbot: <a href="#">P2</a>  `archive-before-delete` accepts the exact failure it's supposed to catch. `DELETE_OBJECT` is in `expect`, and the `check` passes whenever `json.includes('q3')` — which is always true here because the prompt's project ID is `6XQ3Plan99` (lowercased: `6xq3plan99` contains `q3`). So a model that calls `delete-object` with `{type:'project', id:'6XQ3Plan99'}` as its first move — deleting a workspace project without archiving first — passes both the tool and argument checks. The scenario can't fail the destructive case, which means any 100% result for it is vacuous and can't support the PR's "trim is safe" conclusion for this rule. Drop `ToolNames.DELETE_OBJECT` from `expect` so a direct delete registers as a tool-choice failure; the acceptable first calls are `project-management` (archive), `find-projects`, or `get-overview` (lookup).

scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+        id: 'reschedule-not-update',
+        prompt: 'Move task 6XG4Vw2c9J ("Weekly review", repeats every Monday) to next Tuesday.',
+        expect: [ToolNames.RESCHEDULE_TASKS],
+        guards: 'instructions: reschedule-tasks vs update-tasks (recurrence loss)',
+    },
+    {
+        id: 'completed-via-activity',
+        prompt: 'What did I actually get done last week?',
+        expect: [ToolNames.FIND_ACTIVITY],
+        check: (input) => {
```
doistbot, line 80: <a href="#">P2</a>  A `find-activity` call with only `eventType: "completed"` passes here even though it searches an unbounded history and cannot answer “last week.” Require `dateFrom` and `dateTo` (and ideally `objectType: "task"`) so this scenario measures the documented completion-history workflow rather than only tool selection.

scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+        },
+        guards: 'input field description: priority is "p1".."p4", never an integer',
+    },
+    {
+        id: 'recurring-due-string',
+        prompt: 'Add a task "Water the plants" due every Monday.',
+        expect: [ToolNames.ADD_TASKS],
+        check: (input) => {
+            const json = JSON.stringify(input).toLowerCase()
+            if (!json.includes('monday')) return `recurrence missing: ${json}`
```
doistbot: <a href="#">P2</a>  Checking for `monday` anywhere in the serialized input permits an unscheduled task whose content is, for example, “Water the plants every Monday,” with no `dueString`. Inspect the task's `dueString` specifically and require it to contain the recurrence so this scenario catches omission of the scheduling field.

scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+        }),
+        { input: 0, output: 0, cacheWrite: 0, cacheRead: 0 },
+    )
+    console.log(
+        `\ntokens: ${total.input} uncached in, ${total.cacheWrite} cache write, ` +
+            `${total.cacheRead} cache read, ${total.output} out`,
+    )
+
+    mkdirSync('tmp/eval', { recursive: true })
+    const out = `tmp/eval/${label}.json`
```
doistbot, line 402: <a href="#">P2</a>  `label` comes directly from `--label` and is interpolated into the output path, so `--label ../../package` writes `package.json` rather than an eval artifact. Restrict labels to safe filename characters (for example, `[A-Za-z0-9_-]+`) or resolve the path and reject values outside `tmp/eval` before calling `writeFileSync`.

scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+        expect: [ToolNames.RESCHEDULE_TASKS],
+        guards: 'instructions: reschedule-tasks vs update-tasks (recurrence loss)',
+    },
+    {
+        id: 'completed-via-activity',
+        prompt: 'What did I actually get done last week?',
+        expect: [ToolNames.FIND_ACTIVITY],
+        check: (input) => {
+            const json = JSON.stringify(input)
+            return json.includes('completed') ? null : `eventType not "completed": ${json}`
```
doistbot: <a href="#">P2</a>  Match `eventType` exactly here. `"uncompleted"` contains `"completed"`, so a `find-activity` call for reopen events is currently counted as a pass for a completion-history query. Check `input.eventType === 'completed'` instead.

scripts/eval-instructions.ts
```diff
@@ -0,0 +1,345 @@
...
+                calledTool: call.name,
+                pass: false,
+                reason: `expected ${scenario.expect.join(' or ')}`,
+            }
+        }
+        const reason = scenario.check?.(call.input as Record<string, unknown>) ?? null
+        return { ...base, calledTool: call.name, pass: reason === null, reason }
+    } catch (error) {
+        const message = error instanceof Error ? error.message : String(error)
+        return { ...base, calledTool: null, pass: false, reason: `error: ${message}` }
```
doistbot: <a href="#">P2</a>  API/auth/rate-limit failures are counted as failed model choices and the process still exits successfully. A run with invalid credentials or transient request failures can therefore report a 0% routing score and be saved as a comparison result. Track request errors separately and fail the run (or exclude them from the pass-rate denominator) when the requested samples could not be obtained.

doistbot (commented):
This PR adds an eval harness that sends the tool definitions and instructions to real LLMs and asserts which tool a model reaches for and how it fills the arguments — a much-needed behavioral check alongside the existing token-footprint test.

Few things worth tightening:

- **Scenario assertions are too permissive.** `archive-before-delete` lists `delete-object` in `expect`, so a direct delete of a workspace project passes the check — drop it so a destructive first call registers as a failure. `label-by-name` accepts any `find-activity` with `eventType` containing `"completed"`, which also matches `"uncompleted"`; match `eventType === 'completed'` exactly. The `recurring-due-string` check looks for `monday` anywhere in the serialized input, which lets a task like "Water the plants every Monday" pass with no `dueString` — inspect the `dueString` field specifically.
- **Argument checks are missing for some scenarios.** `completed-via-activity` passes with only `eventType: "completed"` and no date bounds, so it can't actually verify the documented completion-history workflow — require `dateFrom`/`dateTo` (and ideally `objectType: "task"`). `today-includes-overdue` has no argument check despite guarding a specific `startDate: 'today'` value.
- **Path traversal in output filenames.** `--label` is interpolated directly into the output path, so `--label ../../package` writes outside `tmp/eval`. Restrict labels to safe filename characters or resolve and validate the path stays within `tmp/eval` before writing.
- **Request errors silently deflate the score.** API, auth, and rate-limit failures are counted as failed model choices while the process still exits successfully, so a run with bad credentials can report 0% and be saved as a comparison result. Track these separately and exclude them from the pass-rate denominator (or fail the run) when the requested samples couldn't be obtained.

<details>

- **<a href="#">P3</a> scripts/eval-instructions.ts:168**: The `--models` flag splits on comma but doesn't trim whitespace. If a user passes `--models "claude-haiku-4-5, claude-sonnet-5"` (space after comma), the second model name becomes `" claude-sonnet-5"` with a leading space, which the Anthropic API rejects. Fix: `.split(',').map(m => m.trim())`.
- **<a href="#">P3</a> scripts/eval-instructions.ts:139**: `today-includes-overdue` claims to guard `startDate: 'today'` but has no `check`, so it only verifies tool choice. `find-tasks-by-date` accepts `startDate` as either `'today'` or a `YYYY-MM-DD` string (its regex is `^(\d{4}-\d{2}-\d{2}|today)$`), so a model passing a concrete date like `2026-08-07` instead of the dynamic `'today'` keyword passes. That defeats the specific rule the scenario is named after. Add a `check` asserting `input.startDate === 'today'` (or that `'today'` appears in the serialized input) to match the stated guard.
- **<a href="#">P3</a> scripts/eval-instructions.ts:250**: `mapWithLimit` reimplements a FIFO semaphore for concurrency-limited parallel execution. The codebase already exports `createLimiter` from `src/utils/concurrency.ts`, which is listed in CODEBASE.md's "don't reimplement" utils catalog. The script already imports from `src/`, so this could be replaced with `createLimiter(MAX_CONCURRENCY, { queueTimeoutMs: 0 })` + `Promise.all(items.map(item => limiter(() => fn(item))))`, which preserves order identically. The `{ queueTimeoutMs: 0 }` disables the default30s queue timeout, which isn't relevant to a dev-only batch script.
- **<a href="#">P3</a> scripts/eval-instructions.ts:187**: `z.toJSONSchema(...)` returns a broad `JSONSchema` type; casting it `as Anthropic.Tool.InputSchema` suppresses any type mismatch. In practice the `InputSchema` interface has an index signature (`[k: string]: unknown`) so extra JSON Schema fields pass through safely, but if the SDK ever tightens that contract the cast would silently succeed while the runtime value triggers an API error. Consider an explicit runtime validation or a narrower helper type, or at least a comment noting the cast is intentional and safe because of the index signature.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F562&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A10831410-9254-11f1-9b7c-8c59546c7c58%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A10831410-9254-11f1-9b7c-8c59546c7c58&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scripts/eval-instructions.ts
scottlovegrove: Correct, and this was the worst of them — thank you. Both halves were broken: `delete-object` was in `expect`, and `'6XQ3Plan99'.toLowerCase().includes('q3')` is unconditionally true, so the scenario could not fail and its 100% was meaningless.

Fixed, and the real behaviour is worse than the review implies: with `delete-object` no longer accepted, **Haiku 4.5 fails this 0/10** against the current instructions — it deletes a workspace project as its first move. Against the trimmed instructions in #563 it is 10/10. So this scenario went from vacuous to the single strongest result in the whole exercise.

I also reshaped the assertion after your point exposed a pattern. An `expect` allowlist is the wrong shape for a "don't do X" rule: Sonnet 5 opened this by calling `fetch-object` to look the project up by ID — sensible, and something the instructions themselves recommend — and the allowlist scored it as a failure. There is now a `forbid` field, and this scenario forbids `delete-object` rather than allowlisting openers.

scripts/eval-instructions.ts
scottlovegrove: Done in 38b6a7c — requires `dateFrom` and `dateTo`, and `objectType` is covered by the exact `eventType` check below.

Worth flagging what it exposes: with the bounds required, this scenario fails on **both** instruction variants — 13%/0% on Haiku, 0% on Sonnet 5 either way. The routing rule has been in the instructions all along and is not landing, which is a real pre-existing bug rather than anything #563 introduced. Filing separately.

scripts/eval-instructions.ts
scottlovegrove: Done in 38b6a7c — `input.eventType === 'completed'` exactly. "uncompleted" containing "completed" is precisely the kind of thing that would have quietly inflated the score.

scripts/eval-instructions.ts
scottlovegrove: Done in 38b6a7c — it now pulls the first item out of the `tasks` array and checks `dueString` on it, so a recurrence smuggled into `content` no longer passes.

scripts/eval-instructions.ts
scottlovegrove: Done in 38b6a7c — `--label` is restricted to `[A-Za-z0-9_-]+` and rejected with a non-zero exit otherwise. Verified: `--label ../../package` now errors instead of writing `package.json`.

scripts/eval-instructions.ts
scottlovegrove: Done in 38b6a7c. Attempts carry an `errored` flag, request failures are excluded from the pass-rate denominator, reported separately at the end, and set a non-zero exit code. A run that could not obtain its samples is not a result worth comparing against — and given the output file is the comparison artefact, silently saving a 0% run was the dangerous part.

rfgamaral (commented):
✅

PR #563, opened 2026-08-07
perf: cut the instructions block to cross-tool guidance only

scottlovegrove:
Part of #558. Independent of #561 — either can merge first.

## What was in there

`Tool Usage Guidelines` was 2,136 of the block's 3,108 tokens: 36 bullets restating what each tool does and how to fill its parameters. Every one of those facts already reaches the same client in the same payload, via the tool's own `description` and its input field descriptions. `Best Practices` and `Common Workflows` then said much of it a third time — the reschedule rule appeared twice, collaborator lookup three times.

I checked each dropped bullet against the surface that replaces it rather than assuming:

| the bullet said | already in |
|---|---|
| `add-tasks` priority / dueString / deadline / duration mechanics | those four input field descriptions, in full |
| "do not prefix recurrences with `recurring`" | `dueString` field description |
| `"dueString: remove"` | `dueString` field description |
| `find-tasks` filter/filterIdOrName exclusivity | both field descriptions |
| "pass the fileUrl from a comment's fileAttachment" | `view-attachment` description |
| "to read a project rather than export it, use find-tasks" | `export-project-template` description |
| `reminderId` vs `locationReminderId` | `find-reminders` description |

One fact had no other home — that `startDate: 'today'` also includes overdue tasks — so it moves into that field's description, where someone filling the field will see it. That is the +6 tokens on `tools/list`.

## What is kept

Only what a tool description structurally cannot say — which tool to reach for, and how tools relate to each other:

- **reschedule-tasks, not update-tasks**, and why (update-tasks replaces the whole due string and destroys recurrence)
- never echo a task's existing `projectId` / `sectionId` / `parentId` back to update-tasks
- **find-activity, not find-completed-tasks**, for "what did I complete" — completion *events*, including recurring occurrences
- resolving a person to a user ID, and what an empty result actually means
- label **names** for filtering; IDs only for delete-object and update-labels
- archive a workspace project before deleting it
- templates cannot be listed; imports are irreversible
- `isStale` on project health
- batch failures are per-item — never retry the whole batch

## Numbers

```
instructions          3,108 → 777     −2,331  (−75%)
combined fixed       35,776 → 33,453  −2,323
```

Rebased onto `main` now that #561 has merged, so these are on top of the output-schema trim. Combined with #561 the surface goes from **39,478 → 33,453, a 15% cut** from where #558 started.

`TOKEN_BUDGET` ratchets 37,000 → 35,000 in this PR, keeping roughly the headroom the repo has used before (41,000 sat ~1,500 above a 39,478 surface).

## This one is measured, not argued — and it is a fix, not just a saving

Instructions are the highest behaviour-risk thing to cut in this effort, so I built [#562](https://github.com/Doist/todoist-mcp/pull/562) first and ran it on both sides. After the review of #562 corrected several assertions that could not fail, the numbers changed materially. Final matched pair, 10 scenarios × 8 repeats × 2 models per side, **320 attempts, 0 request failures**:

```
                        before (3,108)   after (777)
claude-haiku-4-5              81%            90%
claude-sonnet-5               90%            90%
```

Sonnet 5 scores **identically on all ten scenarios** either way. Haiku 4.5 has exactly one real mover, and it is not a saving — it is a fix:

| scenario | before | after |
|---|---|---|
| `archive-before-delete` | **0%** | **100%** |

Asked to delete a workspace project, Haiku calls `delete-object` directly under the current instructions and archives first under the trimmed ones. Verified separately at n=10: **0/10 vs 10/10**.

The likely reason is placement. Today the rule is a clause inside a long General Operations bullet about `delete-object`; in the trimmed block it is a two-line "Deleting and archiving" section that leads with it. Same fact, and the shorter text is the one the model acts on.

## The one scenario that fails on both sides

`completed-via-activity` — "what did I actually get done last week?" — fails either way: 13% → 0% on Haiku, 0% on Sonnet 5 both. The model reaches for `find-completed-tasks` or `get-productivity-stats` rather than `find-activity` with `eventType="completed"`, so recurring-task occurrences get missed.

The instructions have carried that routing rule all along, in both the long and short versions, so this is pre-existing and not caused by the trim. It does show the guidance is not landing where it is written — it probably needs to be in `find-completed-tasks`' own description. Worth its own issue rather than a behaviour change smuggled in here.

## Verification

`npm test` (72 files, 1133 tests), `type-check`, `format:check`, `lint:schemas` (47 tools) all clean.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -37,7 +37,8 @@ What each tool does and how to fill its parameters is in the tool's own descript
 **Finding things**
 
 - For "what did I complete?", use **find-activity** with objectType="task" and eventType="completed". It reports completion events, including each occurrence of a recurring task. **find-completed-tasks** lists completed tasks, which is not the same question.
-- To resolve a person's name or email to a user ID, or to answer "who is X?", use **find-project-collaborators** with just a searchTerm. It covers every shared project you can access plus yourself, so an empty result means they collaborate on none of them — not that they do not exist. Check a collaborator exists before assigning work to them.
+- To resolve a person's name or email to a user ID, or to answer "who is X?", use **find-project-collaborators** with just a searchTerm. It covers every shared project you can access plus yourself, so an empty result means they collaborate on none of them — not that they do not exist.
+- Before assigning work to someone, call **find-project-collaborators** again with the target projectId. Resolving an ID only proves they collaborate on *some* project you can see; assignment fails unless they collaborate on that one.
 - To find out whether a task hides subtasks, use **fetch-object** with includeChildren rather than a speculative **find-tasks** call.
 - Filter tasks by label **name**. Label IDs are only for **delete-object** and **update-labels**. Shared labels can be renamed but not recoloured, reordered or favourited.
 
@@ -57,7 +58,7 @@ What each tool does and how to fill its parameters is in the tool's own descript
 **Batches**
 
 - Prefer a batch tool over one call per item.
-- Batch tools report per-item \`failures\` alongside successes, and one failure does not undo the rest. Never retry a whole batch: re-send only the items whose failure reason is fixable.
+- Where a batch tool returns per-item \`failures\` alongside successes, one failure does not undo the rest: never retry the whole batch, re-send only the items whose failure reason is fixable. **reschedule-tasks** is the exception — it has no per-item failures and throws if any task fails, so retry it as a whole.
 
 Use the **productivity-analysis** prompt for a combined productivity review — it pulls together user-info, get-productivity-stats and find-completed-tasks.
 
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -24,95 +24,42 @@ You have access to comprehensive Todoist management tools for personal productiv
 - Bulk assignment operations for team workflows
 - Get overviews and insights about workload and progress
 
-### Tool Usage Guidelines:
+### Choosing between tools
 
-**Task Management:**
-- **add-tasks**: Create tasks (max 25 per call) with content, description, priority (\`p1\`, \`p2\`, \`p3\`, \`p4\` strings only; \`p1\` highest and \`p4\` lowest/default; integers are not accepted), dueString (natural language like "tomorrow", "next Friday", "2024-12-25"; also use natural language for recurrences and do not prefix them with \`recurring\`), deadlineDate (ISO 8601 format like "2025-12-31" for immovable constraints), duration (formats like "2h", "90m", "2h30m"), and assignments to project collaborators
-- **update-tasks**: Modify existing tasks - get task IDs from search results first, only include fields that need changes. Supports priority updates using \`p1\`/\`p2\`/\`p3\`/\`p4\` string values (\`p1\` highest, \`p4\` lowest/default; integers are not accepted), due date updates via dueString and due date removal via "dueString: remove", plus deadlineDate (ISO 8601 format like "2025-12-31") updates and removals via "deadlineDate: remove". **IMPORTANT: Do NOT use update-tasks to reschedule/move task dates — use reschedule-tasks instead.** update-tasks replaces the entire due string which destroys recurrence patterns on recurring tasks. Never echo back a task's existing projectId/sectionId/parentId — those fields are treated as a move.
-- **reschedule-tasks**: **Always use this tool when moving/rescheduling task due dates to a different date.** This tool preserves recurring schedules and existing time-of-day. Accepts YYYY-MM-DD or YYYY-MM-DDTHH:MM:SS. Works for both recurring and non-recurring tasks. Do NOT use update-tasks for rescheduling.
-- **complete-tasks**: Mark tasks as done using task IDs
-- **uncomplete-tasks**: Reopen completed tasks using task IDs
-- **find-tasks**: Search by text, project/section/parent container, responsible user, labels, a raw Todoist \`filter\` string (e.g. "today", "p1", "##Work", "(today | overdue) & p1"), or a saved filter by ID or name (\`filterIdOrName\`). Requires at least one search parameter. \`filter\`/\`filterIdOrName\` cannot be combined with projectId/sectionId/parentId, and \`filter\` and \`filterIdOrName\` are mutually exclusive.
-- **find-tasks-by-date**: Get tasks by date range (startDate: YYYY-MM-DD or 'today' which includes overdue tasks) or specific day counts
-- **find-completed-tasks**: View completed tasks by completion date or original due date; if since/until are omitted, defaults to the last 7 days (returns all collaborators unless filtered). For a history of actual task-completion events, including recurring task occurrences, use find-activity instead.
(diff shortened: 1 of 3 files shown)


src/tools/find-tasks-by-date.ts
```diff
@@ -26,7 +26,9 @@ export const ArgsSchema = {
         .string()
         .regex(/^(\d{4}-\d{2}-\d{2}|today)$/)
         .optional()
-        .describe("The start date to get the tasks for. Format: YYYY-MM-DD or 'today'."),
+        .describe(
+            "The start date to get the tasks for. Format: YYYY-MM-DD, or 'today', which also includes overdue tasks.",
```
doistbot: <a href="#">P2</a>  Qualify this as default behavior. `startDate: "today"` combined with `overdueOption: "exclude-overdue"` executes the `today` filter without overdue tasks, so this description contradicts a supported request. Say that `today` includes overdue tasks by default, unless excluded.

src/mcp-server.ts
```diff
@@ -24,95 +24,42 @@ You have access to comprehensive Todoist management tools for personal productiv
...
+**Dates**
+
+- To move a task to a different date use **reschedule-tasks**, never **update-tasks**. update-tasks replaces the whole due string, which destroys recurrence on recurring tasks.
+- Never send a task's existing projectId, sectionId or parentId back to **update-tasks** — those fields are treated as a move.
+- All dates respect the user's timezone.
+
+**Finding things**
+
+- For "what did I complete?", use **find-activity** with objectType="task" and eventType="completed". It reports completion events, including each occurrence of a recurring task. **find-completed-tasks** lists completed tasks, which is not the same question.
+- To resolve a person's name or email to a user ID, or to answer "who is X?", use **find-project-collaborators** with just a searchTerm. It covers every shared project you can access plus yourself, so an empty result means they collaborate on none of them — not that they do not exist. Check a collaborator exists before assigning work to them.
```
doistbot: <a href="#">P2</a>  Validate assignment access against the target project. A lookup without `projectId` can find someone who collaborates on a different shared project but not the task's project, and assignment validation then rejects them. Keep the no-project lookup for resolving an ID, but direct assignment checks to `find-project-collaborators` with the target `projectId`.

src/mcp-server.ts
```diff
@@ -24,95 +24,42 @@ You have access to comprehensive Todoist management tools for personal productiv
...
+- Templates cannot be listed through this server, so only use an ID or URL the user gave you. To start a new project from one, call **add-projects** first and import into it. Imports write immediately and cannot be undone.
+
+**Project health**
+
+- Health data may be stale — check the isStale flag, and call **analyze-project-health** to refresh it before reading again.
+
+**Batches**
+
+- Prefer a batch tool over one call per item.
+- Batch tools report per-item \`failures\` alongside successes, and one failure does not undo the rest. Never retry a whole batch: re-send only the items whose failure reason is fixable.
```
doistbot: <a href="#">P2</a>  Scope this rule to batch tools that return per-item failures. `reschedule-tasks` accepts a batch but has no `failures` output and throws when a sync command fails, so clients cannot inspect and resend only the failed items as instructed. Document it as an exception or provide per-item results.

doistbot (commented):
This PR trims the MCP server instructions block down to cross-tool routing guidance only, moving the one orphaned behavioral detail (that `startDate: "today"` includes overdue tasks) into the `find-tasks-by-date` field description.

Few things worth tightening:
- Qualify the "today includes overdue tasks" note as default behavior — `overdueOption: "exclude-overdue"` overrides it, so the description should say "by default" to avoid contradicting a supported request.
- The collaborator-lookup guidance should steer assignment checks toward `find-project-collaborators` with the target `projectId`, since a no-project lookup can resolve someone who isn't a collaborator on the task's actual project.
- The "inspect `failures` and resend only failed items" instruction applies to batch tools with per-item failure output; `reschedule-tasks` throws on sync failure with no `failures` array, so either scope the rule as an exception or add per-item results.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F563&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A8f07bb00-9255-11f1-92ff-b5d63722c738%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A8f07bb00-9255-11f1-92ff-b5d63722c738&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/find-tasks-by-date.ts
scottlovegrove: Done in de16b5d — now reads "which by default also includes overdue tasks". Correct that the unqualified wording contradicted `overdueOption: 'exclude-overdue'`.

Worth recording: my first attempt at this said "unless overdueOption excludes them", and the eval showed Sonnet 5 dropping from ~82% to ~52% on the scenario that covers this field. That turned out not to be the wording at all — the scenario prompt was "What should I focus on today?", which invites a `user-info` timezone lookup as a first call, and only the first call is inspected. With the prompt tightened to "Show me the tasks due today", both variants are 12/12. Mentioning it because the intermediate numbers are in this PR's history.

src/mcp-server.ts
scottlovegrove: Done in de16b5d — split into two bullets. The first keeps the no-`projectId` lookup for resolving an ID; the second says to call `find-project-collaborators` again with the target `projectId` before assigning, and says why: resolving an ID only proves they collaborate on *some* project you can see.

src/mcp-server.ts
scottlovegrove: Done in de16b5d. Confirmed `reschedule-tasks` has no `failures` output — `grep failures src/tools/reschedule-tasks.ts` returns nothing — so "re-send only the failed items" is impossible there. The rule is now scoped to batch tools that return per-item failures, and names `reschedule-tasks` as the exception with the reason (it throws, so retry it whole).

Adding per-item results to `reschedule-tasks` is the better long-term fix, but it is a behaviour change to a tool this PR does not otherwise touch, so documenting the exception here and leaving that separate.

pedroalves0 (commented):
Love the stats backing up this instruction slim-down 👏

PR #569, opened 2026-08-11
fix: route "what did I complete?" to find-activity

### LINKED ISSUE #564: find-activity is not used for "what did I complete?" — recurring occurrences get missed
## The bug

Asked "what did I actually get done last week?", the model calls `find-completed-tasks` (or `get-productivity-stats`) instead of `find-activity` with `eventType="completed"`.

That matters because they answer different questions. `find-activity` reports completion *events*, so a task that recurs daily and was completed five times shows five occurrences. `find-completed-tasks` lists completed *tasks* — the recurring task appears once, or not at all depending on its current state. For "what did I get done", the recurring work silently vanishes from the answer.

## Measured, not anecdotal

Found by `scripts/eval-instructions.ts` (#562). Pass rate on the `completed-via-activity` scenario, 8 repeats per side:

| model | current instructions | trimmed instructions (#563) |
|---|---|---|
| claude-haiku-4-5 | 13% | 0% |
| claude-sonnet-5 | 0% | 0% |

It fails on **both** sides, so this predates #563 and none of the #558 work caused or fixed it. Reproduce with:

```bash
npx tsx scripts/eval-instructions.ts --scenario completed-via-activity --repeats 8
```

## Why it happens

The disambiguation is written on the tool that is *not* being chosen.

**`find-activity`** already says it, clearly:

> "To answer what someone completed in a period, including recurring task occurrences, use objectType "task", eventType "completed", and dateFrom/dateTo."

**`find-completed-tasks`** says nothing about the distinction:

> "Get completed tasks. since/until are optional and default to a 7-day window when omitted. Includes all collaborators by default. Person-specific queries (summaries, plans, reports) require responsibleUser."

A tool named `find-completed-tasks`, described as "Get completed tasks", is a perfect surface match for "what did I get done". The model stops there — it has no reason to read `find-activity`'s description closely once it has a hit. The guidance is real, correct, and in the wrong place.

The `instructions` block carried this rule too, in both the long and short versions, and it did not land either. That is consistent with what #563 showed more generally: placement beats repetition.

## Suggested fix

Put the caveat where the wrong choice is made — in `find-completed-tasks`' own description. Something like:

> "Get completed tasks. **For "what did I complete/get done" questions, use find-activity instead — it reports completion events, including each occurrence of a recurring task, which this tool does not.** since/until…"

Worth also checking `get-productivity-stats`, which Haiku reached for in some runs; it may want a similar pointer.

## Verifying a fix

The scenario already exists, so a fix is measurable rather than argued:

```bash
# before
npx tsx scripts/eval-instructions.ts --label before --scenario completed-via-activity --repeats 8
# apply the change
npx tsx scripts/eval-instructions.ts --label after --scenario completed-via-activity --repeats 8
```

Anything meaningfully above 0% on Sonnet 5 would be a real improvement; the target is a high rate on both models. Note the scenario also requires `dateFrom`/`dateTo`, so a `find-activity` call over unbounded history still counts as a failure — it cannot answer "last week".

## Context

Fell out of #558. The routing rules that eval covers were largely fine; this is the one that is not.

scottlovegrove:
Closes #564.

## The bug

Asked _"what did I actually get done last week?"_, Haiku 4.5 reaches for `get-productivity-stats` and `find-completed-tasks` and never gets to `find-activity`. Those answer different questions: `find-activity` reports completion **events**, so a daily recurring task completed five times shows five occurrences; `find-completed-tasks` lists completed **tasks**, and the recurring work silently vanishes from the answer.

The guidance already existed — on `find-activity`, and in the `instructions` block. Everywhere except the tool that actually gets picked. `find-completed-tasks` opened with "Get completed tasks", a perfect surface match for the question, and a model that has a hit stops reading. So the caveat now sits where the wrong choice is made, on both `find-completed-tasks` and `get-productivity-stats`.

## The scenario could not measure it

Worth flagging, because it is the more interesting half. #564 reports Sonnet 5 at 0%, and Sonnet 5 turns out to answer the question perfectly. It opens with `user-info` — without today's date and the user's timezone, "last week" is not a date range — and then calls exactly what the scenario asks for:

```
user-info({}) -> find-activity({objectType:"task", eventType:"completed",
                 dateFrom:"2026-08-03T00:00:00+01:00", dateTo:"2026-08-10T00:00:00+01:00", ...})
```

The harness judged the first tool call outright, and `user-info` is not in the `expect` allowlist, so correct behaviour scored 0. This is the failure AGENTS.md:150 already warns about: an allowlist naming the tool under test cannot also name every reasonable lookup that precedes it. Fixing tool descriptions against that reading would have been optimising against noise.

Two changes to the harness, both narrow:

- A **context tool** (`CONTEXT_TOOLS`, currently just `user-info`) is answered with a canned result, and the next call is judged. Capped at `MAX_CONTEXT_HOPS` so a model that only ever gathers context still fails. A tool a scenario explicitly expects is never stubbed, so a future scenario testing the route *to* `user-info` stays measurable.
- A turn carrying **several calls at once** is judged on its first non-context call, not on whichever block came first — the order between "ask for the date" and "query the log" is arbitrary. A forbidden call outranks that, since catching it is the point of a forbid rule.

Adding `user-info` to `expect` would have been the cheap fix and the wrong one: the judge would pass on the lookup alone and check no arguments at all, scoring 100% while measuring nothing. AGENTS.md now documents both traps.

## Filtering the answer to the asker

From review, and it turned out to be a real hole rather than a theoretical one. `find-activity` reports every user's events by default, so rerouting a first-person question to it means a shared project answers "what did I get done" with collaborators' completions mixed in. Asserting `initiatorId` in the scenario put it at **0% on both models** — nothing was asking for it.

Sonnet 5 was already passing it unprompted and went to 100% once `find-activity`'s description named it. Haiku needed the routing rule to name the *order* of the two calls — `user-info` first for the ID, then `find-activity` — because it was answering in one shot and had no ID to pass. Every Haiku run that calls `user-info` first passes; the ones that skip it are the remaining failures.

## Numbers

`completed-via-activity`, 12 repeats, `main` vs this branch, both judged by the harness as it stands here (context hops, and `initiatorId` asserted):

| model | main | this branch |
|---|---|---|
| claude-haiku-4-5 | 0% | **67%** |
| claude-sonnet-5 | 100% | **100%** |

Sonnet's "before" is 100% because its routing was never broken — the harness fix alone is what surfaced that, and the argument assertions it already satisfied. Haiku's 67% is the honest 12-repeat figure; smaller samples ran higher, and the residual failure is always the same one, skipping `user-info` and so having no ID to filter by.

Splitting the two halves apart, since they moved for different reasons: on **routing alone** (which tool gets picked, the subject of #564), this branch is 100% on both models against main's 38% for Haiku. On **filtering to the asker**, 0% → 67% Haiku, 0% → 100% Sonnet.

Full sweep, 10 scenarios x 5 repeats: **100/100 on both models**, up from 45/50 — `completed-via-activity` was the only failing scenario, and no other scenario moved.

```bash
npm run eval -- --label after --scenario completed-via-activity --repeats 12
npm run eval -- --label after-full --repeats 5
```

## Cost

Token footprint 33,453 → 33,624 of a 35,000 budget: +138 across three tool descriptions and one input field, plus 33 on the `instructions` block (777 → 810). The budget is a ratchet, so to be explicit — this is a rise, bought with a routing failure fixed on the model that was getting it wrong, and a correctness gap closed on both.

## Also from review

Both P3 notes applied: the `user-info` stub is now tied to `UserInfoStructured` with `satisfies`, so a renamed or added field on the real schema breaks the harness instead of drifting; and `currentLocalTime` now uses the `toLocaleString('en-US', …)` format the tool actually emits rather than ISO.

## Left for later

- The `instructions` block bullet grew rather than shrank here, since naming the call order is what moved Haiku. Whether the tool descriptions can now carry that alone is worth a measured follow-up.
- The scenario's `check` does not assert `objectType === "task"`, though the instructions and `find-activity` both ask for it. Both models set it anyway. Tightening it is a separate change, since it would move the baseline.

### DIFF
--- a/scripts/eval-instructions.ts
+++ b/scripts/eval-instructions.ts
@@ -14,9 +14,11 @@
  *   git stash pop
  *   npx tsx scripts/eval-instructions.ts --label after
  *
- * Tool calls are never executed — only the first call of each turn is
- * inspected — so this touches no Todoist data and needs no Todoist token. It
- * does spend money on model calls; see --repeats and --models.
+ * Tool calls are never executed against Todoist, so this touches no Todoist
+ * data and needs no Todoist token. Judging looks at one call per attempt: the
+ * first, unless it is a context tool (see CONTEXT_TOOLS), in which case a
+ * canned result is fed back and the next call is judged. It does spend money
+ * on model calls; see --repeats and --models.
  *
  * Auth uses the Anthropic SDK's standard credential chain: either
  * ANTHROPIC_API_KEY (a key from the Anthropic Console at platform.claude.com),
@@ -195,6 +197,49 @@ const DEFAULT_MODELS = ['claude-haiku-4-5', 'claude-sonnet-5']
 const DEFAULT_REPEATS = 5
 const MAX_CONCURRENCY = 4
 
+/**
+ * Tools a model may legitimately call before the one a scenario is testing, to
+ * establish context it has no other way to get — today's date, the user's
+ * timezone. "What did I get done last week?" cannot be turned into a date range
+ * without them.
+ *
+ * Judging the very first call outright scores that correct behaviour as a
+ * failure: an `expect` allowlist naming the tool under test cannot also name
+ * every reasonable lookup that precedes it. So a call to one of these is
+ * answered with the canned result below and judging moves to the next call.
+ *
+ * The result is fiction, but only its shape and internal consistency matter —
+ * no scenario asserts on these values, and nothing runs against Todoist. Keep
+ * it in step with the tool's real output schema so the model reads it the way
+ * it would read a live one.
+ */
+const CONTEXT_TOOLS: Record<string, unknown> = {
+    [ToolNames.USER_INFO]: {
+        type: 'user_info',
+        userId: '2671355',
+        fullName: 'Eval User',
+        timezone: 'Europe/London',
+        currentLocalTime: '2026-08-11 09:30:00',
+        startDay: 1,
+        startDayName: 'Monday',
+        weekStartDate: '2026-08-10',
+        weekEndDate: '2026-08-16',
+        currentWeekNumber: 33,
+        completedToday: 3,
+        dailyGoal: 5,
+        weeklyGoal: 30,
+        email: 'eval@example.com',
+        plan: 'Todoist Pro',
+    },
+}
+
+/**
+ * How many context calls an attempt may make before it is judged a failure.
+ * A model that keeps gathering context is not answering the question, and
+ * without a cap a loop would bill for turns forever.
+ */
+const MAX_CONTEXT_HOPS = 3
+
 function parseArgs() {
     const args = process.argv.slice(2)
     const get = (flag: string) => {
@@ -254,6 +299,28 @@ type Attempt = {
      */
     errored: boolean
     usage: Usage
+    /** Context calls answered with a stub before the judged one. */
+    contextHops: number
+}
+
+/** Verdict on the one call a scenario is judged by. */
+function judge(
+    scenario: Scenario,
+    call: Anthropic.ToolUseBlock,
+): { pass: boolean; reason: string } {
+    if (scenario.forbid?.includes(call.name)) {
+        return { pass: false, reason: `called ${call.name}, which this rule forbids` }
+    }
+    if (scenario.expect && !scenario.expect.includes(call.name)) {
+        return { pass: false, reason: `expected ${scenario.expect.join(' or ')}` }
+    }
+    // An argument check written for a specific tool must not run against a
+    // different one a forbid-only scenario legitimately allows.
+    if (scenario.forbid && !scenario.expect) {
+        return { pass: true, reason: '' }
+    }
+    const reason = scenario.check?.(call.input as Record<string, unknown>) ?? null
+    return { pass: reason === null, reason: reason ?? '' }
 }
 
 async function runAttempt(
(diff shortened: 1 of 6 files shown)


src/tools/find-completed-tasks.ts
```diff
@@ -111,7 +111,7 @@ const OutputSchema = {
 const findCompletedTasks = {
     name: ToolNames.FIND_COMPLETED_TASKS,
     description:
-        'Get completed tasks. since/until are optional and default to a 7-day window when omitted. Includes all collaborators by default. Person-specific queries (summaries, plans, reports) require responsibleUser.',
+        'Get completed tasks in a date range. For "what did I complete/get done" questions, use find-activity instead — it reports completion events, including every occurrence of a recurring task, which this tool does not. since/until are optional and default to a 7-day window when omitted. Includes all collaborators by default. Person-specific queries (summaries, plans, reports) require responsibleUser.',
```
doistbot, line 114: <a href="#">P2</a>  This reroutes first-person questions to `find-activity`, whose documented default includes events from all users. In a shared project, “what did I complete?” can therefore include collaborators’ completions. Tell the model to obtain the current user ID (via `user-info`) and pass it as `initiatorId` for first-person completion queries; add that assertion to the eval scenario as well.

scripts/eval-instructions.ts
```diff
@@ -267,51 +334,71 @@ async function runAttempt(
...
-                reason: `expected ${scenario.expect.join(' or ')}`,
+
+            // A turn can carry several calls at once. Judge the substantive one
+            // rather than whichever block came first: a model that asks for the
+            // date and queries the activity log in the same turn has made its
+            // choice, and the order between the two is arbitrary. A forbidden
+            // call outranks that — catching it is the point of a forbid rule.
+            const substantive =
+                calls.find((c) => scenario.forbid?.includes(c.name)) ??
+                calls.find((c) => !(c.name in CONTEXT_TOOLS))
```
doistbot: <a href="#">P2</a>  Use an own-property check here rather than `in`. A hallucinated tool name such as `__proto__` is currently treated as a context tool because it is inherited from `Object.prototype`; it receives a fake result and a later expected call can incorrectly pass the attempt. `Object.hasOwn(CONTEXT_TOOLS, c.name)` should define the context-tool predicate.

doistbot: <a href="#">P2</a>  This globally consumes `user-info`, so a future scenario with `expect: [ToolNames.USER_INFO]` can never pass: its expected call is stubbed instead of judged, then the next response is judged or fails for no tool call. Only bypass a context tool when it is not explicitly expected by the scenario, while retaining the existing forbid precedence.

doistbot (commented):
This PR reroutes "what did I complete?" to `find-activity` by adding routing caveats to `find-completed-tasks` and `get-productivity-stats`, and upgrades the eval harness to stub context-tool lookups like `user-info` so the substantive call is judged instead of the preliminary one.

Few things worth tightening:
- `find-activity` defaults to all users, so in shared projects a first-person query can surface collaborators' completions — guide the model to fetch the current user ID via `user-info` and pass it as `initiatorId`, and add that assertion to the eval scenario.
- Use `Object.hasOwn(CONTEXT_TOOLS, c.name)` instead of `in` so a hallucinated name like `__proto__` isn't treated as a context tool via prototype inheritance.
- Only bypass a context tool when it isn't explicitly expected by the scenario, otherwise a future `expect: [ToolNames.USER_INFO]` scenario can never pass — its expected call gets stubbed instead of judged.

<details>

- **<a href="#">P3</a> scripts/eval-instructions.ts:216**: The `user-info` stub duplicates the `UserInfoStructured` shape by hand, and the comment above it notes the need to keep it in step with the real output schema. That link already exists: `UserInfoStructured` is exported from `src/tools/user-info.ts`. Annotating the entry with `satisfies UserInfoStructured` (keeping the outer map as `Record<string, unknown>` so future context tools with other shapes still fit) would turn the manual "keep in step" discipline into a compile-time check, so a renamed or newly-required field on the real schema breaks here instead of silently drifting.
- **<a href="#">P3</a> scripts/eval-instructions.ts:222**: The `currentLocalTime` stub is `'2026-08-11 09:30:00'` (YYYY-MM-DD HH:MM:SS), but the real `user-info` tool formats it via `toLocaleString('en-US', …)`, which produces `'08/11/2026, 09:30:00'` (MM/DD/YYYY, HH:MM:SS). The comment on `CONTEXT_TOOLS` says to keep the stub in step with the real output so the model reads it as it would a live one. The other date fields (`weekStartDate`, `weekEndDate`) already match the real `getDateString` format; `currentLocalTime` is the only one that doesn't. Low impact today (both models pass at 100%, and `weekStartDate` is the primary field for deriving "last week"), but worth aligning so a future model that leans on `currentLocalTime` isn't trained on a format it won't see in production.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F569&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A85b608f0-9583-11f1-9385-9f1f72057d68%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A85b608f0-9583-11f1-9385-9f1f72057d68&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/find-completed-tasks.ts
scottlovegrove: Done in fafcdfa — and it was a real hole, not a theoretical one. Asserting `initiatorId` in the scenario put it at 0% on both models, since nothing was asking for it.

Sonnet 5 was already passing it unprompted, so it went to 100% as soon as `find-activity`'s description named it. Haiku needed more: it answered in one shot, never calling `user-info`, so it had no ID to pass. Naming the order in the routing rule — `user-info` first for the ID, then `find-activity` — is what moved it. Every Haiku run that calls `user-info` first now passes; the ones that skip it are the remaining failures.

12 repeats, judged with the new assertion: Haiku 0% → 67%, Sonnet 0% → 100%.

scripts/eval-instructions.ts
scottlovegrove: Done in fafcdfa — `Object.hasOwn(CONTEXT_TOOLS, c.name)`, wrapped in an `isContext` predicate so the check has one definition rather than two.

scottlovegrove: Done in fafcdfa — `isContext` is now `Object.hasOwn(CONTEXT_TOOLS, c.name) && !scenario.expect?.includes(c.name)`, so a scenario that expects a context tool judges it instead of stubbing it. Forbid precedence is unchanged, still checked ahead of everything else.

PR #570, opened 2026-08-12
fix: preserve nullable fields in MCP tool output

doistbot:
# Pull Request

Closes https://github.com/Doist/Issues/issues/20842

## Short description

Preserves nullable structured-output fields through MCP serialization so unscoped workspace insights responses retain `folderId: null`. Added a regression test covering both structured content and legacy JSON output.

## PR Checklist


scottlovegrove:
@doistbot I don't think the solution here is to just _stop removing null fields_. I think the solution here is to update the schema (or whatever is validating) to accept a missing folderId field.

scottlovegrove:
Superseded by #571. The field is not just failing validation — the API removed `folder_id` from the workspace insights endpoint in Doist/todoist#27366, so nothing can populate it. #571 drops it from the tool, gives the three other tools with the same required-and-nullable output fields the same treatment, keeps `removeNullFields`, and adds a registry-wide test that rejects nullable output fields.

PR #580, opened 2026-08-18
feat(comments): notify collaborators when adding a comment

### LINKED ISSUE #509: Feature request: Support user mentions and notifications in Todoist MCP
Currently, the Todoist MCP integration doesn't support mentioning or notifying users when posting comments on tasks.

It would be valuable to add support for selecting project collaborators (e.g. via @username or user ID) and triggering the same notifications that are sent when a user is mentioned through the native Todoist app.

This would enable:

Mention teammates in task comments.
Notify specific collaborators that their attention is needed.
Create more collaborative automation workflows without requiring users to manually notify teammates afterward.
Better support team-based task management 

Example use case:

`Comment on the task: '@John Could you review this by tomorrow?'`

The MCP server would resolve the mentioned user and send the appropriate Todoist notification, matching the existing in-app behavior.

This would significantly improve collaboration capabilities for teams using Todoist MCP with AI tools such as Claude, ChatGPT, Cursor, and VS Code.

scottlovegrove:
Closes #509

## The problem

Comments posted through this server notify nobody. `add-comments` never sent `uidsToNotify`, so a teammate named in an agent's comment found out only if they happened to open the task.

That also breaks the *next* comment, which is the part users notice and can't reproduce. Todoist's clients pick the recipients themselves and send them with each comment; the API notifies exactly who it is handed and derives nobody on its own. On a first comment the clients notify the assignee, the assigner and the creator; on a reply they notify the previous comment's participants, so threads keep flowing without everyone being re-tagged. A comment posted with an empty list therefore silences the comment that follows it — including one a human writes in the app. That's the behaviour @MelisUnal described on the issue (Zendesk 940230), where a scheduled agent replying in-task was "cut off right at the delivery point".

## What this does

`add-comments` gains a per-comment `notifyUsers`, accepting user IDs, emails, full names or `"me"` for each person — the same reference shapes `responsibleUser` already takes — resolved through the existing `userResolver`. A new `resolveUserRefs` wraps the singular resolver for lists, deduping and reporting every unresolvable reference in one error rather than failing on the first.

Omitting `notifyUsers` mirrors the clients: the assignee, assigner and creator on a task's first comment, or the previous comment's participants on a reply. Passing `["none"]` stays silent, following the repo's existing `"remove"` / `"unassign"` convention.

Recipients are worked out once per distinct target, so several comments on one task in a single batch read the thread once and notify the same people. Comments now also report `notifiedUserIds`, which lets a caller see a thread's participants before replying, and flows through `find-comments` and `update-comments` for free.

`@mentions` in the comment text are deliberately **not** parsed — the model passes `notifyUsers` explicitly, which avoids guessing at multi-word names, email addresses and code snippets. The server instructions tell it to do so.

Notification on *edit* is out of scope: `UpdateCommentArgs` is `{ content }` only, and only the sync `note_update` command carries recipients.

## The SDK fix this needed

Live testing turned up a bug in `@doist/todoist-sdk`: `addComment` joined `uidsToNotify` into a comma-separated string, which the API rejects outright with `400 INVALID_ARGUMENT_VALUE / uids_to_notify / "Input should be a valid list"`. No caller could notify anyone, and nothing caught it because no consumer set the field. Fixed in Doist/todoist-sdk-typescript#664 and released as 14.0.1, which this PR bumps to.

## Verification

`npx tsc --noEmit`, all 1205 tests, `npm run format:check` and `npm run build` pass.

End-to-end against the live API on the published 14.0.1, using a throwaway task in a shared project (since removed):

| case | result |
| --- | --- |
| first comment, unassigned self-created task | no recipients |
| `notifyUsers: ["Ada Lovelace"]` | `notifiedUserIds: ["9876543"]`, "Notified 1 person" |
| reply, defaults | previous comment's participants, author excluded |
| `notifyUsers: ["none"]` | silent, no recipient field sent |
| `notifyUsers: ["ada@example.com", "9876543"]` | both refs resolve to one user, deduped to 1 |
| two unresolvable names | one error naming both |

The multi-user paths — a reply picking up someone else's participants, and a first comment on a task assigned to another person — are covered by unit tests rather than live calls, since verifying them for real means sending test notifications to an actual colleague.

### DIFF
--- a/src/tools/add-comments.ts
+++ b/src/tools/add-comments.ts
@@ -1,9 +1,16 @@
-import type { AddCommentArgs } from '@doist/todoist-sdk'
+import type { AddCommentArgs, TodoistApi } from '@doist/todoist-sdk'
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { isInboxProjectId, mapComment, resolveInboxProjectId } from '../tool-helpers.js'
+import {
+    type CommentTarget,
+    NO_NOTIFY_KEYWORD,
+    getDefaultCommentRecipients,
+    isNoNotifyList,
+} from '../utils/comment-recipients.js'
 import { CommentSchema as CommentOutputSchema } from '../utils/output-schemas.js'
 import { ToolNames } from '../utils/tool-names.js'
+import { resolveUserRefs } from '../utils/user-resolver.js'
 
 const CommentSchema = z.object({
     taskId: z.string().optional().describe('The ID of the task to comment on.'),
@@ -14,6 +21,12 @@ const CommentSchema = z.object({
             'The ID of the project to comment on. Project ID should be an ID string, or the text "inbox", for inbox tasks.',
         ),
     content: z.string().min(1).describe('The content of the comment.'),
+    notifyUsers: z
+        .array(z.string().min(1))
+        .optional()
+        .describe(
+            `Who to notify about this comment — a user ID, email, full name, or "me" for each person. Set this whenever the comment mentions someone; the text of an @mention notifies nobody on its own. Omit to notify whoever the Todoist apps would (the task's assignee, assigner and creator on a first comment, or the previous comment's participants on a reply). Pass ["${NO_NOTIFY_KEYWORD}"] to notify nobody.`,
+        ),
 })
 
 const ArgsSchema = {
@@ -26,10 +39,13 @@ const OutputSchema = {
     addedCommentIds: z.array(z.string()).describe('The IDs of the added comments.'),
 }
 
+type CommentInput = z.infer<typeof CommentSchema>
+type TodoistUser = Awaited<ReturnType<TodoistApi['getUser']>>
+
 const addComments = {
     name: ToolNames.ADD_COMMENTS,
     description:
-        'Add multiple comments to tasks or projects. Each comment must specify either taskId or projectId.',
+        'Add multiple comments to tasks or projects, optionally notifying collaborators. Each comment must specify either taskId or projectId.',
     parameters: ArgsSchema,
     outputSchema: OutputSchema,
     annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false },
@@ -50,25 +66,48 @@ const addComments = {
             }
         }
 
-        // Check if any comment needs inbox resolution
+        // Every comment needs the current user, either to resolve "inbox" or to
+        // keep the author out of their own comment's recipients.
         const needsInboxResolution = comments.some((comment) => isInboxProjectId(comment.projectId))
-        const todoistUser = needsInboxResolution ? await client.getUser() : undefined
+        const needsDefaultRecipients = comments.some((comment) => !comment.notifyUsers)
+        const todoistUser =
+            needsInboxResolution || needsDefaultRecipients ? await client.getUser() : undefined
 
-        const addCommentPromises = comments.map(async ({ content, taskId, projectId }) => {
-            // Resolve "inbox" to actual inbox project ID if needed
-            const resolvedProjectId = await resolveInboxProjectId({
-                projectId,
-                user: todoistUser,
-                client: todoistUser ? undefined : client,
-            })
+        const targets = await Promise.all(
+            comments.map(async (comment) => {
+                // Resolve "inbox" to actual inbox project ID if needed
+                const resolvedProjectId = await resolveInboxProjectId({
+                    projectId: comment.projectId,
+                    user: todoistUser,
+                    client: todoistUser ? undefined : client,
+                })
 
-            return await client.addComment({
-                content,
-                ...(taskId ? { taskId } : { projectId: resolvedProjectId }),
-            } as AddCommentArgs)
+                return (
+                    comment.taskId ? { taskId: comment.taskId } : { projectId: resolvedProjectId }
+                ) as CommentTarget
+            }),
+        )
+
+        const recipients = await resolveRecipientsPerComment({
+            comments,
+            targets,
+            client,
+            currentUser: todoistUser,
         })
 
-        const newComments = await Promise.all(addCommentPromises)
+        const newComments = await Promise.all(
+            comments.map(async ({ content }, index) => {
+                const uidsToNotify = recipients[index] ?? []
+                return await client.addComment({
+                    content,
+                    ...targets[index],
+                    // Nobody to notify means no recipient field at all,
+                    // rather than an empty one.
+                    ...(uidsToNotify.length > 0 && { uidsToNotify }),
+                } as AddCommentArgs)
+            }),
+        )
+
         const mappedComments = newComments.map(mapComment)
         const textContent = generateTextContent({ comments: mappedComments })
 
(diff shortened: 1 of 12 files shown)


src/utils/user-resolver.ts
```diff
@@ -381,3 +381,39 @@ export async function resolveUserNameToId(
...
+ * failing on the first one, so the caller learns about all the bad references
+ * at once.
+ *
+ * @param client - Todoist API client
+ * @param refs - User references to resolve
+ * @returns The resolved users, deduplicated by user ID
+ * @throws Error naming every reference that could not be resolved
+ */
+export async function resolveUserRefs(client: TodoistApi, refs: string[]): Promise<ResolvedUser[]> {
+    const resolutions = await Promise.all(refs.map((ref) => userResolver.resolveUser(client, ref)))
```
doistbot: <a href="#">P2</a>  Resolving refs with `Promise.all` makes each non-ID reference independently miss the user and collaborator caches before any call has populated them. A comment naming several people therefore runs `getProjects` and every shared project's collaborator lookup once per name, concurrently, which can rate-limit or time out on shared accounts. Load the current user/collaborators once for the list (then match each ref), or otherwise deduplicate/in-flight-cache this shared lookup.

src/tools/add-comments.ts
```diff
@@ -14,6 +21,12 @@ const CommentSchema = z.object({
             'The ID of the project to comment on. Project ID should be an ID string, or the text "inbox", for inbox tasks.',
         ),
     content: z.string().min(1).describe('The content of the comment.'),
+    notifyUsers: z
+        .array(z.string().min(1))
```
doistbot, line 26: <a href="#">P2</a>  `notifyUsers` has no upper bound, unlike every other batch/array input in the codebase. A single `add-comments` call can pass an arbitrarily long list, and `resolveUserRefs` fans the whole list out with `Promise.all` *before* deduping — so N duplicates or N unresolvable names each trigger `userResolver.resolveUser`, which on a cold cache runs `getUser` plus `getAllCollaborators` (paginating all shared projects and their collaborators) concurrently. One untrusted argument can therefore multiply into unbounded parallel SDK requests against the operator's token and rate limit. Cap the array with `.max(...)` (matching the other tool inputs) and/or dedupe the refs before resolving them.

doistbot, line 26: <a href="#">P2</a>  `notifyUsers: []` is accepted, but it is truthy in `resolveRecipientsPerComment`, so it bypasses default recipients and posts silently. This contradicts the documented opt-out shape (`["none"]`) and makes an accidental empty list cut off the notification chain. Require at least one entry with `.min(1)`, or explicitly normalize an empty list to the default behavior.

src/utils/comment-recipients.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+async function getLatestComment(
+    client: TodoistApi,
+    target: CommentTarget,
+): Promise<Comment | undefined> {
+    // The SDK brands the unused key as `never` on each half of
+    // `GetTaskCommentsArgs | GetProjectCommentsArgs`, which a generic cannot
+    // infer across, so pin one half and let the target supply either key.
+    const comments = await fetchAllPages<GetTaskCommentsArgs, GetCommentsResponse, Comment>({
```
doistbot: <a href="#">P2</a>  `fetchAllPages` materializes every comment in the thread solely for the `reduce` below to select one item. Long-lived tasks can therefore allocate an array for their entire comment history on each defaulted add-comment. Iterate the cursors here and retain only the latest comment seen, which keeps the required unordered-page scan but uses constant memory.

doistbot (commented):
This PR adds per-comment `notifyUsers` to `add-comments`, resolving user refs through the existing `userResolver` and defaulting to the Todoist clients' recipient logic (assignee, assigner, creator on first comment; previous participants on replies) when the field is omitted.

Few things worth tightening:

- **Cap and dedupe `notifyUsers` before resolving.** The array has no `.max(...)` bound and `resolveUserRefs` fans out every ref via `Promise.all` *before* deduping, so duplicates or unresolvable names each trigger `getUser` plus full collaborator pagination against the operator's token — a single untrusted argument can amplify into unbounded parallel SDK calls.
- **Load shared collaborator data once per list, not per ref.** Because each non-ID ref independently misses the user/collaborator caches before any concurrent call populates them, a comment naming several people re-runs `getProjects` and every shared project's collaborator lookup once per name. Resolve the current user and collaborators up front, then match each ref against that pre-loaded set.
- **Reject or normalize `notifyUsers: []`.** An empty array is truthy in `resolveRecipientsPerComment`, so it bypasses the default-recipient path and posts silently — contradicting the documented `"none"` opt-out and risking an accidental notification-chain break. Add `.min(1)` or explicitly treat `[]` as "use defaults."
- **Stream the latest-comment lookup instead of materializing the full thread.** `fetchAllPages` loads every comment in the thread just for the `reduce` to pick one item, which can be a large allocation on long-lived tasks. Iterate the cursors and retain only the latest comment seen to keep the scan constant-memory.

<details>

- **<a href="#">P3</a> src/mcp-server.ts:47**: This bullet duplicates the `notifyUsers` parameter guidance that is already fully described in `add-comments`' own input schema. The instructions block (and AGENTS.md) says to reserve this section for cross-tool routing guidance — the next bullet about add-vs-edit is that — and to keep per-tool parameter detail in the tool description. Consider keeping only the routing-relevant sentence and dropping the `notifyUsers`/`["none"]` instructions from here to avoid drift and duplicated token cost.
- **<a href="#">P3</a> src/tools/add-comments.ts:152**: `if (!target || !currentUser) return []` silently degrades to "notify nobody" — the exact failure this PR is meant to eliminate. Both conditions are already guaranteed by the caller: `needsDefaultRecipients` fetches `currentUser` whenever any comment omits `notifyUsers`, and `targets` is built from the same `comments` array, so `target` is always present in the default branch. Prefer throwing an invariant error here so a future refactor fails loudly instead of quietly dropping notifications.
- **<a href="#">P3</a> src/tools/add-comments.test.ts:526**: The first half of "should surface who was notified" posts with `notifyUsers: ['none']` (line 529) while the `addComment` mock returns `uidsToNotify: ['111', '222']`, so the `notifiedUserIds: ['111', '222']` assertion passes only because `mapComment` copies the SDK field back. It can't fail if the tool misroutes recipients, and the `['none']` input vs. two-user output reads as a bug. Use a coherent scenario — e.g. `notifyUsers: ['111', '222']` with `resolveUserRefs` mocked to return those two IDs — so the assertion reflects the tool's own recipient computation. The "none" → omitted-field behavior is already covered separately.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F580&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A07094580-9af1-11f1-9f3b-27064837c979%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A07094580-9af1-11f1-9f3b-27064837c979&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/utils/user-resolver.ts
scottlovegrove: Fixed in fa0d4a7. `resolveUserRefs` now deduplicates references and resolves them one at a time, so the first lookup populates the collaborator cache the rest read instead of every name racing its own `getProjects` + per-project collaborator fetch. Covered by two tests asserting `getProjects`/`getProjectCollaborators` are each called once across a multi-name list and a repeated one.

src/tools/add-comments.ts
scottlovegrove: Fixed in fa0d4a7. Capped with `.max(ApiLimits.NOTIFY_USERS_MAX)` (25, added alongside the other comment limits), and `resolveUserRefs` now dedupes before resolving rather than after — combined with the sequential resolution in the other thread, a list of N duplicates or N unresolvable names is one lookup, not N concurrent ones.

scottlovegrove: Fixed in fa0d4a7 with `.min(1)`. You're right that this was an undocumented second opt-out — `["none"]` was chosen over `[]` deliberately when scoping this, so accepting both was an oversight rather than an intent. Rejecting the empty list keeps `["none"]` the single documented way to stay silent.

src/utils/comment-recipients.ts
scottlovegrove: Fixed in fa0d4a7. `getLatestComment` now walks the cursors keeping only the newest comment seen, so memory is constant regardless of thread length, and it still compares every page rather than trusting the order.

Worth noting for anyone reading this later: the request count can't be improved the same way. I checked against the live API and comments come back oldest-first with no reverse option, so the newest is always on the last page and reaching it means paging through the thread either way.

PR #581, opened 2026-08-19
feat(filters): expose filter descriptions in the filter tools

craigcarlyle:
# Pull Request

## Short description

Filters are getting descriptions, so the filter tools can read and write them. `add-filters` sets one, `update-filters` changes or clears it, `find-filters` returns it.

- Clearing uses `"remove"` rather than null, matching how `update-tasks` clears a due date. The schema lint rejects nullable optional strings because they break the Gemini API.
- `readFilterDescription` is scaffolding. Delete it once the SDK PR lands and the field is properly typed.
- Blocked on Doist/todoist-sdk-typescript#668.

Backend: Doist/Todoist#29669

## PR Checklist

No new tools, so no registration or export changes.

### DIFF
--- a/src/tools/add-filters.ts
+++ b/src/tools/add-filters.ts
@@ -2,6 +2,7 @@ import { type ColorKey, createCommand } from '@doist/todoist-sdk'
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { ColorOutputSchema, ColorSchema } from '../utils/colors.js'
+import { optionalString } from '../utils/schema-helpers.js'
 import { ToolNames } from '../utils/tool-names.js'
 import { FilterOutputSchema } from './find-filters.js'
 
@@ -14,10 +15,7 @@ const FilterSchema = z.object({
             'The filter query string. Examples: "today & p1", "#Work & overdue", "@email & today", "(p1 | p2) & !assigned". ' +
                 'Operators: | (OR), & (AND), ! (NOT), () grouping, , (multiple queries).',
         ),
-    description: z
-        .string()
-        .optional()
-        .describe('A markdown description explaining what the filter is for.'),
+    description: optionalString('A markdown description explaining what the filter is for.'),
     color: ColorSchema,
     isFavorite: z
         .boolean()
--- a/src/tools/add-filters.ts
+++ b/src/tools/add-filters.ts
@@ -14,6 +14,10 @@ const FilterSchema = z.object({
             'The filter query string. Examples: "today & p1", "#Work & overdue", "@email & today", "(p1 | p2) & !assigned". ' +
                 'Operators: | (OR), & (AND), ! (NOT), () grouping, , (multiple queries).',
         ),
+    description: z
+        .string()
+        .optional()
+        .describe('A markdown description explaining what the filter is for.'),
     color: ColorSchema,
     isFavorite: z
         .boolean()
@@ -49,6 +53,9 @@ const addFilters = {
                 {
                     name: filter.name,
                     query: filter.query,
+                    ...(filter.description !== undefined
+                        ? { description: filter.description }
+                        : {}),
                     ...(filter.color !== undefined ? { color: filter.color as ColorKey } : {}),
                     ...(filter.isFavorite !== undefined ? { isFavorite: filter.isFavorite } : {}),
                 },
@@ -73,6 +80,10 @@ const addFilters = {
                     id: realId,
                     name: filter.name,
                     query: filter.query,
+                    // Omitted rather than null when absent: structured content is sanitised
+                    // with `removeNullFields`, so a null would be stripped and then fail
+                    // validation against the declared output schema.
+                    ...(filter.description ? { description: filter.description } : {}),
                     color: outputColor,
                     isFavorite: filter.isFavorite ?? false,
                     itemOrder: 0,
--- a/src/tools/find-filters.ts
+++ b/src/tools/find-filters.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod'
 import type { TodoistTool } from '../todoist-tool.js'
 import { ColorOutputSchema } from '../utils/colors.js'
+import { readFilterDescription } from '../utils/filter-description.js'
 import { ToolNames } from '../utils/tool-names.js'
 
 const ArgsSchema = {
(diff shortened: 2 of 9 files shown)


src/tools/find-filters.ts
```diff
@@ -48,6 +55,7 @@ const findFilters = {
             id: f.id,
             name: f.name,
             query: f.query,
+            description: readFilterDescription(f),
```
doistbot, line 58: <a href="#">P2</a>  Only update-filters.test.ts gained coverage for the new field. Per the repo's testing requirements for new tool parameters, the other two paths are untested: add-filters.test.ts has no test that a passed description is forwarded in the filter_add command and echoed in structuredContent (including the absent case at line 86), and find-filters.test.ts never exercises readFilterDescription — e.g. a mock filter with description: '' should come back with the key omitted, while a non-empty one is returned. A small test in each file would cover the sentinel's read path and the empty-string normalization.

src/tools/update-filters.ts
```diff
@@ -15,6 +16,12 @@ const FilterUpdateSchema = z.object({
         .describe(
             'The new filter query string. Examples: "today & p1", "#Work & overdue", "@email & today".',
         ),
+    description: z
+        .string()
```
doistbot: <a href="#">P2</a>  Accept legacy `null` here by preprocessing it to `"remove"`, as `update-tasks` does. The current string schema rejects `null` before the command-building logic runs, so existing programmatic clients that clear optional fields with `null` cannot clear a filter description.

src/tools/add-filters.ts
```diff
@@ -14,6 +14,10 @@ const FilterSchema = z.object({
             'The filter query string. Examples: "today & p1", "#Work & overdue", "@email & today", "(p1 | p2) & !assigned". ' +
                 'Operators: | (OR), & (AND), ! (NOT), () grouping, , (multiple queries).',
         ),
+    description: z
```
doistbot: <a href="#">P2</a>  Use the existing `optionalString` helper from `../utils/schema-helpers.js` instead of raw `z.string().optional()`. Its documented purpose is exactly this case: LLMs often send `""` for omitted optional string fields, and forwarding that empty value to the Todoist API can fail the request. `optionalString('A markdown description explaining what the filter is for.')` gives the same `string | undefined` type and normalizes `""` to `undefined`, so the existing `filter.description !== undefined` spread omits it.

src/tools/add-filters.ts
```diff
@@ -49,6 +53,9 @@ const addFilters = {
                 {
                     name: filter.name,
                     query: filter.query,
+                    ...(filter.description !== undefined
```
doistbot, line 54: <a href="#">P2</a>  The repo's testing requirements ask for setting-value coverage when adding a new tool parameter, but add-filters has no test for a description on create: nothing verifies description lands in the filter_add command args, that it's omitted from both the command and structuredContent when absent, or that an empty-string description is omitted from the output while still sent in the command. Mirror the update-filters tests: a set test asserting commands[0].args.description, and an omit test asserting the key is absent in both places.

doistbot (commented):
This PR wires filter descriptions through `add-filters`, `update-filters`, and `find-filters`, using the repo's standard `"remove"` sentinel and Gemini-safe optional schemas — the scaffolding `readFilterDescription` shim is nicely documented until the SDK field is typed.

Few things worth tightening:
- Test coverage only landed for `update-filters`; `add-filters` needs set/omit tests for the new field, and `find-filters` should exercise `readFilterDescription` (empty string omitted, non-empty returned).
- `update-filters` rejects legacy `null` before the command-building logic runs — preprocess it to `"remove"` like `update-tasks` does, so existing programmatic clients can still clear the field.
- Use the existing `optionalString` helper from `schema-helpers.ts` instead of hand-rolled `z.string().optional()`; it already normalizes the empty strings LLMs send for omitted optional fields.

<details>

- **<a href="#">P3</a> src/utils/filter-description.ts:13**: The doc comment says the API clears the description on null, but update-filters now also accepts "remove" as the clearing sentinel and translates it to null before it reaches the API. Once the SDK PR lands and this helper is deleted, a reader has no in-repo note that "remove" is the clearing sentinel for descriptions (only the input schema description carries it). Either mention "remove" in the comment or drop the now-wrong framing once the field is properly typed.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F581&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab5bd69b0-a714-11f1-9f3d-0bedf15f0b9d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab5bd69b0-a714-11f1-9f3d-0bedf15f0b9d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/find-filters.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-mcp/commit/be0fd44

src/tools/update-filters.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-mcp/commit/be0fd44

src/tools/add-filters.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-mcp/commit/be0fd44

src/tools/add-filters.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-mcp/commit/be0fd44

src/utils/filter-description.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Normalizes a filter's description for structured output.
+ *
+ * A filter with no description can arrive as either `null` or `""`: the column is
```
engfragui, line 4: > A filter with no description can arrive as either `null` or `""`

Why is it the case? Is this intentional that it can arrive either way? 🤔

engfragui: Found some more details in the find-filters test:

> // The column is nullable and the sync view normalizes NULL to "" on the way out, so a
> // filter with no description arrives as either, and a payload predating the field has
> // no key at all. All three have to surface as undefined rather than null: structured
> // content is sanitised with removeNullFields, so a null would be stripped and then
> // fail validation against the optional output schema.

I guess I'll accept it 🤝

PR #585, opened 2026-08-28
fix: log the batch items a tool could not act on

scottlovegrove:
## What

Seven batch tools collect per-item failures into a `failures` array and return them in the tool result rather than throwing. `registerTool` only logs when a tool throws, so these failures never reach the logs: the server answers 200 and records nothing while the caller's items stay untouched.

`logBatchFailures` records the failure count, the batch size and a capped sample, and every one of the seven now calls it: `complete-tasks`, `uncomplete-tasks`, `update-tasks`, `add-tasks`, `add-projects`, `update-projects`, `add-sections`.

## Why

`complete-tasks` failing on all 50 ids is currently indistinguishable in the logs from one that completed all 50. `add-tasks` is only partly covered: it throws when every task fails, which is where the `All 4 task(s) failed to create: ... MAX_ITEMS_LIMIT_REACHED` events in Datadog come from, but three of four failing the same way is silent.

This came out of a Comms incident where the equivalent gap in `@doist/comms-mcp` left an automation failing with `markRead: fetch failed; archive: fetch failed` while Datadog held nothing but three successful 200s for the same call (Doist/comms-mcp#50 fixed that side).

A failed batch now leaves:

```
complete-tasks: items failed { failed: 50, of: 50,
  sample: [{ item: '6X4Vw2Hfmg73Q2XR', error: 'HTTP 404: Not Found' }, ...] }
```

## Notes

- The sample is capped by a new `LogLimits.MAX_FAILURES_LOGGED`. It is deliberately not `DisplayLimits.MAX_FAILURES_SHOWN`, which governs how many failures the model is shown rather than how many the server records.
- The sample keeps the `code` a tool sets where there is one, so `update-tasks` reporting `PARTIAL_MOVE_APPLIED` stays distinguishable from a task that did nothing at all.
- A batch is as large as the caller's input, so one outage logs one line with a sample, not one line per item.
- `failures[].item` is the task content for the `add-*` tools, so titles reach the logs. That matches what the repo already does: the `registerTool` error log records the full `args`, and the all-failed `add-tasks` throw embeds the contents in its message. Happy to switch those three to an index if you would rather not widen it.

## Testing

`npm test` (1217 passing, 76 files), `npx tsc --noEmit` and `npm run format:check` all pass locally. Four tests cover the helper (counts, the `code` field, a 200-item batch capped to one line, silence on full success) and two cover a real call site in `complete-tasks`.

https://claude.ai/code/session_01T1D2UgUZiqouii9bXebRVS

### DIFF
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -9,6 +9,7 @@ import {
 } from '../tool-execution-error.js'
 import { isInboxProjectId, mapTask } from '../tool-helpers.js'
 import { assignmentValidator } from '../utils/assignment-validator.js'
+import { logBatchFailures } from '../utils/batch-failures.js'
 import { BatchLimits } from '../utils/constants.js'
 import { DurationParseError, parseDuration } from '../utils/duration-parser.js'
 import { FailureSchema, TaskSchema as TaskOutputSchema } from '../utils/output-schemas.js'
@@ -162,6 +163,8 @@ const addTasks = {
         }
 
         const mappedTasks = newTasks.map(mapTask)
+
+        logBatchFailures(ToolNames.ADD_TASKS, tasks.length, failures)
 
         const textContent = generateTextContent({
             tasks: mappedTasks,
--- a/src/tools/add-tasks.ts
+++ b/src/tools/add-tasks.ts
@@ -164,7 +164,7 @@ const addTasks = {
 
         const mappedTasks = newTasks.map(mapTask)
 
-        logBatchFailures(ToolNames.ADD_TASKS, tasks.length, failures)
+        logBatchFailures(ToolNames.ADD_TASKS, tasks.length, failures, { redactItems: true })
 
         const textContent = generateTextContent({
             tasks: mappedTasks,
--- a/src/tools/complete-tasks.test.ts
+++ b/src/tools/complete-tasks.test.ts
@@ -290,4 +290,37 @@ describe(`${COMPLETE_TASKS} tool`, () => {
             expect(result.textContent).toMatchSnapshot()
         })
     })
+
+    describe('logging', () => {
+        it('logs the tasks it could not complete', async () => {
+            const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+            mockTodoistApi.closeTask.mockImplementation(async (id: string) => {
+                if (id === 'task-2') {
+                    throw new Error('HTTP 404: Not Found')
+                }
+                return true
+            })
+
+            await completeTasks.execute({ ids: ['task-1', 'task-2'] }, mockTodoistApi)
+
+            expect(consoleErrorSpy).toHaveBeenCalledWith(`${COMPLETE_TASKS}: items failed`, {
+                failed: 1,
+                of: 2,
+                sample: [{ item: 'task-2', error: 'HTTP 404: Not Found' }],
+            })
+
+            consoleErrorSpy.mockRestore()
+        })
+
+        it('stays quiet when every task completes', async () => {
+            const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+            mockTodoistApi.closeTask.mockResolvedValue(true)
+
+            await completeTasks.execute({ ids: ['task-1'] }, mockTodoistApi)
+
+            expect(consoleErrorSpy).not.toHaveBeenCalled()
+
+            consoleErrorSpy.mockRestore()
+        })
+    })
 })
--- a/src/tools/complete-tasks.test.ts
+++ b/src/tools/complete-tasks.test.ts
@@ -290,37 +290,4 @@ describe(`${COMPLETE_TASKS} tool`, () => {
             expect(result.textContent).toMatchSnapshot()
         })
     })
-
-    describe('logging', () => {
-        it('logs the tasks it could not complete', async () => {
-            const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
-            mockTodoistApi.closeTask.mockImplementation(async (id: string) => {
-                if (id === 'task-2') {
-                    throw new Error('HTTP 404: Not Found')
-                }
-                return true
-            })
-
-            await completeTasks.execute({ ids: ['task-1', 'task-2'] }, mockTodoistApi)
-
-            expect(consoleErrorSpy).toHaveBeenCalledWith(`${COMPLETE_TASKS}: items failed`, {
-                failed: 1,
-                of: 2,
-                sample: [{ item: 'task-2', error: 'HTTP 404: Not Found' }],
-            })
-
-            consoleErrorSpy.mockRestore()
-        })
-
-        it('stays quiet when every task completes', async () => {
-            const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
-            mockTodoistApi.closeTask.mockResolvedValue(true)
-
-            await completeTasks.execute({ ids: ['task-1'] }, mockTodoistApi)
-
-            expect(consoleErrorSpy).not.toHaveBeenCalled()
-
-            consoleErrorSpy.mockRestore()
-        })
-    })
 })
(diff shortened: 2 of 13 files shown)


src/tools/add-tasks.ts
```diff
@@ -163,6 +164,8 @@ const addTasks = {
 
         const mappedTasks = newTasks.map(mapTask)
 
+        logBatchFailures(ToolNames.ADD_TASKS, tasks.length, failures)
```
doistbot: <a href="#">P2</a>  For the `add-*` tools, `failures[].item` is the caller's content (task title, and project/section name in `add-projects`/`add-sections`), so this new call writes user-authored text into server logs on partial failures. `mcp-helpers.ts` states the opposite intent — its `REDACTED_ARG_KEYS` comment says values that "carry user content rather than identifiers ... must not reach server logs". Task titles can contain PII, and this is a second, unredacted path for that content to reach Datadog (the throw path already logs it, but partial failures previously did not). Consider logging the array index for these three tools instead — `task/project/section ${index + 1}` is already produced as the fallback `item` — or redacting `item` before writing. The `item` field in the structured tool result stays unchanged; only the log sample would switch to a positional index.

src/tools/complete-tasks.ts
```diff
@@ -39,6 +40,8 @@ const completeTasks = {
             }
         }
 
+        logBatchFailures(ToolNames.COMPLETE_TASKS, args.ids.length, failures)
```
doistbot, line 46: <a href="#">P2</a>  The failures logged here (and the same call in uncomplete-tasks.ts:43) come from `catch` blocks that push `error.message`, so the new log line records messages like a bare `HTTP 403: Forbidden` with no status code or `error_tag`. Every other batch tool builds its failures with `formatBatchItemError` precisely to keep signals like `MAX_ITEMS_LIMIT_REACHED` recognizable after aggregation — and the PR's whole purpose is diagnosing these failures from logs. An outage or permission rejection on complete/uncomplete-tasks will now log five indistinguishable generic HTTP messages. Consider switching both tools' catch blocks to `formatBatchItemError(result)`, matching add-tasks/add-projects/add-sections/update-*.

src/tools/complete-tasks.test.ts
```diff
@@ -290,4 +290,37 @@ describe(`${COMPLETE_TASKS} tool`, () => {
             expect(result.textContent).toMatchSnapshot()
         })
     })
+
+    describe('logging', () => {
```
doistbot: <a href="#">P2</a>  The logging coverage only exercises one of the seven call sites. If `add-tasks` or `update-projects` loses its `logBatchFailures` call in a refactor, nothing fails — yet that wiring is the entire point of the PR. The two call-site tests here largely duplicate what `batch-failures.test.ts` already covers (log shape and silence), so consider shrinking this to a single wiring test and adding a small parameterized test that runs each of the seven batch tools against a mocked failing API and asserts one error log per call. That covers all the call sites once, cheaply.

doistbot (commented):
This PR adds a `logBatchFailures` helper wired into all seven batch tools, so per-item failures that are returned rather than thrown finally reach the server logs as one capped-sample line per call. Few things worth tightening:
- The `add-*` tools log the caller's task/project/section titles, which conflicts with the repo's stated redaction stance for user content in `mcp-helpers.ts` — consider logging the positional index instead (the fallback `item` already exists) or redacting `item` in the log sample only.
- `complete-tasks` and `uncomplete-tasks` push bare `error.message` from their catch blocks, so their log lines lose the status codes and error tags (e.g. `MAX_ITEMS_LIMIT_REACHED`) the other five tools preserve via `formatBatchItemError` — switching those two would make the logs this PR adds actually diagnosable.
- Logging coverage exercises only one call site; a small parameterized test asserting one error log per call across all seven tools would catch a lost `logBatchFailures` call in a future refactor, and the two existing call-site tests largely duplicate `batch-failures.test.ts`.
- Also worth a follow-up: document the new util in CODEBASE.md per the repo's catalog trigger rules.

<details>

- **<a href="#">P3</a> src/utils/batch-failures.ts:17**: Per the repo's CODEBASE.md rules, adding a broadly-reusable `src/utils/` helper (this one has seven callers) and a new constants group (`LogLimits`) are triggers to update CODEBASE.md — its utils catalog doesn't list `batch-failures.ts` and the `constants.ts` entry doesn't mention `LogLimits`. Worth a one-line addition so the 'don't reimplement' map stays accurate.
- **<a href="#">P3</a> src/utils/batch-failures.ts:3**: `BatchFailure` re-declares the failure shape that `FailureSchema` in src/utils/output-schemas.ts already encodes (`{ item, error, code? }`). Derive the type from the schema instead — `type BatchFailure = z.infer<typeof FailureSchema>` — so the logged sample shape stays in sync with the output schema (e.g. if `code` or another field changes). All seven tool files already import `FailureSchema`, so this avoids a second hand-maintained copy of the same contract.
- **<a href="#">P3</a> src/utils/batch-failures.ts:18**: `toolName` is an open `string`, but every caller passes a value from the `ToolNames` constant. Type it as `ToolName` (imported from `./tool-names.js`) so a typo or an unregistered name is caught at compile time, matching the repo's avoid-hardcoded-strings convention.
- **<a href="#">P3</a> src/tools/complete-tasks.test.ts:312**: The spy is created and restored inside each test body, so a failed `expect` before line 312 leaves `console.error` mocked for every remaining test in the file, which can mask later failures. Move the spy setup and restore into `beforeEach`/`afterEach` in the `logging` block, matching what `batch-failures.test.ts` does.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F585&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A10fe0620-a2cb-11f1-97b6-cec1c07fe5bc%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A10fe0620-a2cb-11f1-97b6-cec1c07fe5bc&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tools/add-tasks.ts
scottlovegrove: Done in 3f494f0. `logBatchFailures` now takes a `redactItems` option, and the three `add-*` tools pass it, so the log sample carries `item: '[redacted]'` while the structured result keeps the real label. I went with the redaction token rather than a positional index because the index would have to be threaded through the failures array to be the batch position rather than the position within the failures list, and the error plus its code is what you diagnose from.

src/tools/complete-tasks.ts
scottlovegrove: Done in 3f494f0. Both catch blocks now build the failure with `formatBatchItemError`, so the status and error tag survive into both the result text and the new log line.

src/tools/complete-tasks.test.ts
scottlovegrove: Done in 3f494f0. Added `src/tools/batch-failure-logging.test.ts`, one parameterized case per call site, asserting exactly one log per call with the expected `of` count and item. The `add-tasks` case is a partial failure since the all-failed path throws instead. The two duplicated tests here are removed; the shape and silence cases stay in `batch-failures.test.ts`.

PR #597, opened 2026-09-04
perf: cache MCP tool schemas

gnapse:
## Short description

The hosted integration creates an isolated MCP server for each request. Registering its 47 tools previously converted the same static Zod schemas to JSON Schema each time. This repeated a small amount of CPU work and created short-lived objects on every request.

This change compiles each tool's input and output schema once per process and reuses that immutable representation when later server instances register the tool. The cache applies only to the draft 2020-12 conversion with default options, which is what the current MCP SDK requests. Other JSON Schema targets or custom library options still delegate to Zod, so a future SDK change cannot receive a cached schema in the wrong dialect.

Requests still receive separate MCP servers and request-scoped clients, so user state, validation, and concurrent requests remain isolated. Validation continues to use the original Zod schema.

A local benchmark measured registration at about 3.4 ms before this change and 0.2 ms after it, with roughly 120 KB less short-lived JSON data per server creation. This is a small cleanup of repeated work. It is not expected by itself to produce a visible latency change or explain the earlier incident.

## Verification

- Confirms that repeated registrations reuse separate cached input and output conversions.
- Confirms that other schema dialects and custom library options delegate to Zod.
- Confirms through a real MCP `tools/list` request that every advertised tool schema matches direct Zod conversion.
- Full test suite: 1,264 tests passed.

## PR Checklist

### DIFF
--- a/src/mcp-helpers.test.ts
+++ b/src/mcp-helpers.test.ts
@@ -157,6 +157,40 @@ describe('registerTool config', () => {
             value: { limit: 10 },
         })
     })
+
+    it('preserves the uncached side of piped input and output schemas', () => {
+        const { mock, server, client } = captureRegisterToolMock()
+        const pipedDate = z.string().pipe(z.iso.date())
+        const tool = buildToolFixture({
+            name: 'piped-schema-tool',
+            description: 'Tool with different input and output schemas',
+            parameters: { date: pipedDate },
+            outputSchema: { date: pipedDate },
+            execute: async () => ({ textContent: 'ok' }),
+        })
+
+        registerTool({ tool, server, client })
+
+        type SchemaConfig = {
+            inputSchema: StandardSchemaWithJSON<unknown, unknown>
+            outputSchema: StandardSchemaWithJSON<unknown, unknown>
+        }
+        const config = mock.mock.calls[0]?.[1] as SchemaConfig
+        const options = { target: 'draft-2020-12' as const }
+
+        expect(config.inputSchema['~standard'].jsonSchema.input(options)).not.toMatchObject({
+            properties: { date: { format: 'date' } },
+        })
+        expect(config.inputSchema['~standard'].jsonSchema.output(options)).toMatchObject({
+            properties: { date: { format: 'date' } },
+        })
+        expect(config.outputSchema['~standard'].jsonSchema.input(options)).not.toMatchObject({
+            properties: { date: { format: 'date' } },
+        })
+        expect(config.outputSchema['~standard'].jsonSchema.output(options)).toMatchObject({
+            properties: { date: { format: 'date' } },
+        })
+    })
 })
 
 describe('registerTool error path', () => {
--- a/src/mcp-helpers.test.ts
+++ b/src/mcp-helpers.test.ts
@@ -191,6 +191,36 @@ describe('registerTool config', () => {
             properties: { date: { format: 'date' } },
         })
     })
+
+    it('delegates conversions that do not match the cached options', () => {
+        const { mock, server, client } = captureRegisterToolMock()
+        const parameters = { tuple: z.tuple([z.string(), z.number()]) }
+        const tool = buildToolFixture({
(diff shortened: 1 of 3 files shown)


doistbot (commented):
This PR caches each tool's JSON Schema compilation per process (via a WeakMap keyed on the static tool objects) so per-request MCP server instances reuse the compiled schemas instead of redoing the same Zod-to-JSON-Schema conversion, cutting repeated CPU work and temporary memory pressure while keeping request isolation and per-request validation intact. No inline issues were flagged; reviewers noted the caching uses the same io/target options the SDK itself relies on and introduces no security or reuse concerns.

<details>

- **<a href="#">P3</a> src/mcp-helpers.ts:89**: `withCachedJsonSchema` hand-builds `version`, `vendor`, and `validate`, which are already provided by `schema['~standard']` (the previous code could pass raw Zod objects to the SDK because Zod already satisfies `StandardSchemaWithJSON`). Spread the existing standard object and override only `jsonSchema` to return the cached JSON: `'~standard': { ...standard, jsonSchema: { input: () => jsonSchema, output: () => jsonSchema } }`. That keeps vendor metadata and validation semantics in sync with Zod instead of duplicating them.
- **<a href="#">P3</a> src/mcp-helpers.ts:86**: These serialization options are now the third place in the repo that must stay in step with how the SDK serializes Standard Schemas (token-footprint.test.ts even carries a "Keep these in step with the SDK" comment for its copies). Since this compiled schema is now what actually goes on the wire, drift between this hardcode and the SDK's expectations would silently produce a wrong tools/list schema while token-footprint keeps measuring the old shape. Worth hoisting a shared constant into a non-test util so the cache, the footprint measurement, and the eval harness can't diverge.
- **<a href="#">P3</a> src/mcp-helpers.ts:94**: Both `input()` and `output()` return the same compiled JSON Schema, but for piped fields the two sides differ — e.g. find-activity's `optionalString(description).pipe(IsoDateOrDateTime.optional())` compiles to `string` on the input side and a date/datetime union on the output side. The input wrapper is compiled with `io: 'input'`, so its `output()` accessor advertises the input side, and vice versa for the output wrapper. This is fine as long as the SDK only calls the accessor matching each schema's role, but it's a latent trap; a short comment noting that `input()`/`output()` deliberately alias the role-specific compile would make that assumption explicit.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F597&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A227977f0-a897-11f1-9b6d-cf6741e05659%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A227977f0-a897-11f1-9b6d-cf6741e05659&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

goncalossilva (approved):

scottlovegrove (commented):
Nice! We should port this to the Comms MCP too, right?

Issue #596, opened 2026-09-04
[bug] Hosted server ai.todoist.net/mcp returning 504 / timing out since 2026-09-04 14:48 UTC (right after the 13.2.1 deploy)


santiagomalter:
## Summary

The hosted server at `https://ai.todoist.net/mcp` has been returning **HTTP 504 Gateway Time-out** (served by the nginx edge, ~10 s) or hanging past 20 s for the large majority of requests since **2026-09-04 14:48 UTC**, still ongoing at 17:10 UTC. A minority of requests still succeed, so the service flaps rather than being fully down.

The first failure lands ~10 minutes after the `v13.2.1` release pipeline completed (14:38 UTC), whose last step dispatches `todoist-ai-updated` to the integrations repo — so this looks like the hosted deploy of 13.2.1 rather than the change in #595 itself (which is a benign cursor normalization). Filing here because it is the public repo for the hosted server; happy to be redirected.

## Timeline (UTC, 2026-09-04)

| Time | Event |
|---|---|
| 14:35:31 | #595 merged (`c1e77e7`) |
| 14:35:34 → 14:38:08 | `Release` workflow: `v13.2.1` published, then `repository_dispatch todoist-ai-updated` (version 13.2.1) |
| 14:41:50 | OAuth refresh against `todoist.com/oauth/access_token` → 200 (auth is healthy throughout) |
| **14:48:00** | **First 504 from `ai.todoist.net/mcp`** |
| 14:48 → 17:10+ | Flapping: in any 10-minute window roughly 5–9 × 504 vs 1–3 × 200; occasional 202 |

Baseline for the 12 hours before 14:48: 630 × 200, 31 × 202, **0 × 504, 0 × 401** on the same client.

## Reproduction (no auth needed — the edge itself fails)

```bash
curl -s -m 25 -o /dev/null -w '%{http_code} %{time_total}s\n' \
  -X POST https://ai.todoist.net/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}'
```

Observed across ~2 hours, from two independent networks (EU residential and a DigitalOcean droplet):

```
504 10.8s
000 20.0s   (client timeout)
504 10.7s
401 2.0s    (occasional healthy response)
000 12.0s
```

[...]

PR #598, opened 2026-09-04
fix: bound per-account concurrency limiters

gnapse:
Related to #596

## Short description

The MCP server creates a new `TodoistApi` client for each HTTP request. To keep task updates safe, clients for the same Todoist account share one pair of concurrency limiters. However, the server previously retained every account's limiter pair for the lifetime of the process. Memory could therefore grow with every distinct account that reached a long-running server instance.

This change makes that registry bounded:

- It removes a per-account limiter entry after five minutes of inactivity, but only when the limiter has no running or queued work.
- It retains at most 10,000 per-account limiter entries per process.
- At capacity, it evicts the least-recently-used idle entry and gives the new account a dedicated limiter pair.
- It uses one shared overflow limiter pair only when every retained entry has running or queued work and none is safe to evict.

The request is not silently dropped at capacity. In the usual capacity case, it replaces an idle entry and runs through its own dedicated limiter pair. If all retained entries are busy, a task move or field update uses the shared overflow pair. It runs immediately when a concurrency slot is available, or waits in the existing FIFO queue. If it cannot get a slot within 30 seconds, it returns an explicit error and does not send the Todoist API request.

## Before and after

| Scenario | Before | After |
| --- | --- | --- |
| A request arrives for an account whose dedicated limiter pair already exists | Reuse that account's limiter pair | Unchanged: reuse it and mark it as most recently used |
| A dedicated limiter pair has no running or queued work for five minutes | Retain it until the process exits | Remove it when a later request triggers cleanup |
| A limiter pair has running or queued work | Retain it | Unchanged: never evict it while work remains |
| A request arrives for a new account while the registry is below capacity | Create and retain a dedicated limiter pair | Unchanged: create and retain a dedicated pair |
| A request arrives for a new account at capacity and at least one retained pair is idle | Create and retain entry 10,001, then continue growing | Evict the least-recently-used idle pair and create a dedicated pair for the new account |
| A request arrives for a new account at capacity and all retained pairs are busy | Create and retain entry 10,001, then continue growing | Use the shared overflow pair and keep the registry at 10,000 entries |
| A task move uses the shared overflow pair | No overflow path; every account added permanent state | Run immediately if the shared move lane is free, otherwise wait; one move runs at a time |
| A task field update uses the shared overflow pair | No overflow path; every account added permanent state | Run immediately if a shared write slot is free, otherwise wait; up to four writes run at a time |
| Queued work cannot start within 30 seconds | Return the existing queue-timeout error without sending the API request | Unchanged; this also protects the shared overflow path |

The shared overflow limiter applies collectively only when all retained account entries in that MCP server process are busy. This can add latency during an extreme burst, but it keeps both retained memory and concurrent API calls bounded. An account stays on the overflow pair until all of its current overflow work finishes, so concurrent calls for that account cannot split across different limiter pairs.

## Request flow

```mermaid
flowchart TD
    A[Receive authenticated MCP request and<br/>create request-scoped Todoist client] --> B[Compute an account key<br/>from the API token]
    B --> C{Does this account already have<br/>a dedicated limiter pair?}
    C -- Yes --> D[Reuse the dedicated pair<br/>and mark it as most recently used]
    C -- No --> E[Remove entries idle for 5 minutes<br/>with no running or queued work]
    E --> F{Registry below<br/>10,000 entries?}
    F -- Yes --> G[Create a dedicated limiter pair<br/>for this account]
    F -- No --> H{Is any retained<br/>limiter pair idle?}
    H -- Yes --> I[Evict the least-recently-used idle pair<br/>and create a dedicated pair]
    H -- No --> J[All retained pairs are busy<br/>Use the shared overflow pair]
    D --> K{Which internal Todoist API<br/>operation is about to run?}
    G --> K
    I --> K
    J --> K
    K -- Other operation --> L[Continue without these<br/>concurrency limiters]
    K -- Task move --> M[Enter move lane<br/>1 operation at a time]
    K -- Task field update --> N[Enter write lane<br/>Up to 4 operations at a time]
    M --> O{Concurrency slot available?}
    N --> O
    O -- Yes --> P[Send request to Todoist API]
    O -- No --> Q[Wait in FIFO queue<br/>for up to 30 seconds]
    Q -- Slot opens --> P
    Q -- 30 seconds elapse --> R[Return an explicit queue-timeout error<br/>Todoist API request was not sent]
    P --> S[Return the Todoist API result or error<br/>to the MCP tool]
```

The concurrency lanes currently apply to task moves and task field updates. Read-only tool calls are not delayed by these limiters.

## Safety properties

- The per-account limiter registry cannot contain more than 10,000 entries.
- The registry stores a SHA-256 digest, not the raw API token.
- A limiter pair with running or queued work is never evicted.
- At capacity, the least-recently-used idle pair is evicted before overflow is used.
- When every retained pair is busy, all affected request-scoped clients share one overflow pair, so creating a client per request cannot bypass the concurrency limits.
- An account stays assigned to overflow until all of its current overflow work finishes.
- Request-scoped client associations use a `WeakMap`, so they do not keep completed request clients alive.

## Verification

- Confirms that capacity pressure evicts an idle entry before using overflow.
- Confirms that a client which selected a limiter before eviction resolves the account's current pair when work starts.
- Confirms with two distinct account tokens that overflow clients share the concurrency limits when every retained entry is busy.
- Confirms that overflow clients remain separate from the fallback path for unregistered SDK clients.
- Confirms behaviorally that an expired entry with active work is not evicted under capacity pressure.
- Full test suite: 1,266 tests passed.

## PR Checklist

### DIFF
--- a/src/utils/concurrency.ts
+++ b/src/utils/concurrency.ts
@@ -4,7 +4,11 @@ import { ConcurrencyLimits } from './constants.js'
 /**
  * Wraps a task so it only runs once a slot is free.
  */
-type Limiter = <T>(fn: () => Promise<T>) => Promise<T>
+type Limiter = {
+    <T>(fn: () => Promise<T>): Promise<T>
+    /** Whether no task is running or waiting for this limiter. */
+    isIdle: () => boolean
+}
 
 type LimiterPair = {
     /** Serialises task-move requests, which contend on server-side tree locks. */
@@ -89,7 +93,7 @@ function createLimiter(
         active--
     }
 
-    return async function limit<T>(fn: () => Promise<T>): Promise<T> {
+    const limit = async function limit<T>(fn: () => Promise<T>): Promise<T> {
         await acquire()
         try {
             return await fn()
@@ -98,6 +102,8 @@ function createLimiter(
             release()
         }
     }
+    limit.isIdle = () => active === 0 && queue.length === 0
+    return limit
 }
 
 function createLimiterPair(): LimiterPair {
@@ -115,7 +121,14 @@ function createLimiterPair(): LimiterPair {
  * This map lives in one process, so the limits it enforces are per process rather
  * than per account globally. See the note on `ConcurrencyLimits`.
  */
-const limitersByAccount = new Map<string, LimiterPair>()
+type AccountLimiterEntry = {
+    limiters: LimiterPair
+    lastUsedAt: number
+}
+
+const MAX_ACCOUNT_LIMITERS = 10_000
+const ACCOUNT_LIMITER_IDLE_TTL_MS = 5 * 60_000
+const limitersByAccount = new Map<string, AccountLimiterEntry>()
 
 /** Lets a tool find its account's limiters from the client it was handed. */
 const limitersByClient = new WeakMap<object, LimiterPair>()
@@ -139,18 +152,50 @@ function accountKeyFromApiKey(apiKey: string): string {
     return createHash('sha256').update(apiKey).digest('hex')
 }
 
+function isLimiterPairIdle({ moves, writes }: LimiterPair): boolean {
+    return moves.isIdle() && writes.isIdle()
+}
+
+/**
+ * Removes expired idle entries from the oldest end of the insertion-ordered
+ * map. Active entries stay until their work finishes, even after they expire.
+ */
+function pruneIdleAccountLimiters(now: number): void {
+    for (const [key, entry] of limitersByAccount) {
+        if (now - entry.lastUsedAt < ACCOUNT_LIMITER_IDLE_TTL_MS) {
+            break
+        }
+        if (isLimiterPairIdle(entry.limiters)) {
+            limitersByAccount.delete(key)
+        }
+    }
+}
+
 /**
  * Associates a client with its account's limiters, sharing one pair across every
  * client built for the same account.
  */
 function registerClientLimiters(client: object, apiKey: string): void {
     const key = accountKeyFromApiKey(apiKey)
-    let limiters = limitersByAccount.get(key)
-    if (!limiters) {
-        limiters = createLimiterPair()
-        limitersByAccount.set(key, limiters)
+    const now = Date.now()
+    let entry = limitersByAccount.get(key)
+
+    if (entry) {
+        // Refresh insertion order so pruning can stop at the first fresh entry.
+        limitersByAccount.delete(key)
+        entry.lastUsedAt = now
+        limitersByAccount.set(key, entry)
+    } else {
+        pruneIdleAccountLimiters(now)
+        entry = { limiters: createLimiterPair(), lastUsedAt: now }
+
+        // Keep the registry bounded. If every slot belongs to a recently used
+        // account, the new client still gets local limits but is not retained.
+        if (limitersByAccount.size < MAX_ACCOUNT_LIMITERS) {
+            limitersByAccount.set(key, entry)
+        }
     }
-    limitersByClient.set(client, limiters)
+    limitersByClient.set(client, entry.limiters)
 }
 
 function getLimiters(client: object): LimiterPair {
@@ -171,8 +216,14 @@ function resetLimitersForTesting(): void {
     fallbackLimiters = undefined
 }
 
+/** Test-only: reports retained account entries for bounded-cache assertions. */
+function getAccountLimiterCountForTesting(): number {
+    return limitersByAccount.size
+}
+
 export {
     createLimiter,
+    getAccountLimiterCountForTesting,
     getMoveLimiter,
     getWriteLimiter,
     type Limiter,
--- a/src/utils/concurrency.ts
+++ b/src/utils/concurrency.ts
@@ -138,10 +138,16 @@ const limitersByClient = new WeakMap<object, LimiterPair>()
  * every call site is bounded even when nothing registered it.
  */
 let fallbackLimiters: LimiterPair | undefined
+let overflowLimiters: LimiterPair | undefined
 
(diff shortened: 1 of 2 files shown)


src/utils/concurrency.ts
```diff
@@ -139,18 +152,50 @@ function accountKeyFromApiKey(apiKey: string): string {
...
+        limitersByAccount.delete(key)
+        entry.lastUsedAt = now
+        limitersByAccount.set(key, entry)
+    } else {
+        pruneIdleAccountLimiters(now)
+        entry = { limiters: createLimiterPair(), lastUsedAt: now }
+
+        // Keep the registry bounded. If every slot belongs to a recently used
+        // account, the new client still gets local limits but is not retained.
+        if (limitersByAccount.size < MAX_ACCOUNT_LIMITERS) {
```
doistbot: <a href="#">P2</a>  An account that arrives after the registry reaches 10,000 is never retained, so every later client for that same account creates a separate limiter pair. Since the HTTP transport creates a client per request, concurrent requests for that overflow account can each admit `WRITES`/`TASK_MOVES` independently, making its effective per-account concurrency unbounded. Use a shared bounded overflow limiter or an eviction/association strategy that still resolves all clients for one account to the same pair.

doistbot (commented):
This PR bounds the per-account concurrency limiter registry, pruning idle pairs after five minutes and capping retention at 10,000 accounts while keeping busy limiters alive.

Few things worth tightening:
- Once the registry is full, accounts that arrive afterward are never retained, so each new client for those accounts builds its own limiter pair. Because the HTTP transport creates a client per request, concurrent requests for such an overflow account can each admit writes/task-moves independently, undoing the per-account concurrency bound. Consider a shared bounded overflow limiter or an eviction/association strategy that keeps all clients of one account on the same pair.

<details>

- **<a href="#">P3</a> src/utils/concurrency.ts:163**: This bounded TTL registry duplicates `BoundedTtlCache` in `src/utils/user-resolver.ts`: same 5-minute TTL, same 10,000-entry cap, and the same Map + timestamp + prune + insertion-order-refresh mechanics. The policies do differ (busy entries must survive past TTL, and at capacity new entries are dropped instead of evicting a live LRU entry), so it isn't a drop-in — but extracting `BoundedTtlCache` into a shared util and adding an eviction predicate (e.g. keep entries whose limiter is busy) would let both callers share one implementation instead of two hand-rolled caches.
- **<a href="#">P3</a> src/utils/concurrency.test.ts:250**: Missing coverage for the contract named in the PR: a client registered when the map is full still gets its own limiters rather than falling back to the shared pair. If `limitersByClient.set` were ever moved inside the `size < MAX_ACCOUNT_LIMITERS` branch, overflow clients would silently share `fallbackLimiters`, reintroducing cross-account contention — and this test would still pass on count alone. Consider keeping the overflow client (the 10,001st) and asserting it runs concurrently with a fresh unregistered client's move limiter, mirroring the 'keeps separate accounts independent' test.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F598&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A223da770-a897-11f1-81b4-c95b8bf1f116%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A223da770-a897-11f1-81b4-c95b8bf1f116&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/utils/concurrency.ts
```diff
@@ -169,10 +225,17 @@ function getWriteLimiter(client: object): Limiter {
 function resetLimitersForTesting(): void {
     limitersByAccount.clear()
     fallbackLimiters = undefined
+    overflowLimiters = undefined
+}
+
+/** Test-only: reports retained account entries for bounded-cache assertions. */
+function getAccountLimiterCountForTesting(): number {
```
goncalossilva: I was going to flag this as a smell, but eventually was able to put my finger on it: we should test behaviors, not implementation details.

Could we test this behaviorally instead? Use two different overflow tokens and verify they share the limiter. That would let us remove this test-only export and stop indirectly asserting on counts.

goncalossilva (approved):
One testing nit, otherwiss LGTM.

src/utils/concurrency.ts
gnapse: Yes, will do. I agree.

scottlovegrove (commented):
👍🏻

goncalossilva:
Did we measure the impact of this and/or #597?

gnapse:
I do not recall. What measurements do you think we could've done? The focus at the time was on releasing a fix to avoid crashing the server once more.

PR #606, opened 2026-09-12
fix: surface deleted tasks in task outputs (Doist/Issues#20642)

PotHix:
# Pull Request

Refs Doist/Issues#20642

## Short description

`mapTask` never included `isDeleted`, so no consumer of it could tell a soft-deleted
task from a live one. In Doist/Issues#20642 that meant `fetch` reported a task deleted
weeks earlier as an ordinary document whose only state field was `checked: false` —
read by support and engineering as "active" — while every backend list path correctly
excluded it. Seven weeks of investigation went into a task that was never there.

### What changed

- **`mapTask` emits `isDeleted`** (`src/tool-helpers.ts`, `TaskSchema` in
  `src/utils/output-schemas.ts`), so `fetch`, `fetch-object`, `find-tasks` and the rest
  can all see it. It is emitted **only when true**: a deleted task is the exception, and
  the alternative is `isDeleted: false` on every task of every list response. Output for
  live tasks is therefore byte-identical to before — no snapshot changed.
- **`fetch` throws for a deleted task** rather than returning a document. The backend
  wrongly answers HTTP 200 for one today (being fixed separately in Doist/Todoist), but a
  "fetch the full contents" tool presenting a tombstone as live content is what caused the
  incident, so it should not depend on that fix. The error path is the repo's existing one:
  `execute` throws, `registerTool` turns it into `getErrorOutput` / `isError: true`, the
  same shape `fetch` already uses for a malformed ID and `fetch-object` for a missing
  section.

### On `checked`

Checked as asked — **not the same defect, so left alone.** `checked` and `completedAt` are
both already in `mapTask` and both declared in `TaskSchema`; `fetch` puts both in its
metadata. Completion state was always visible. What made `checked: false` misleading in
#20642 was the absence of any deletion signal beside it, which is what this PR adds.

### Notes for review

- `fetch-object` still returns a deleted task rather than erroring, now with
  `isDeleted: true` on it. It is the general-purpose object reader and its structured
  output can carry the flag; `fetch` returns free-text `text` for a model to read and
  cannot. Happy to make both error if you would rather they match.
- **Deleted _projects_ are untouched** — `mapProject` omits `isDeleted` too, and `fetch`
  would still return a deleted project as a document. Same class of bug, out of scope for
  this incident; say the word and I will do it here or in a follow-up.
- **Token footprint:** tools/list payload 33205 → 33341 (+136), combined fixed cost
  34233 against the 35000 budget. `TaskSchema` is inlined once per tool that uses it, so
  the field's description is deliberately terse.
- `src/mcp-server.ts` `instructions` is unchanged — this adds an output field, not
  cross-tool routing guidance — so no `npm run eval` run was needed.
- `npm test` (1275 passed), `npm run type-check` and `npm run check` all green.

## PR Checklist

https://claude.ai/code/session_01MQi5BfMsFf7f4issZD6bay

### DIFF
--- a/src/tool-helpers.test.ts
+++ b/src/tool-helpers.test.ts
@@ -82,6 +82,9 @@ describe('shared utilities', () => {
             const result = mapTask(mockTask)
 
             expect(result.isDeleted).toBe(true)
+            // The other half of the contract: a live task omits the field rather
+            // than carrying `isDeleted: false` through every list response.
+            expect(mapTask(createMockTask()).isDeleted).toBeUndefined()
             // Regression guard for Doist/Issues#20642: a deleted task is not a
             // completed one, so `checked` alone cannot be read as "active".
             expect(result.checked).toBe(false)
--- a/src/tool-helpers.test.ts
+++ b/src/tool-helpers.test.ts
@@ -71,8 +71,20 @@ describe('shared utilities', () => {
                 completedAt: undefined,
                 deadlineDate: undefined,
                 responsibleUid: undefined,
+                isDeleted: undefined,
                 addedAt: '2025-08-13T22:09:56.123Z',
             })
+        })
+
+        it('should flag a deleted task', () => {
+            const mockTask = createMockTask({ id: '321', isDeleted: true })
+
+            const result = mapTask(mockTask)
+
+            expect(result.isDeleted).toBe(true)
+            // Regression guard for Doist/Issues#20642: a deleted task is not a
+            // completed one, so `checked` alone cannot be read as "active".
+            expect(result.checked).toBe(false)
         })
 
         it('should handle recurring tasks', () => {
--- a/src/tool-helpers.ts
+++ b/src/tool-helpers.ts
@@ -333,6 +333,9 @@ function mapTask(task: Task) {
         duration: task.duration ? formatDuration(task.duration.amount) : undefined,
         responsibleUid: task.responsibleUid ?? undefined,
         assignedByUid: task.assignedByUid ?? undefined,
+        // Only emitted when true: a deleted task is the exception, and every list
+        // response would otherwise carry `isDeleted: false` on every task.
+        isDeleted: task.isDeleted || undefined,
         checked: task.checked,
         completedAt: task.completedAt?.toISOString() ?? undefined,
         addedAt: task.addedAt?.toISOString() ?? undefined,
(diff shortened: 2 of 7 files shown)


doistbot (commented):
This PR surfaces soft-deleted tasks by emitting `isDeleted` from `mapTask` and `TaskSchema` (only when true, keeping live-task output byte-identical) and makes `fetch` throw for a deleted task instead of returning it as a live document, following the repo's existing error path. No inline issues were flagged — the change follows the established schema, conditional-optional-field, and not-found-error conventions, with regression tests covering both the mapping and the fetch behavior, and the intentional `fetch-object` behavior for deleted tasks is clearly documented in the description.

<details>

- **<a href="#">P3</a> src/tool-helpers.test.ts:74**: Weak assertion: `toEqual` ignores `undefined` properties, so expecting `isDeleted: undefined` here also passes if mapTask ever emits `isDeleted: false` for live tasks — the 'only emitted when true' contract from the PR is unguarded. Add an explicit absence check, e.g. `expect(mapTask(createMockTask()).isDeleted).toBeUndefined()`, or use `toStrictEqual` in this test.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F606&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A23e68280-ae97-11f1-9ea2-75f123a6c73e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A23e68280-ae97-11f1-9ea2-75f123a6c73e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (approved):

PR #592, opened 2026-09-01
fix: preserve day units in task duration output (Doist/Issues#21004)

doistbot:
# Pull Request

Closes https://github.com/Doist/Issues/issues/21004

## Short description

Task mapping now preserves duration units: minute durations retain hour/minute formatting, and day durations are returned with a `d` suffix. Added regression coverage and documented the day-duration output format.

## PR Checklist

- [x] Added tests for bugs / new features
- [x] Updated docs (README, etc.)
- [ ] New tools added to `getMcpServer` AND exported in `src/index.ts`.

PR #609, opened 2026-09-18
feat: support day-based and 24h+ task durations (Doist/Issues#21004)

### LINKED ISSUE #591: Support day-unit durations on add-tasks and update-tasks
Hello team 🤗

### What the user is trying to achieve

Set and read task durations measured in days through the MCP server, so an integration can round-trip a multi-day task without leaving MCP.

The request comes from the owner of the Ganttify for Todoist integration, which renders Todoist tasks on a Gantt chart. Multi-day durations are the normal case there, and they currently have to read the Sync API directly to get them.

### Current behaviour

The MCP layer models duration in minutes on both read and write.

On write, `add-tasks` accepts only the restricted `2h` / `30m` / `2h30m` / `1.5h` syntax, capped at 24h:

```
Invalid duration format "2": Use format like "2h", "30m", "2h30m", or "1.5h"
```

Passing the amount as a number instead of a string returns:

```
tasks.0.duration: Invalid input: expected string, received number
```

So there is no way to express a day duration on input, and the 24h cap rules it out even if there were.

The Sync API documents both `minute` and `day` as valid duration units, so the two surfaces disagree about what a duration can be.

### Suggested direction

Whatever shape suits the tool schema — a `4d` token in the existing string syntax would be the smallest change, but the maintainers may prefer to carry the unit explicitly rather than extend the mini-language.

The reading side is a separate defect rather than part of this request: day-unit durations that already exist are currently returned as minutes. That is filed internally and is not something this request needs to cover.

scottlovegrove:
# Pull Request

Closes https://github.com/Doist/Issues/issues/21004
Closes #591

## Short description

**Bug.** `mapTask` formatted `task.duration.amount` without looking at `task.duration.unit`, so a task stored as `{ amount: 4, unit: "day" }` came back as `"4m"` — understated by 1440× with no error. It now formats by unit: day durations become `"4d"`; minute durations keep hours/minutes (`"48h"`, never `"2d"`) so a day-based and a minute-based duration stay distinguishable and the string round-trips through `parseDuration`.

**Input.** `add-tasks` and `update-tasks` now accept `"3d"` for whole days (mapped to the API's `day` unit) and no longer reject hour/minute durations over 24h. Mixing days with hours or minutes (`"1d2h"`) is rejected because the API stores a single unit. The 24h cap was a first-party product limit; the API accepts these values from integrations and now the MCP server does too.

**Docs.** The `duration` input description (now shared via `DURATION_INPUT_DESCRIPTION`) and the server guidelines say that Todoist apps only display durations under 24 hours, and that a longer or day-based duration is stored and reported faithfully but appears as no duration in the apps. The output schema only gains the `"3d"` example: output descriptions are kept to the bare shape (see #561) and the guidelines block already carries the caveat for every tool result.

`update-tasks` already only sends `duration` when the caller supplies one, so an unrelated edit through MCP never clobbers a stored day/24h+ duration. A test now pins that.

### Why now

The product decision landed in the Comms thread [Inconsistent support for day-based and 24h+ task durations](https://comms.todoist.com/69/ch/BM6NhpsJb8boj8wZGnPvf/t/CedWbmL7eeX7Hd7sxVhAk/): integrations writing these durations is expected behaviour, first-party clients keep their 24h display limit but must preserve what is stored (web: Doist/todoist-web#19347, iOS merged), and the API/CLI/MCP docs should note the display caveat. #592 and #591 were closed pending exactly that decision.

### Verified live

Against a real account via `scripts/run-tool.ts`: `add-tasks` with `"4d"` and `"36h"` → API accepted both and the output showed `"4d"` / `"36h"`; `fetch`, `fetch-object` and `find-tasks` read the day task back as `"4d"`; `update-tasks` with `"2d"` → `"2d"`; renaming the day task left its duration at `"4d"`; `"1d2h"` was rejected with the new message. Probe tasks deleted afterwards.

## PR Checklist


src/utils/duration-parser.ts
```diff
@@ -87,26 +138,31 @@ export function parseDuration(durationStr: string): ParsedDuration {
         throw new DurationParseError(durationStr, 'Duration must be greater than 0 minutes')
     }
 
-    // Validate maximum duration (24 hours = 1440 minutes)
-    if (totalMinutes > 1440) {
-        throw new DurationParseError(durationStr, 'Duration cannot exceed 24 hours (1440 minutes)')
-    }
-
-    return { minutes: totalMinutes }
+    return totalMinutes
```
doistbot, line 155: <a href="#">P2</a>  Reject non-finite amounts before returning them. With the 24-hour cap removed, a sufficiently long numeric input (for example, `'9'.repeat(400) + 'h'`) parses and rounds to `Infinity`, then gets passed as `duration` to the SDK instead of raising `DurationParseError`. Day parsing has the same issue. Check `Number.isFinite` after parsing/rounding in both paths so these inputs fail validation rather than producing an invalid API request.

doistbot (commented):
This PR adds day-based and 24h+ task duration support: `mapTask` now formats durations by their stored unit (e.g. `"4d"` instead of a misleading `"4m"`), and `add-tasks`/`update-tasks` accept whole-day inputs while lifting the old 24-hour cap, with docs noting that first-party apps won't display durations of 24h or more.

Few things worth tightening:
- Guard against non-finite amounts: with the 24h cap removed, very long numeric inputs (e.g. 400 digits) can parse to `Infinity` and reach the API as `duration` instead of raising `DurationParseError` — add a `Number.isFinite` check after parsing/rounding in both the day and minute paths.

<details>

- **<a href="#">P3</a> src/utils/output-schemas.ts:23**: The output schema only gained the `3d` example; it does not state the under-24h display caveat that the PR description says the output schema now documents. Clients reading `duration` from a tool result will see `"3d"` or `"36h"` with no indication that Todoist apps render those as no duration. Add the caveat here or correct the PR description if the omission is deliberate.
- **<a href="#">P3</a> src/tools/add-tasks.test.ts:266**: Day durations are now exercised twice at the tool level: the `4d` row in `should handle various duration formats` (line 238) already asserts the same `addTask` arguments, and this test's only additional assertion — `result.structuredContent.tasks[0]?.duration === '4d'` — is the same `mapTask` formatting already pinned by the `mapTask` day test and the `fetch` day test. Keep one path: drop this standalone test, or drop the `4d` matrix row and keep the richer end-to-end assertion here. `update-tasks.test.ts` has the same duplication between its `2d` matrix row (line 385) and standalone test (line 413).
- **<a href="#">P3</a> src/utils/duration-parser.test.ts:51**: Some new cases repeat inputs already asserted elsewhere in this file: '24h' and '1440m' appear here and in both the 'valid formats' and 'edge cases' blocks, and '25h'/'48h' appear here and again in the round-trip table. Keeping one copy of each (e.g. letting the round-trip table own '25h'/'48h' and trimming the exact-24h edge block to the '23h60m' overflow case) would shrink the suite without losing regression signal.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F609&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad14be4f0-b37a-11f1-8f26-d9dd3c49484a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad14be4f0-b37a-11f1-8f26-d9dd3c49484a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/utils/duration-parser.ts
scottlovegrove: Fixed in eec1158 — both paths now check `Number.isFinite` and throw `DurationParseError("Duration is too large")`. The minutes component is checked before the whole-number test, since `Infinity % 1` is `NaN` and would otherwise have surfaced as the misleading "Minutes must be a whole number". Tests cover `"9" × 400` with `h`, `m`, `d`, and as the minutes part of `1h…m`.

pawelgrimm (commented):
🚀

PR #610, opened 2026-09-20
fix: upload CSV template contents as a Blob

BloomTreasure:
## Short description

Importing a template with `csvFileContent` fails with `Unable to open file as blob` before making an API request. The tool passes the CSV string directly to `importTemplateIntoProject`, but `@doist/todoist-sdk` 15.3.0 treats a string `file` argument as a filesystem path and calls `fs.openAsBlob` on it.

Wrap CSV content in a `Blob` with MIME type `text/csv`. This uses the SDK's existing in-memory upload path and leaves template-ID imports unchanged.

The existing unit test mocked the SDK and expected the incorrect string. It now checks the Blob contents. A separate regression test uses the real SDK with only the HTTP transport replaced, and verifies the multipart filename, target project, and exact UTF-8 content (BOM, accents, quoted fields, commas, CRLF and a multiline field). It fails with the original error before the fix and passes afterward. All test data and credentials are fictitious; no live API access is required.

Validation on Node 26.8.2:

- 21 focused tests passed.
- Full suite: 79 test files and 1,305 tests passed.
- `npm run type-check`, `npm run format:check` and `npm run build` passed.

## PR Checklist

- [x] Added regression tests for the bug.
- No public schema, tool description, dependency or documentation change required.

Prepared with OpenAI Codex; the regression was run against the real SDK locally.


scottlovegrove (approved):
Thanks for fixing this.

PR #614, opened 2026-09-22
docs: hint at comments in fetch

### LINKED ISSUE #613: fetch should include a task's recent comments
## Summary

The OpenAI-spec `fetch` tool returns a task's title, description, due date, labels and metadata, but not its comments. Clients that read tasks as documents (ChatGPT's `search` → `fetch` flow) therefore never see comments unless the model separately decides to call `find-comments`, which it doesn't always do.

## Why it matters

A user coordinating with AI assistants through task comments reported a comment as "invisible to the connector for 18 minutes". The comment was on the server the whole time (`find-comments` reads live, no cache). The first check went through `fetch`, which cannot surface comments; a later check happened to use `find-comments` and returned it immediately. From the user's side that looks like a replication delay, and it undermines using comments as a coordination channel with assistants.

## Suggestion

Include the most recent comments in the fetched task document:

- Append them to the `text` body (author display name, timestamp, content), and/or
- Add a `comments` array to `metadata`

Cap the count (for example the 10 most recent) and note in the tool description that the document includes recent comments, so the model knows it doesn't need a second call for the common case. Projects could get the same treatment for project comments.

## Reference

[Reported via a support ticket](https://todoist.zendesk.com/agent/tickets/943656); no customer identifiers included here.

gnapse:
## Overview

`fetch` now notes that a task or project may have comments and directs models to `find-comments` when they matter. This keeps comments opt-in, avoiding extra API requests and automatic context growth on every fetch.

The current comments API only paginates from oldest to newest and has no sort or filter option, so the MCP server cannot efficiently retrieve a bounded set of the most recent comments.

Closes #613


doistbot (commented):
This PR adds a small hint to the `fetch` tool description pointing models toward `find-comments` when comments matter, keeping comment retrieval opt-in rather than automatic on every fetch. No inline issues were flagged — the change is a documentation-only edit that accurately references the existing `find-comments` tool, follows the repo convention of placing cross-tool routing guidance in the tool's own description, and doesn't touch runtime behavior, schemas, security surface, or the token budget.

<details>

- **<a href="#">P3</a> src/tools/fetch.ts:44**: Reference `${ToolNames.FIND_COMMENTS}` here instead of hardcoding `find-comments`. `ToolNames` is the shared source of truth, so interpolation keeps this routing hint correct if the tool is renamed.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-mcp&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-mcp%2Fpull%2F614&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac70fa110-b682-11f1-9e20-72609531a243%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac70fa110-b682-11f1-9e20-72609531a243&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (commented):
👍🏻

